PR #4, opened 2012-12-21
Update wca-regulations-2012.md

SAuroux:
A3d3: I must have missed that change before. Why would we change this? "12 seconds" is long and unfamiliar. It is better to have a short signale at this point of the preinspection.
A6h: The penalty made no sense there. With that wording it would be a DNF if ANYONE made a move or an alignment. If a competitor does so, then this case is covered by A6e.

### DIFF
--- a/wca-regulations-2012.md
+++ b/wca-regulations-2012.md
@@ -24,7 +24,7 @@ The use of the words "must", "must not", "should", "should not" and "may" comply
   - 1c10) Making a copy of the WCA Regulations available at the competition.
 - 1e) Each event must have one or more judges.
   - 1e1) A judge is responsible for executing the procedures of the event.
-    - 1e1a) A judge may judge multiple competitors simultaneously at the discretion of the Delegate, as long as the judge is able ensure that all WCA Regulations are followed at all times.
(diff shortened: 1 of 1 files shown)


iVinshe:
A3d3: wasn't that supposed to be a "twelve"? I think this was the consensus
of Worlds Staff that most people called "twelve" instead of "go".

A6h: it's not an auto-DNF. It simply spells out that the Delegate can DNF
someone's attempt if they feel they have to (e.g. competitor deliberately
being a nuisance).

Vincent Sheu
University of California, Berkeley
vsheu@berkeley.edu

SAuroux:
A3d3: I'd say that in Europe most people say "GO". Let's keep it that way.
Also no one will blame you for saying "twelve".

A6h: How does your comment effect my point? It is still 100% valid.

Also please don't miss the few text mistakes I fixed.
Am 22.12.2012 06:10 schrieb "Vincent" notifications@github.com:

lgarron:
If it doesn't really matter between GO and 12 SECONDS, I would recommend the latter, since it's consistent with 8 SECONDS and more clear for people inexperienced with the regulations.

»Lucas Garron

iVinshe:
I'll concede on A6h.
Vincent Sheu
University of California, Berkeley
vsheu@berkeley.edu

PR #5, opened 2012-12-23
Restore GO at 12 seconds.

lgarron:
So, I would prefer "12 seconds".
- It's consistent with 8 seconds.
- As has been pointed out, it isn't strictly important what the judge says. "GO" is still acceptable.

I don't care strongly, but if we want to revert this, this pull request should make it easy to include.


lgarron:
@SAuroux @iVinshe

PR #18, opened 2013-01-31
fix spelling issues pointed out by Patrick Hetco

timhabermaas:
Pointed out [here](http://forum.speedcubers.de/showthread.php?tid=10287&pid=170838#pid17083
8).


lgarron:
Now that the German repo has a shared history, this pull request should be submitted there.

PR #162, opened 2014-03-11
Fixing A1b1) possible ambiguity with Stackmat (as defined)

cubizh:
Clearing the possible ambiguity between the previously defined term Stackmat (which is the combination of timer + mat) and the Stackmat timer only, in A1b1).


lgarron:
Hmm, this "ambiguity" is what makes the mat itself optional for e.g. big cube BLD.
But if that was the main intent of the end of A1b1, it's definitely not clear enough, so I'm fine with taking it out.

PR #164, opened 2014-03-11
Fixing numbering from text to numbers to maintain consistency

cubizh:
Previous rules have numbering (4,7, 11), using '2' instead of 'two' maintains the consistency.


lgarron:
Thanks!

PR #167, opened 2014-04-04
3. April update

SAuroux:


lgarron:
Updated at https://github.com/cubing/wca-documents-translations/tree/german

PR #186, opened 2014-05-21
Solve-someone-elses-scramble.md

avgalen:
Prevent cheating by scrambling for yourself or by judging a cube that you scrambled for a friend

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -155,6 +155,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 - 4f) Competition scramble sequences must be generated using the current official version of the official WCA scramble program (available [via the WCA website](https://www.worldcubeassociation.org/regulations/scrambles/)).
 - 4g) After scrambling a puzzle, the scrambler must verify that he has scrambled the puzzle correctly. If the puzzle state is wrong, he must correct it (e.g. by solving the puzzle and applying the scramble sequence again).
     - 4g1) Exception: For the 6x6x6 Cube, 7x7x7 Cube, and Megaminx, it is not necessary to correct the scramble, at the discretion of the WCA Delegate.
+- 4h) After scrambling a puzzle, the scrambler is not allowed to purposefully judge or solve a puzzle that he scrambled himself. Penalty: Disqualification from the competition of the scrambler
 
 
 ## <article-5><puzzle-defects><puzzledefects> Article 5: Puzzle Defects


Claster:
A scrambler can't solve a puzzle scrambled by himself by 1h1. As for judging: sometimes at the beginning of a new group or event, things can be sped up if several judges also scramble puzzles for their competitors. If someone wants to cheat, he can do it while being only a judge ([example](https://www.worldcubeassociation.org/forum/viewtopic.php?f=9&t=1067)). Solution is using only trustworthy judges.

avgalen:
Thank you for pointing me at 1h1. I had to read it a few times to see if it would really cover all abuse but you are right: It covers all abuse for a scrambler to scramble for himself.

For judging your example is well taken and your solution is (of course) the only real solution. I also recognize the (unfortunate) situation that judges help with scrambling at the beginning of a new group but NOT that they scramble for their own competitor. In that situation they should scramble a few puzzles and then pick up a random puzzle that was (probably) scrambled by somebody else. This is the reason I put purposefully in the rule.

Laura-O:
How can you judge if this was done by purpose (and therefore a
disqualification from the competition is appropriate)?
Furthermore: why do we need this regulation? Is this really a problem in
competitions? Is this something we can not manage with the existing
regulations (e.g. 2k)?

2014-05-21 16:20 GMT+02:00 avgalen notifications@github.com:

avgalen:
"Scramblers scramble, judges judge, and never the two shall meet"
If a judge brings a cube to the scrambling table, scrambles it and walks away I don't trust it.
If a scrambler scrambles 20 cubes, finishes scrambling and helps with judging at the end of the group he MIGHT pick up one of the cubes that he scrambled but I would trust that.
The whole point is that I don't trust little boy X that scrambles for his friend Y and also judges him. Apparently this is happening: (source http://www.speedsolving.com/forum/showthread.php?47691-Multiple-judges-per-average) Originally Posted by hcfong  View Post
...One person would pretty much claim to cube of his or her friend and hold on to it and both scramble and judge the solves...
And yes, this pretty much falls under https://www.worldcubeassociation.org/regulations/#2k2 so both my concerns are already covered under the current regulations and I withdraw my pull request (how do I do that in github? just "Close & Comment"?

lgarron:
Yeah, `Close & Comment` should work. But I'd like to leave it open for now. I think we could use a guideline that encourages accountable behaviour.

lgarron:
Closing in favor of https://github.com/cubing/wca-documents/issues/121#issuecomment-64724071.

PR #274, opened 2015-06-17
A7c) Removing +2 penalty

KitClement:
Requested for the sake of simplicity -- would create a 9th +2 penalty.


lgarron:
Really? Wasn't the sentiment on the Delegate list that this is too harsh?

(And if it's so harsh that Delegates/judges would decline to enforce it... what's the point of an official standard?)

KitClement:
Yep, I meant to change back to the judge signature too with this commit,
will do when not on mobile. Thanks for catching.

PR #285, opened 2015-06-22
Fix spellings related to organise.

kotarot:
I changed three `organization` to `organisation`, one `organizer` to `organiser`.


KitClement:
Thank you!

PR #292, opened 2015-07-22
FMC is undefined in the regulations

cubizh:
Although FMC is well known, it is undefined and should be replaced by what's used elsewhere "Fewest Moves"


Laura-O:
That's true. Thank you, @cubizh!

lgarron:
Indeed; thanks!

PR #293, opened 2015-07-22
Result instead of attempt time for time limit

cubizh:
In the "attempt" definition, I think that "results" is a better fit here and for the continuation of the next sentence.

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -349,7 +349,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 - A1) Speed Solving attempts must abide by the following procedures.
     - A1a) The organisation team must enforce time limits for attempts and/or rounds.
         - A1a1) The default time limit per solve is 10 minutes, though the organisation team may announce a higher or lower time limit.
-        - A1a2) Cumulative time limits may be enforced (e.g. 3 attempts with a cumulative time limit of 20 minutes). The time elapsed in a DNF result counts towards the cumulative time limit.
+        - A1a2) Cumulative time limits may be enforced (e.g. 3 results with a cumulative time limit of 20 minutes). The time elapsed in a DNF result counts towards the cumulative time limit.
         - A1a3) For each round, any time limits must be announced before the round starts, and should not be changed after it has begun. Changes must be made at the discretion of the WCA Delegate, who must carefully consider the fairness of the change.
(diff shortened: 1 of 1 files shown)


Laura-O:
Can you explain that a bit more, @cubizh?
I think _attempt_ fits quite well here and it's also used in A1a).

cubizh:
Of course.
I think A1a) is on the fence too, but it is more noticable on A1a2) with my interpretation.
It all boils down to what an "attempt" is, as defined by the regulations.
An attempt is defined by regulation A3b2) (how it starts) and by regulation A7c) (how it ends). 
In practice, we don't use A1a)/A1a2) with the time that the attempt takes (with inspection time added and signing the solve) but really only use the actual solve time or the "result" (what appears on the timer) for time limits. At least I think that's the spirit of this regulation. I feel that A1a2) should refer to the same "solve time" as A1a1). That's why I am not much of a fan, for consistency, of having the word "attempt" there, but I welcome the opinion of others. :)

Laura-O:
Thank you, I got your point now and I completely agree to your explanation.
That's also consistent with how _attempt_ and _result_ is used in 9f) and other articles.

lgarron:
I don't think it makes sense to change this without A1a).

I think we should change everything under A1a) to say "solve" or "attempt", and I personally think it's most natural to say that the attempt has a time limit (which is enforced based on the final result based from the solve).

Laura-O:
What about "results of 3 attempts..."?
This emphasizes that it's not a time limit for the whole attempt (from A3 to A7c, as @cubizh explained) and also connects the first and second sentence.

P.S.: sorry for merging this a bit too hasty. :-)

cubizh:
I agree A1a) could be rephrased but I don't think A1a) is that much wrong, because by setting a time limit to the solve, you are already (sort of) establishing a time limit to the attempt, because of the 1 minute pre-inspection + 15s inspection + solve time. The only gray area is the signing part that is untimed. But if that last part is neglected, the attempt has a time limit if a limit is set on the solve. Some people may misinterpret timing the attempt and timing the solve, unless there's somewhere explicitly said that the time of an attempt is measured by the solve time only.

lgarron:
Is A1a5 not clear enough?
(It currently still says "solve", although I think "attempt" would be better.)

cubizh:
I agree with changing A1a5) to "attempt" instead of "solve", but it still is unclear that the time of the attempt is supposed to be the time of the solve for A1a2) purposes (or others where it applies). Perhaps an [addition] guideline?

PR #302, opened 2015-08-20
3d1) Specify "blind" and "colour blind".

lgarron:


lgarron:
@KitClement, @Laura-O : Does this sound good to you?

Laura-O:
Yes, sounds good.

KitClement:
I agree, thanks for the fix here as well as keeping up with other issues
while I've been moving. Just starting to get settled here and I still have
quite a bit to catch up on, but we should have a Skype meet soon.

PR #310, opened 2015-11-01
Let's keep the spelling of colour consistent

gyroninja:
Just noticed that "colors" is used at one location in the regulations instead of "colours".


lgarron:
Closing per the last comment.

PR #323, opened 2016-02-10
Fix issue 315

Claster:
Clarify that a competitor may, not must, remove the object in Square-1 during inspection, and they can do it during the solve as well.


viroulep:
Hey @Claster,
FYI you can amend your last commit using `git commit --amend` to solve issues like adding swap files ;)
That way it doesn't even appear in the history.
You may want to put these in your global gitignore too.

Claster:
Hi Philippe,

Thanks! I've already added this to my global gitignore after that commit.
WRT amending, is it safe to do it now? Or maybe not worthy?

viroulep:
You're welcome !
I don't think it's worth : it's not a big issue, and I don't like rewriting history on a public branch, since forcing the push will probably mess up with someone's local history.

Issue #300, opened 2015-08-20
Clarify that any part of the body is permitted, by default


lgarron:
From [this thread](https://groups.google.com/forum/#!msg/wca-wrc/92nUKJ6gRtw/W-gP1murAAAJ) in the Delegate list.

Most Delegates agree that using the knee (or any other body part) is fine. But people sometimes ask about this, and I think it's better to be explicit. An `A5b+` Guideline is probably the most appropriate place.

pedrosino:
I agree with this. 
What is the policy on these small changes for the WRC?

Laura-O:
I don't see a reason why such an issue should be changed in-between, separated from the next major change. It's just a small, uncontroversial clarification for a reappearing question, not an urgent fix.

In my opinion we should collect these issues (there are several more) and release them together. Anything else would eventually lead to an unstructured regulation-updates chaos and it would be really hard to keep track of the changes.

Laura-O:
A similar incident was reported [here](https://groups.google.com/d/msg/wca-delegates/YmwvySyQWMk/MXG3LoSxzHsJ): two competitors used their legs to fix pops.

PR #324, opened 2016-02-12
Clarify that any body part can be used (#300)

Claster:


lgarron:
Shouldn't the exception also include Feet?

Claster:
Indeed, thanks! PTAL.

lgarron:
I would suggest splitting the parenthesis into two so that it's
crystal-clear which event which link belongs to.

Issue #189, opened 2014-06-09
Clock Pin Caps


KitClement:
This is pertaining to the issue brought up in the Guangzhou Wushan Open 2014 delegate report.  Here are the two additions:

https://github.com/KitClement/wca-documents/commit/883abec0b7cbca921757510dabb44e289c3c5c1a

My wording might not be perfect here, let me know if you have any changes/additions you would make.

lgarron:
Thanks for making this.

One this is that "pins must not be identifiable from any other pin of the same side" allows you to feel which side is in front without relying on visual markings.
Unless there is really good reason to allow it (is the availability still so low that competitors are forced to settle for broken clocks for some reason than their own level of use?), I think we should treat them as similar _from either side_ by default.

Should we get information on whether anyone is using faulty clocks?

I agree that missing caps at the end of the solve should be treated the same as any other puzzles.
(Although I'm still of the opinion that "solved or not" is sufficiently fair and simple.)

KitClement:
I'll poll a few cubing facebook groups I'm in to see if people do use clocks with missing pin caps.  I know that at Carnegie Mellon Fall 2013, I let one of the kids there borrow my clock, and he returned it with a cap missing - presumably he did some official solves with a missing cap.  

Clocks (at least Rubik's brand) are becoming somewhat difficult to find, I often search for new ones myself and it's difficult to find shipping to the USA, let alone internationally.  I wonder if the knock off clock pin caps will fit into a Rubik's?  I don't own one myself to try.

I'm pretty indifferent to any of the issues you brought up, whatever you think might be best.

Laura-O:
From my experience those faulty clocks (with one or more missing pin caps) are used quite regularly. I always asked the delegate about this when I saw them at the scrambling table. Competitors never questioned that such a missing pin cap might help you to identify the orientation of the clock.

Nevertheless I would like to bring something else into discussion: due to their age, their design and their availability, there are more attributes which make a clock "faulty". A lot of clocks have some kind of flaws and or damages. Cases break when they are slammed too hard on the table, the inner clocks sometimes grind the case from the inside, the inlays are shredded or there is dirt inside.
Furthermore some people mark their clocks in order to identify them better. I put a small sticker on the upper side of the clock (inside the case) after it was mixed up at the scrambling table more than one time. All these things can be utilized more or less to have the same advantage that a missing pin cap might give you.

KitClement:
[...]

Issue #165, opened 2014-03-26
Can an FMC judge also be an FMC competitor?


sarahstrong314:
Just a thought. I can't find anything in the Regulations that say that a someone judging FMC can't also compete in it.

lgarron:
Is there anything against a competitor judging themselves for  speedsolving?»Lucas Garron

sarahstrong314:
Common sense for speedsolving events, I suppose, but for FMC the judge has few responsibilities and could easily compete without breaking any rules.

Laura-O:
This question is still asked frequently: https://groups.google.com/d/msg/wca-delegates/G6cvoSV1g3I/_OymNqnLBwAJ

Although 1h) does not clearly forbid to be judge and competitor simultaneously, fulfilling both roles at a time is rather unusual. I think we should finally clarify this in a guideline for E2a)

PR #329, opened 2016-02-22
Fixes #68, #165 and #189

### LINKED ISSUE #68: FMC Tools
[Regulation E3](http://www.worldcubeassociation.org/regulations/#E3) is not very clear.

See [WCA delegate discussion started by Yohei Suzuki on August 03, 2013](https://groups.google.com/forum/#!searchin/wca-delegates/FMC$20suzuki/wca-delegates/mA9BpLyqn-U/T1mOfqVuStQJ).

Laura-O:


Claster:
So, transparent tape is also incluced into "stickers", isn't it?
Do you think it should be clarified what exactly "sticker" is, or just use the common sense that says "anything that sticks"?

Laura-O:
I think the definition of "sticker" is just any adhesive label or decal. So if you use the transparent tape in this way, that is OK.

I don't think giving a further definition will help us. Competitors have so many ideas what they can use as stickers and I am sure we will not be able to find a definition for all these ideas.

lgarron:
I would have split these into separate pull requests to manage the discussion, but since we're already here, I'll comment on the pins.

Issue #227, opened 2014-12-10
Gen 3 vs. Pro timers


lgarron:
[8a7](https://www.worldcubeassociation.org/regulations/#8a7)) Use the authentic Speed Stacks Stackmat timer (Generation 2 or Pro) for time measurement.

I still don't know if there is an actual difference between Gen 3 and Pro, but we should probably make this as clear as possible.

(Reported to the WRC by Stefan on Nov. 23.)

lgarron:
I sent [an email](https://groups.google.com/forum/#!topic/wca-delegates/HTJoEMQblmk) to the Delegates list:

> The Regulations currently state:
> - 8a7) Use the authentic Speed Stacks Stackmat timer (Generation 2 or Pro) for time measurement.
> 
> Technically, this means Gen 3 timers (which look like Gen 2 timers without a data port) are not allowed. It's easy to add them to 8a7, though. Is anyone currently using them?

lgarron:
Also [posted to speedsolving.com](https://www.speedsolving.com/forum/showthread.php?51081-Have-you-been-to-any-competitions-using-Gen-3-timers&p=1048007#post1048007).

lgarron:
No one [on the Delegate list](https://groups.google.com/forum/#!msg/wca-delegates/HTJoEMQblmk/F0ghgSc7VxwJ) or [on speedsolving.com](https://www.speedsolving.com/forum/showthread.php?51081-Have-you-been-to-any-competitions-using-Gen-3-timers&p=1048007#post1048007) reports active use of Gen 3 timers.

I'm closing this issue; if someone "needs" to use Gen 3 timers, we can bring it up again.

PR #333, opened 2016-02-28
Added the new timer generation

Laura-O:


viroulep:
Looks good !

Claster:
Sorry for looking at it only now, after it was merged.

Don't you actually confuse two different timers? There is gen 3 timer (without data port IIRC) which is not allowed at the comps currently, and there is pro timer which is allowed. Is it a good idea to write "Generation 3 Pro Timer"?

viroulep:
I wondered about this too and found out that Speedstacks also refers to the pro timer using "Gen 3 pro timer" (see [here](https://www.speedstacks.com/store/retail/speed-stacks-stackmat-pro-timer/), the note line).

Laura-O:
Yes, I had also asked Speed Stacks about the proper naming of the timer generations and they confirmed that.

There is also an issue about the difference between Generation 3 and Pro Timers here: #227 
As obviously nobody was using them back then and they are not available anymore, I think it's correct to use the currently used naming in the regulations.

lgarron:
> I wondered about this too and found out that Speedstacks also refers to the pro timer using "Gen 3 pro timer" (see here, the note line).

Hmm, it seems you're right. Also here: https://www.speedstacks.com/store/retail/speed-stacks-stackmat-pro/

I wish they were more clear about this. I think the change we have now is accurate enough, but I'm worried it will cause confusion about whether "non-Pro" Generation 3 timers are now allowed. Not everyone is likely to know the distinction. Maybe we should have a page which lists permitted timer equipment, too. :-/

Claster:
Or maybe add a clarification to guidelines: "Gen 3 timers which are not pro are not allowed"?

TimMcMahon:
Should particular batches of Speed Stacks timers be reviewed and added to a list of approved timers?

Speed Stacks might modify the Gen 4 Pro Timer over the next 12 months if they identify issues with earlier batches.

Laura-O:
What would be the aim of that?

Those "batches" are released without an announcement. When the battery compartment of the Gen3 timers was updated, online shops sold the old versions even a year later...
It would make things really complicated to differentiate between these batches.

I agree that it would be good to make a list of approved equipment, but breaking this down to the updates of timer generations seems to be more confusing than helpful.

PR #335, opened 2016-02-28
Define size of full-size mats

### LINKED ISSUE #304: Allow custom mats?
I remember someone saying they made custom mats or took a gen2 one and cut it in half.

I think this is a good and valid idea, and don't see the point in requiring a 80cm-wide mat just because they are sold that way.

Laura-O:
Fixes #304

I was not sure if we should also define the size in inches. But as the majority of the WCA countries is using the metric system, I think it's fine this way.


viroulep:
Neither cm or inches has been used yet in the Guidelines, maybe we could include something like `(approximately 11.81" x 9.84")` at the end of the guideline ?

Laura-O:
Changed "dimension" to "dimensions", so LGTM!

lgarron:
Excellent; thanks for handling this!

PR #337, opened 2016-03-01
Clarify 3d)

### LINKED ISSUE #312: Misinterpretations of 3d)
As a result of [this discussion on the delegate list](https://groups.google.com/d/msg/wca-delegates/p3W7ots20_M/E47ifjNUBAAJ), I would like to suggest to improve the formulation of 3d). This was not the first time 3d) was brought into question, so I think it's time to change it.

My suggestions:
- Clarify that the "or" in 3d) is meant as an exclusive-or, i.e. coloured plastic, stickers and tiles may not be combined.
- Clarify that these coloured parts of the puzzle define the colour scheme and the paragraph is not about any other parts of the puzzle.

Spontaneous suggestion: _Puzzles must have coloured parts, which define the colour scheme of the puzzle and must be only one of the following: coloured stickers, coloured tiles, coloured plastic, or painted/printed colours. All coloured parts of a puzzle must be made of a similar material._.

Laura-O:
As mentioned in the issue, this addition will help to clarify that 3d) is about the parts of the puzzle that define the colour scheme and not any parts of the puzzle. Furthermore, "either" emphasizes that combinations of those coloured parts are not allowed. 

Fixes #312

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -116,7 +116,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 3a1) Competitors must be ready to submit their puzzles when they are called (see [Regulation 2u](regulations:regulation:2u)).
     - 3a2) Puzzles must be fully operational, such that normal scrambling is possible.
     - 3a3) Polyhedral puzzles must use a colour scheme with one unique colour per face in the solved state. Each puzzle variation must have moves, states, and solutions functionally identical to the original puzzle.
-- 3d) Puzzles must have coloured parts, which define the colour scheme of the puzzle and must be either coloured stickers, coloured tiles, coloured plastic, or painted/printed colours. All coloured parts of a puzzle must be made of a similar material.
+- 3d) Puzzles must have coloured parts, which define the colour scheme of the puzzle and must be one and only one of the following: coloured stickers, coloured tiles, coloured plastic, or painted/printed colours. All coloured parts of a puzzle must be made of a similar material.
     - 3d1) For competitors with a medically documented visual disability, the following exceptions apply:
         - 3d1a) Blind competitors may use textured puzzles with different textures on different faces. Each face should have a distinct colour, to aid in scrambling and judging.
         - 3d1b) Colour blind competitors who cannot distinguish between the necessary number of colours may use coloured parts with patterns, if it has been explicitly approved according to [Regulation 2s](regulations:regulation:2s). Patterns may come from stickers or be drawn.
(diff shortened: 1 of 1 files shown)


wca-regulations.md
```diff
@@ -116,7 +116,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 3a1) Competitors must be ready to submit their puzzles when they are called (see [Regulation 2u](regulations:regulation:2u)).
     - 3a2) Puzzles must be fully operational, such that normal scrambling is possible.
     - 3a3) Polyhedral puzzles must use a colour scheme with one unique colour per face in the solved state. Each puzzle variation must have moves, states, and solutions functionally identical to the original puzzle.
-- 3d) Puzzles must have coloured parts, which must be one of the following: coloured stickers, coloured tiles, coloured plastic, or painted/printed colours. All coloured parts of a puzzle must be made of a similar material.
+- 3d) Puzzles must have coloured parts, which define the colour scheme of the puzzle and must be either coloured stickers, coloured tiles, coloured plastic, or painted/printed colours. All coloured parts of a puzzle must be made of a similar material.
```
Claster: Shouldn't it be "either A or B or C or D"? Or maybe "A, B, C or D"? Though the latter doesn't emphasize that we can't pick several options.

I'm not English native speaker, just found a couple of links:
http://english.stackexchange.com/questions/170409/whats-the-best-way-to-use-either-on-more-than-two-options
http://english.stackexchange.com/questions/40950/is-either-only-used-with-two-options

lgarron: "A, B, C, or D" is correct, since [the Regulations use](https://github.com/cubing/worldcubeassociation.org/blob/master/regulations/style-guide.md) the [serial comma](https://en.wikipedia.org/wiki/Serial_comma).

Laura-O: In the links you quoted, people answered that "either" can be used for more than one option if you want to emphasize the exclusive-or.

[Here](http://www.thefreedictionary.com/either) is an explanation for "more than two":
 When either is used as a conjunction, no paraphrase with any is available, and so either is unexceptionable even when it applies to more than two clauses: Either the union will make a counteroffer or the original bid will be refused by the board or the deal will go ahead as scheduled.

So, I'm not a native speaker either and I think that would not be appropriate for a novel, but to emphasize the exclusive options here it's fine. Someone correct me if I am wrong.

viroulep: If @lgarron and @KitClement think it's ok we should merge it as is.

Random idea in case it's not : what about changing "must be one" to "must be one and only one"?

Laura-O:
I changed the regulation to Philippe's suggestion.
Anyone want to review and merge it? :-)

viroulep:
LGTM :)

PR #341, opened 2016-03-03
Change Regulations and Guidelines about extra attempts

viroulep:
This is a tentative update of 11e2-related Regulations and Guidelines, any remarks and suggestions are welcome !

I'm not sure how the example should looks like now, since the resolution of an extra attempt is pretty straightforward after this update: feel free to suggest a more relevant or useful scenario.


viroulep:
I updated the PR.
Interestingly, Laura's comment doesn't appear as usual (ie on an "outdated diff"), so FYI she has made it on commit e979e02a8976cf61c3b9b8a66c20240efd178496.

wca-regulations.md
```diff
@@ -295,7 +295,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
...
 - 11e) If an incident occurs during an attempt, the WCA Delegate may grant a competitor an extra attempt, replacing the attempt during which the incident occurred. The competitor must appeal verbally or in writing to the judge and WCA Delegate at the time of the incident, before finishing the original attempt, to be eligible for an extra attempt. An appeal does not guarantee the competitor an extra attempt.
     - 11e1) If a competitor is granted an extra attempt, the extra attempt must be scrambled using a different scramble sequence. This scramble sequence must be generated using the current official version of the official WCA scramble program (see [Regulation 4f](regulations:regulation:4f)).
-    - 11e2) If a competitor receives any extra attempts, the regularly numbered attempts should be completed first, and extra attempts should be done afterwards. The results from the extra attempts should replace the regularly numbered attempts that caused them, in increasing order.
+    - 11e2) If a competitor is granted an extra attempt, the extra attempt should be done right after the attempt that caused it, and should replace the original regularly numbered attempt.
```
Claster: 1) Maybe "... and **the result of the extra attempt** should replace ..."?
2) Maybe "... and the result of the extra attempt **must** replace ..." (must instead of should)? Because, well, what else can it replace, I see no reason why this part can be not applied.

viroulep: For 2 you are right, there is not much choice :p 

For 1 I think it's not semantically correct : with your suggestion a _result_ would replace an _attempt_. Adding "... should replace **the result of** the orginal regularly numbered attempt." would solve the issue, but I would go with the version using _attempt_, since 11e deals with attempts and not results.

Laura-O: Also, for 1:
An extra attempt can be replaced by another extra attempt and an extra attempt is not a result (or is it?), so I think it's clearer to say that the attempt replaces the attempt.

viroulep: It's not, from the style guide : 
- Attempt: The time frame starting when the competitor acknowledges his readiness to begin (see Regulation A3b2) and ending when the score sheet is complete (see Regulation A7c).
- Result: An official time, either the time for an attempt including penalties, or a derived result (e.g. mean/average).

viroulep:
PR updated with Oleg's comment ("should" -> "must").

wca-regulations.md
Claster: Well, I meant actually "result replaces result" :-P
My point was that if attempt is a time frame, it can't replace another attempt, but the result can.
But then there is an issue that an original attempt often has no result at all.
If you think it is too verbose, go ahead with "attempt replaces attempt".

PR #342, opened 2016-03-07
Added 4b4) time frame for scramble sequences

### LINKED ISSUE #295: Proposal: Require that scramble sequences for one round are only used on one calendar day
There are competitions where 4/5 BLD can be done on two consecutive days.
I am not sure if this is done likewise in other parts of the world, so I will explain it briefly: when a competitor wants to start an attempt he finds someone to judge him, informs the delegate and submits his puzzle. The judge receives the scrambled puzzle and the scoresheet. The competitor and the judge can then start the attempt (usually in a side room).

This is generally a good approach as it makes it possible to do these unpopular events without blocking a specific time slot. BigBlinders also like it because they can choose when they want to compete and have longer pauses between their attempts. However, delegates have to be very careful and always keep an eye on who is judging and that all regulations are followed.
As a down side this approach gives competitors some additional ways to cheat. In a worst case situation a competitor uploads a video of a solve on the first day of the competition and another competitor analyses the solve, practices the memo and does the attempt on the next day. Hypothetically he could have more than 24 hours to do this.

I know that some delegates already do so, but I want to suggest to add a paragraph to the regulations that a scramble sequence for a round may only be used on one calendar day. This would significantly reduce possible ways of cheating. 
This could be either added as [4b4](https://www.worldcubeassociation.org/regulations/#4b) or added to [4b2](https://www.worldcubeassociation.org/regulations/#4b2).

Laura-O:
This commit fixes #295.

2 hours might seem a bit harsh at first sight, but by limiting the time frame to individual attempts that's OK and it's not difficult to implement it in practice.

### DIFF
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -75,6 +75,7 @@ To be more informative, each Guideline is classified using one of the following 
 
 - 4b1+) [REMINDER] The WCA Delegate must never re-generate any scrambles to replace other ones for the purpose of filtering. For example, it is not be permitted to glance at the scrambles for a competition and generate the entire set again in order to generate "fairer" scrambles.
 - 4b2+) [CLARIFICATION] In general, all official scramble sequences should be kept secret during the competition and published together after the end of the competition (see [Regulation 1c3a](regulations:regulation:1c3a)). In some cases (e.g. world records), the organisation team may wish to release specific scrambles sooner after the end of a round.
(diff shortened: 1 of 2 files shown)


viroulep:
We should handle this PR at some point.

A 2 hours time frame sounds ok, and nobody complained about the suggestion in #295 so I'm merging this.
Also I know we should not re-use a guideline number (4b4+), but this seems like the situation where we want to do it.

PR #346, opened 2016-03-18
Fixes #314

### LINKED ISSUE #314: Number of competitors who advance
As per the discussion on the delegates list: https://groups.google.com/forum/#!topic/wca-delegates/CPqUZ4acu6c

Chris' suggestion:

> Again, the suggestion was that 9m simply be applied to all rounds, rather than just the first. Consider this wording:
> 9m) Events must have at most four rounds.
> 9m1) Rounds with 99 or fewer competitors must have at most two subsequent rounds.
> 9m2) Rounds with 15 or fewer competitors must have at most one subsequent round.
> 9m3) Rounds with 7 or fewer competitors must not have subsequent rounds.

I am ok with this. 

However, there were other suggestions to limit the number of advancing competitors as "percentage of people who get eliminated must be "at most equal, but no greater than" the one used for the previous round".
I don't think this is very practical, because there are many variables to consider when determining this number. Having a minimum % of people who must advance sounds bad to me.

Laura-O:
Checked the database for recent competitions that would have been affected by this regulation... and I can only say it makes sense. :-)
#314

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -256,9 +256,9 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 - 9k) All competitors may participate in all events of a competition, except in cases specifically approved by the Board.
 - 9l) Each round must be completed before any following round of the same event starts.
 - 9m) Events must have at most four rounds.
-    - 9m1) Events with 99 or fewer competitors must have at most three rounds.
-    - 9m2) Events with 15 or fewer competitors must have at most two rounds.
-    - 9m3) Events with 7 or fewer competitors must have at most one round.
+    - 9m1) Events with 99 or fewer competitors must have at most three subsequent rounds.
+    - 9m2) Events with 15 or fewer competitors must have at most two subsequent rounds.
+    - 9m3) Events with 7 or fewer competitors must have at most one subsequent round.
 - 9o) Combined rounds count as one round when counting the number of rounds per event.
 - 9p) If an event has multiple rounds, then:
     - 9p1) At least 25% of competitors must be eliminated between consecutive rounds of the same event.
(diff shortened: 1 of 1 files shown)


Claster:
You probably meant "Events with 99 or fewer competitors must have at most **two** subsequent rounds" etc.

Laura-O:
Yes, sure... no idea how that happened. Thanks for pointing out.

wca-regulations.md
```diff
@@ -256,9 +256,9 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 - 9k) All competitors may participate in all events of a competition, except in cases specifically approved by the Board.
 - 9l) Each round must be completed before any following round of the same event starts.
 - 9m) Events must have at most four rounds.
-    - 9m1) Events with 99 or fewer competitors must have at most three rounds.
-    - 9m2) Events with 15 or fewer competitors must have at most two rounds.
-    - 9m3) Events with 7 or fewer competitors must have at most one round.
+    - 9m1) Events with 99 or fewer competitors must have at most two subsequent rounds.
```
viroulep: Shouldn't "events" be turned into "rounds" ?
The aim is to prevent scenario like 30->6->4 right ? With this wording the event still has 30 people, and this scenario is still legal.

lgarron:
What was wrong with the old version?

wca-regulations.md
Laura-O: Yes.... I should have added the change via copy & paste or I shouldn't propose changes when I am tired.
I hope everything is reasonable now.

Laura-O:
@lgarron You are asking for the reason for this change, right?
See #314, please.

wca-regulations.md
viroulep: Np, that's what reviews are for.

viroulep:
Looks good now :)

lgarron:
> @lgarron You are asking for the reason for this change, right?
> See #314, please.

Thanks. fda6f698c0833388562e833e8f2afb03b48c9494 makes things better.

Issue #343, opened 2016-03-07
Extends A3d1 to "anywhere but the mat"


viroulep:
Right now A3d1 states : 
`A3d1) The puzzle must not rest on the timer. Penalty: time penalty (+2 seconds).`

Following this thread : https://groups.google.com/forum/#!topic/wca-delegates/K0CBpf-wdTY, I believe we should extend this to handle the case where the competitor put the puzzle anywhere but on the mat (like in this video : https://youtu.be/7E6ZfCoYplA?t=163).

What do you think ?

SAuroux:
But this is handled already by:

A3d) At the end of the inspection, the competitor places the puzzle on the mat, in any orientation.

What could be clarified however is, that the penalty also applies if the puzzle is put anywhere else.

viroulep:
I must have expressed myself poorly, your suggestion is exactly what I intended to suggest : literally changing "on the timer" to "anywhere but on the mat".

Claster:
Do we want only to give +2 to such solves, or disqualify them completely?

pedrosino:
I believe DNFing is better.

Laura-O:
For Clock a penalty is defined in F3). Although the main intention of this regulation is probably to penalize changing the pin positions, it also mentions that the puzzle has to be placed on the mat. Therefore, in Clock this would be a DNF:
_F3) At the end of the inspection period, the competitor places the puzzle onto the mat in a standing position. The competitor must not change the positions of any pins from their scrambled positions before the beginning of the solve. Penalty: disqualification of the attempt (DNF)._

SAuroux:
Why is that? +2 is absolutely consistent with A3d1.

2016-03-08 12:27 GMT+01:00 Pedro Santos notifications@github.com:

pedrosino:
@Laura-O I didn't realise that part. It is weird that only clock has this "almost-well-defined" DNF and the other puzzles only have A3d1.
( @SAuroux maybe I subconciously knew this and wanted everything to be DNF)

SAuroux:
Haha, Laura is techniocally correct, even though this certainly isn't the
intention of F3. The empesis here is on the standing position, not on the
mat.

2016-03-08 13:22 GMT+01:00 Pedro Santos notifications@github.com:

pedrosino:
[...]

PR #347, opened 2016-03-22
+2 for putting puzzle anywhere outside the mat. Also clarify that F3 doesn't override it. #343

Claster:
The change clarifies a penalty for a puzzle being put anywhere outside the mat (including in front of the timer).
Also now we have to clarify F3, otherwise it unintentionally overrides this effect for clock only.

Should we clarify that a DNF in F3 must be issued if the puzzle is not in a standing posiiton? If a competitor puts the puzzle lying, pins likely will be changed anyway, unless
1) all pins are in the same position already;
2) the puzzle has really hard to press pins.
Do we care about these two cases?

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -495,7 +495,8 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 
 - F1) Standard speed solving procedures are followed, as described in [Article A](regulations:article:A) (Speed Solving). Additional regulations that supersede the corresponding procedures in [Article A](regulations:article:A) are described below.
 - F2) The judge places the scrambled puzzle onto the mat in a standing position.
-- F3) At the end of the inspection period, the competitor places the puzzle in a standing position. The competitor must not change the positions of any pins from their scrambled positions before the beginning of the solve. Penalty for not putting the puzzle in a standing position, or for changing the positions of any pins: disqualification of the attempt (DNF).
+- F3) At the end of the inspection period, the competitor places the puzzle in a standing position. Penalty: disqualification of the attempt (DNF).
+    - F3a) The competitor must not change the positions of any pins from their scrambled positions before the beginning of the solve. Penalty: disqualification of the attempt (DNF).
 
 
 ## <article-H><multiple-blindfolded><multipleblindfoldedsolving> Article H: Multiple Blindfolded Solving
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -59,7 +59,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 1g1) A score taker is responsible for compiling results.
     - 1g2) Changes to the result on a score sheet are only permitted at the discretion of the WCA Delegate.
 - 1h) Competitors in the same round of an event may be divided into groups.
-    - 1h1) Scramblers and judges for a round should not scramble for/judge competitors in their own group before they have finished all of their attempts for the round. They may scramble for/judge competitors in their own group at the discretion of the WCA Delegate, but the organisation team must ensure that scramblers and judges cannot see any scrambles for their attempts that they have not attempted yet.
+    - 1h1) Scramblers and judges for a round should not scramble for/judge competitors in their own group before they have finished all of their attempts for the round. They may scramble for/judge competitors in their own group at the discretion of the WCA Delegate, but the organisation team must ensure that scramblers and judges cannot see any scramble sequences for their attempts that they have not attempted yet.
 - 1j) All officials may compete in the competition.
 - 1k) Officials may serve multiple roles (e.g. organisation team, WCA Delegate, judge, score taker, scrambler).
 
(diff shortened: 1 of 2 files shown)


lgarron:
> Also now we have to clarify F3, otherwise it unintentionally overrides this effect for clock only.

Another reason I prefer DNF. Barely anyone does it, and trying to be lenient just gives us complications. :-/

> Do we care about these two cases?

I think that's fine, but as I just said above, I would prefer to avoid the potential for edge cases. We don't need more fodder for Regulations trivia. :-P

Laura-O:
I don't have a good solution for this, but the cases you mentioned are difficult to notice by judges and I think they should concentrate on their core tasks, instead of trying to follow the pin state of the scramble in case someone wants to put his Clock flat on the table.

Claster:
Indeed, having judges figured out whether all pins were in one state and whether lying on a table was legitimate or not is too overwhelming. PTAL.

@lgarron: I would prefer DNF also, but the case this issue was motivated by seems to be resolved as +2 only.

lgarron:
How do other WRC members feel about DNF vs. +2?

It's possible to rule one way on the existing case (based on the ambiguity in the current Regulations) but then tighten up the Regulations to be more clear in the future.

Laura-O:
I don't have a strong preference.
+2 would be in accordance with the penalty for putting the puzzle on the 
timer, but you could also say that the timer is part of the stackmat, so 
the penalty for putting it not on the mat should be stricter.

Claster:
So lets proceed with +2, according to the Board.

wca-regulations.md
```diff
@@ -378,8 +378,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - A3c2) If the parts of the puzzle are not fully aligned, then the competitor may align the faces, as long as misalignments stay within the limits of [Regulation 10f](regulations:regulation:10f).
         - A3c3) The competitor may reset the timer before the start of the solve.
         - A3c4) For Square-1, if the use of a thin object inside the puzzle has been enforced by the organisation team (see [Regulation A2b1](regulations:regulation:A2b1)), the competitor may remove the object from the puzzle during inspection.
-    - A3d) At the end of the inspection, the competitor places the puzzle on the mat, in any orientation.
-        - A3d1) The puzzle must not rest on the timer. Penalty: time penalty (+2 seconds).
+    - A3d) At the end of the inspection, the competitor places the puzzle on the mat, in any orientation. Penalty for placing it not on the mat: time penalty (+2 seconds).
```
viroulep: What about not changing A3d, and changing A3d1 to : 
"The puzzle must not rest outside the mat. Penalty: time penalty (+2 seconds)."
?

wca-regulations.md
```diff
@@ -496,7 +495,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 
 - F1) Standard speed solving procedures are followed, as described in [Article A](regulations:article:A) (Speed Solving). Additional regulations that supersede the corresponding procedures in [Article A](regulations:article:A) are described below.
 - F2) The judge places the scrambled puzzle onto the mat in a standing position.
-- F3) At the end of the inspection period, the competitor places the puzzle onto the mat in a standing position. The competitor must not change the positions of any pins from their scrambled positions before the beginning of the solve. Penalty: disqualification of the attempt (DNF).
+- F3) At the end of the inspection period, the competitor places the puzzle onto the mat in a standing position. The competitor must not change the positions of any pins from their scrambled positions before the beginning of the solve. Penalty for putting the puzzle not in a standing position, or for changing the positions of any pins: disqualification of the attempt (DNF).
```
viroulep: Since the "onto the mat" part is already specified by A3d, couldn't we just remove it and keep only the "standing position" part ?
Like this : 
"At the end of the inspection period, the competitor places the puzzle in a standing position."
(and keeping as is the remaining part of the regulation)

wca-regulations.md
Claster: Well, that becomes a bit repetitive: first we say "the competitor places the puzzle on the mat" and then "The puzzle must not rest outside the mat".

wca-regulations.md
Claster: Indeed, that is shorter, thanks!

wca-regulations.md
viroulep: Hmm you're right, maybe just change "not on" to "outside" in your original proposition ?
The "not on" actually sounds weird, maybe that's just me :p

Claster: Well, I don't have strong preferences towards "not on" or "outside". What others think?

Laura-O: Avoiding "not" is always a good idea as it makes things clearer and people cannot say they skipped the "not". So I vote for "outside".

Claster:
I changed "not on" to "outside" and reworded F3. I think that "Penalty: DNF" is not enough in F3, as it may be not clear that DNF is also for not standing position (the only word "must" in this regulation concerns changing of pins). So I kept "Penalty for blablabla or blablabla: DNF". PTAL.

wca-regulations.md
```diff
@@ -496,7 +495,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 
 - F1) Standard speed solving procedures are followed, as described in [Article A](regulations:article:A) (Speed Solving). Additional regulations that supersede the corresponding procedures in [Article A](regulations:article:A) are described below.
 - F2) The judge places the scrambled puzzle onto the mat in a standing position.
-- F3) At the end of the inspection period, the competitor places the puzzle onto the mat in a standing position. The competitor must not change the positions of any pins from their scrambled positions before the beginning of the solve. Penalty: disqualification of the attempt (DNF).
+- F3) At the end of the inspection period, the competitor places the puzzle in a standing position. The competitor must not change the positions of any pins from their scrambled positions before the beginning of the solve. Penalty for putting the puzzle not in a standing position, or for changing the positions of any pins: disqualification of the attempt (DNF).
```
viroulep: Sorry for being picky about this, but shouldn't we write "not doing something this way" instead of "doing something not this way" ?
I think "Penalty for not putting the puzzle in a standing position" would sound better, but I'm not sure.

viroulep:
This looks good :)
We just need to clear that small langage nitpick I commented on and I'm unsure about.

I'm going to make the same purely technical comment I made to Laura in #348 ;D
The commits history is starting to get a bit messy for a "small" resulting diff, would you mind [_squashing_](https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/) your 4 commits into one, and `git push --force` it to `fix-issue-343` ?

wca-regulations.md
lgarron: "Penalty for not putting the puzzle in a standing position" definitely sounds better to me.

lgarron: Alternatively, split up F3 into F3 (standing) and F3a (no pin changes)?

Claster:
I changed the order of "not".

Philippe, I wouldn't mind, but unfortunately I can squash it only on Sunday evening.

Claster:
Wo-hoo, squash was successfull! Merge it?

viroulep:
Nice :D 

I would merge it, just to check : did you see Lucas' comment about alternatively splitting F3 into F3+F3a ?

Claster:
Yes, I did. I don't have strong preference towards changing it (and Lucas seemingly doesn't, as he proposed it as an alternative). However, if anybody thinks that F3 is overloaded now, I could split.

lgarron:
I believe simplicity is important, and each Regulation should be as clear as possible on its own. If no one feels strongly, I'd prefer to split.

Claster:
All right, now it is simpler, as the condition for penalty is not written.

Laura-O:
Yes, that looks good.

I just merged #350 where "beginning of the solve" was changed to "start of the solve", so this branch has conflicts now. We have to change this here as well and everything should be fine.

PR #348, opened 2016-03-22
Fixes #344

### LINKED ISSUE #344: E2e): FMC solutions similar or identical with the inverse scramble
After the recent discussions about Fewest Moves solutions being identical or similar with the inverse scramble, several people have suggested a clarification or modification of [E2e](https://www.worldcubeassociation.org/regulations/#E2e):

> E2e) The competitor's solution must not be directly derived from any part of the scrambling algorithm. Penalty: disqualification of the attempt (DNF).
> E2e1) The WCA Delegate may ask the competitor to explain the purpose of each move in their solution, irrespective of scrambling algorithm. If the competitor cannot give a valid explanation, the attempt is disqualified (DNF).

At the moment, I can't think of a way how we could specify which solutions can be accepted and which should be declined. For sure, it would be possible to define some rules, but these would include an advanced analysis and understanding of the solution, which means that every Delegate would need to have an advanced knowledge of FMC techniques.

However, I think one of the main problems with this regulation is the second part of E2e1), i.e. an attempt is disqualified, if the competitor cannot give a valid explanation. Many seem to interpret that this also means that a solution is valid if it can be explained, although this is a logical fallacy:
1. Every solution, which is derived from a scramble, is a DNF.
2. Every solution, which cannot be explained by the competitor, is a DNF.

If a solution is equal or similar to the inverse scramble and the competitor is able to explain his solution, only 1. (i.e. E2e)) is relevant, which means we have to decide whether a solution is derived from the scramble or not. Similarly, E2e) does not mention any intention by the competitor, so we do not have to decide whether a competitor is trustful or not or if he is capable of finding this solution.

I don't know if this is a regional phenomenon, but this is obviously another regulation which is commonly interpreted in a different way than intended. Many competitors believe that their solution is valid as long as they are able to explain it, which is definitely wrong.

An additional problem I see is the word "derived". I am not a native speaker, but "derive" sounds rather vague to me. It has also been discussed before, how FMC techniques like NISS/switching/premoves make use of the scramble, so that the solution is actually derived from the scramble when these techniques are used. What we don't want is a relation between the solution and the scramble, i.e. using the inverse scramble or parts of the scramble for the solution.

Would it improve the future application of E2e to replace "directly derived" with "related"?

> E2e) The competitor's solution must not be related to any part of the scrambling algorithm. Penalty: disqualification of the attempt (DNF).

Further sources:
- [Discussion on speedsolving.com, 2014](https://www.speedsolving.com/forum/showthread.php?13599-The-FMC-thread&p=940511&viewfull=1#post940511)
- [Another discussion, 2010](https://www.speedsolving.com/forum/showthread.php?13599-The-FMC-thread&p=546070&viewfull=1#post546070)

Laura-O:
See #344 for a longer explanation.

These changes clarify that E2e can also be relevant when there is no proof that a competitor used the scramble sequence to derive his solution intentionally.
With the new Fewest Moves scramble sequences implemented in TNoodle, starting a solution with the inverse of more than the last 4 moves of the scramble is a good evidence that the solution was derived from the scramble as the first and the last three moves of the scramble sequence are pointless. However, this is still a _"should"_, so in the rare case that beginning a solution with these moves is reasonable, this is not necessarily a DNF.

**Question**
Is "solutions beginning with more than 4 moves of the inverse scramble sequence" clear?
It can be interpreted that solutions should not begin with any 4 moves of the inverse scramble. But I have no idea how to clarify that here.

### DIFF
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -191,6 +191,8 @@ To be more informative, each Guideline is classified using one of the following 
 - E2b+) [CLARIFICATION] A competitor may choose to stop an attempt early by handing in a solution before the time limit.
 - E2c+) [CLARIFICATION] Although the judge may provide a standard sheet for competitors to submit solutions, a competitor may submit a solution on a different piece of paper, as long as a single move sequence is unambiguously indicated as the submitted solution. (Note that the piece of paper must come from the judge, according to [Regulation E3a](regulations:regulation:E3a).)
 - E2c++) [CLARIFICATION] The competitor's solution must only use moves that are exactly defined in [Regulation 12a](regulations:regulation:12a). Examples of notation and moves that are not permitted in competition: [F], [R, U], [R: U], R'2, L'w, f, M, U2'.
+- E2e+) [CLARIFICATION] Competitors must not derive solutions from any part of the scramble sequence, and solutions should not share significant parts with the inverse scramble sequence.
+- E2e++) [EXAMPLE] Example of solutions that should result in a disqualification of the attempt (DNF): solutions beginning with the same 4 or more moves as the inverse scramble sequence.
 - E3b+) [REMINDER] "Rubik's Cube" refers only to the standard 3x3x3 puzzle.
 - E3d+) [CLARIFICATION] Any stopwatch or watch used by a competitor must not appear to have any functions that would help the competitor find a solution.
 - E3d++) [CLARIFICATION] Competitors should not consider a personal stopwatch or watch as the official time, and must submit their solution after the judge calls "STOP".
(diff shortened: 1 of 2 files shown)


viroulep:
I like how the final changes look.

Purely technical comment : the commits history is starting to get a bit messy for a "small" resulting diff, would you mind [_squashing_](https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/) your 6 commits into one, and `git push --force` it to `fixes-FMC` ?

PS: PRs are exactly the places where you want to ignore the note at the end of the linked article :p

Laura-O:
Good idea! :-)

Claster:
Looks good to me.
I like how it says "example of solutions that SHOULD result in a DNF". That is, if there is a very obvious block that uses 4 moves from scramble, it could be kept, at the discretion of a delegate. I just hope that competitors and delegates won't interpret this word "should" as "must".

PR #350, opened 2016-03-26
Some minor fixes to follow the style guide, fixes #339

### LINKED ISSUE #339: Unify start/begin and stop/end
Right now the [style guide](https://github.com/cubing/worldcubeassociation.org/blob/master/regulations/style-guide.md) states:

> - "Start" instead of "begin".
> - ...
> - "Start" and "stop" (verbs).
> - "Beginning" and "end" (nouns).

The Regulations follow this, although the Guidelines use "begin" instead of "start". (There are also a few occurrences of "finish", but these seem appropriate.)
It would be nice to simplify this and use fewer words. However:
- DNS stands for "Did Not **Start**", so we should continue to say that the competitor "starts the attempt".
- There is no natural noun form of "stop" – "the stop of the solve/attempt/round" is wrong.

Proposal: Start by changing "begin"/"beginning" to "start".

Laura-O:
These are some fixes following the style guide.

Specifically:
- begin -> start
- scramble -> scramble sequence
- allowed -> permitted

Closes #339

### DIFF
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -88,7 +88,7 @@ To be more informative, each Guideline is classified using one of the following 
 - 5b5++) [EXAMPLE] Examples of parts physically detached, resulting in a "DNF": two 3x3x3 center caps, two big cube center pieces, a 3x3x3 edge piece, a 4x4x4 edge piece ("wing"), any combination of two pieces that both have coloured parts.
 - 5b5+++) [EXAMPLE] Example of parts not fully attached, resulting in a solved state: a 5x5x5 centre piece twisted in its spot.
 - 5b5++++) [EXAMPLE] Example of parts physically attached but not fully placed, resulting in a "DNF": a 3x3x3 edge piece slightly popped out.
-- 5c+) [REMINDER] If a competitor has a puzzle defect, this does not give them the right to an extra attempt.
+- 5c+) [REMINDER] If a competitor has a puzzle defect, this does not grant them the right to an extra attempt.
 
 
 ## <article-6><awards><awards> Article 6: Awards/prizes/honours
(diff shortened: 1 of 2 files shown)


wca-guidelines.md
```diff
@@ -161,9 +161,9 @@ To be more informative, each Guideline is classified using one of the following
...
-- A7g++) [CLARIFICATION] Individual penalties cannot be waived for new competitors. A full extra attempt must be given.
-- A7g+++) [RECOMMENDATION] The WCA Delegate should use stronger discretion when awarding more than one extra attempt to the same new competitor. If more than one extra attempt is awarded, it should be for a different incident.
+- A7g+) [EXAMPLE] Examples of incidents that could grant a new competitor an extra attempt: not waiting for the green light, improperly starting or stopping the timer, going over inspection time due to misunderstanding how to start an attempt, or other procedural time penalties.
+- A7g++) [CLARIFICATION] Individual penalties cannot be waived for new competitors. A full extra attempt must be granted.
+- A7g+++) [RECOMMENDATION] The WCA Delegate should use stronger discretion when granting more than one extra attempt to the same new competitor. If more than one extra attempt is graned, it should be for a different incident.
```
viroulep: A typo slipped in ! "Graned" instead of granted.

Laura-O: Thanks for checking so carefully. I'm incompetent...

viroulep:
I just pushed another standardization pass to this PR.
Mainly, I implemented Lucas' suggestion to also use "start" as a noun instead of "beginning". The result looks fine, let me know what you think.

I also went through all the style guide and substituted the words by their recommended alternative.
There is a bunch of "give" uses left, but I don't think it's appropriate to use "grant" in their context.

Laura-O:
I thought about changing these "start" mentions as well, but I didn't because they are rather general descriptions and not concerning the actual solving procedure.
However, that was just a spontaneous decision, I'm completely fine with these changes.

Issue #49, opened 2013-07-27
Remove Qualification Rounds?


lgarron:
Ron asked us not to remove them for 2013, but they're hardly ever used.

---

[Delegate email, December 08, 2013](https://groups.google.com/forum/#!topic/wca-delegates/iue2cPOSc0U)

lgarron:
# Proposal (December 07, 2013)

**Remove qualification rounds.**
View exact proposed changes: https://github.com/cubing/wca-documents/compare/qualification-rounds

## Pros

Removing qualifications rounds would:
- Simplify the Regulations a bit (we can remove all of [9r](https://www.worldcubeassociation.org/regulations/#9r), and I'm always a fan of removing unneeded Regulations).
- Make all events to be consistent: every competitor an event proceeds through rounds in the same way.
- Not really change how competitions are run: Qualification rounds have only been used at two competitions since 2010.
  - [AsianChampionship2012](https://www.worldcubeassociation.org/results/c.php?i=AsianChampionship2012): qualification rounds could have been replaced with combined finals .
  - [RabbitAndTurtle2013](https://www.worldcubeassociation.org/results/c.php?i=RabbitAndTurtle2013): the qualification round for 3x3x3 was basically a first round that some fast competitors skipped.

## Cons

As far as I understand, the main reasons for a qualification round are:
- A qualification rounds round allows more competitors to participate in an event.
- A qualification round allows new competitors a chance to participate in large championship (that may require pre-qualification otherwise). See [9r+](https://www.worldcubeassociation.org/regulations/guidelines.html#9r+)

But these competitors can just be added to the first round. If there isn't enough time to give slow competitors all their attempts, the first round can be turned into a combined round. Thus, _qualification rounds, in their current format, don't really save time versus other options_ (and therefore, don't really allow more competitors). The 2013 World Championships ran fine with 580 competitors and no qualification rounds, and [very few competitions have had even 1/3 as many competitors](https://www.worldcubeassociation.org/results/statistics.php#most_persons).

Another possible argument:
- A qualification rounds allows (once-)fast competitors guaranteed entry into the first round, and extra solves for newcomers.

However, I don't think anyone would argue that there is a significant need for this.

okayamat:
I'd like to comment on the competitions I participated in.
- AsianChampionship2012: qualification rounds could have been replaced with combined finals .

The qualification rounds were held on Friday, and the first rounds were held on Saturday.
[...]

PR #352, opened 2016-03-30
Removes 8a4+.

### LINKED ISSUE #332: 8a4+ refers to an unexisting regulation
8a4+ refers to qualification rounds that were removed with #49, I think we should simply remove this guideline, right ?

(I'm also going to assume we should follow the same policy as the Regulations and not rename the 8a4++ guideline)

viroulep:
This guideline is related to qualification rounds which were removed with #49.
Fixes #332.


Laura-O:
Well, I think there is not much to discuss here. :-)

PR #353, opened 2016-03-31
DNS in rounds with cumulative time limits

### LINKED ISSUE #303: Results for rounds with cumulative time limits
We had some discussions about the results of attempts done in rounds with a cumulative time limit, particularly results for attempts done (or better to be done) after the time limit has been reached. This case is not mentioned in the current regulations.
So, if the cumulative time limit is 10 minutes in a "Best of 3"-format and the competitor has done 2 solves in 10 minutes (e.g. each solve 5 minutes, so the time limit has been reached), is the third solve not recorded at all, a DNS or a DNF?

From my point of view, a DNF would be the right decision:
- Leaving the result blank would change the format of the round. If the format is "Best of 3", recording only two results would change it to "Best of 2". As every competitor is allotted the number of attempts defined by the round format, having no results for one or more attempts is not possible.
- A DNS doesn't seem to be appropriate either, as 9f5) says "The result of an attempt is recorded as DNS (Did Not Start) if the competitor is eligible for an attempt but declines it.". In this case, the competitor is not eligible for an attempt, so it can't be a DNS.
- A DNF seems to be the most reasonable result for me, although it's a bit strange. According to A1a4) the judge stops an attempt when the competitor reaches the time limit of a solve/round and the result is DNF. In this case, the time limit would be reached after 0 seconds, so the judge would literally stop him even before the attempt has begun.

I suggest to add a way to handle these cases to A1a2). Any thoughts or ideas?

Laura-O:
Fixes #303


viroulep:
Looks good !

PR #363, opened 2016-05-05
Add the style guide to the Regulations repository.

lgarron:
@Laura-O, does this sound alright to you? I think here makes the most sense.


Laura-O:
Yes, that's a good idea!

Issue #236, opened 2015-01-06
Instructions/Guides


lgarron:
I've become convinced that we will never be able to slim down the Regulations. With more and more competitions all over the world each weekend, we find the need to standardize on every possible detail.
However, I think what we _can_ do is provide an instruction sheet for each role that gives them the right ideas. (I don't believed a [simplified version of the regular Regulations](https://github.com/cubing/wca-documents/issues/69) is the right way to go.) These would ideally be one page per role, and have pictures with good cues. We should make them available as HTML on the website, with PDF downloads available.

The following roles should have such guides:
- Competitor
- Judge
- Scrambler
- Score Taker
- Audience Member (no flash, keep distance from competitors, don't pick up pieces)
- [Equipment setup](https://github.com/cubing/wca-documents/issues/235)?

I think we should also have one that covers what to do for various incidents (e.g. inspecting too long, pop, corner twist, misalignment, touching/dropping the cube during various parts of the attempt).

If these turn out to work well, we can adapt [2t)](https://www.worldcubeassociation.org/regulations/#2t) to emphasize knowing these guides (while still accepting responsibility to follow the Regulations).
We should also make sure that they are easy to translate (and make translations available).

Various things that should go in these guides, that I might otherwise forget:
- Cutoffs

lgarron:
Here is a rough concept for the competitor guide I made a few weeks ago:

![2014-12-21 01 38 31](https://cloud.githubusercontent.com/assets/248078/5628403/77726728-955b-11e4-822d-7716ba5a47ff.jpg)

A few flaws:
- No diagram showing you can hold the cube during inspection.
- No speech bubbles for 8 and 12 seconds (the diagrams should be as standalone as possible, for clarity and internationalization).
- No explanation that the competitor needs to put their hands on the timer well ahead of 15 seconds.
- Doesn't remind competitors _not to talk about scrambles_.
- Doesn't tell them about handling incident. (Always [call the Delegate](https://www.speedsolving.com/forum/showthread.php?51857-Why-did-I-get-a-DNF&p=1062050#post1062050) - see [11e](https://www.worldcubeassociation.org/regulations/#11e).)

lgarron:
[Suggestion from Natán](https://groups.google.com/forum/#!msg/wca-delegates/m92AyELlPFw/jAIZf4oCK-4J):

[...]

PR #385, opened 2016-09-28
A6e: Explicitly reference A7a to make it clear what counts as inspected.

lgarron:
I have a long-standing personal note that I just came across to clarify A6e a bit.

In the real world, applications of A7e are somewhat fuzzy, so I think it's worth stating exactly what counts as "inspected".


HippolyteM:
Hi,
Maybe the regulations should clearly specify what is already done most of the time (when cases are tricky), and choose a word, as "ok", which would be part of the judging routine and that the judge should always say before the competitor touch his cube.

lgarron:
> Maybe the regulations should clearly specify what is already done most of the time (when cases are tricky), and choose a word, as "ok", which would be part of the judging routine and that the judge should always say before the competitor touch his cube.

They do. ;-)

In fact, that's exactly what A7a says:

> - A7a) The judge tells the competitor the result.
>   - A7a1) If the judge finds that the puzzle is solved, they call "OKAY".
>   - A7a2) If the judge assigns any penalties, they call "PENALTY".
>   - A7a3) If the result is DNF, the judge calls "DNF".

I don't know if I used to do this consistently, but I noticed that [Olivér does this very well](https://youtu.be/ss-D6YB_Lew?t=46s), and make sure to do it every time now.
I think it's a small but important detail. We should include it in guides (#236), although perhaps we don't need to call too much attention to it.

Laura-O:
Same here. I used to make this calls very inconsistently in the past, but started to do it consistently about 1.5 years ago. Competitors here are definitely not used to these calls. I remember a few situations when competitors asked me something like "What do you mean?" when I said "OK". :stuck_out_tongue:

This is not relevant for this pull request, but related: I just noticed that the word "result" in [A7a)](https://www.worldcubeassociation.org/regulations/#A7a) might be not well-chosen. Someone correct me if I miss something, but according to the regulations a result can be the original recorded time, the recorded time with penalties, "DNF" and "DNS".
[A7a)](https://www.worldcubeassociation.org/regulations/#A7a) suggests that the calls are also a result. I do not have a good suggestion how to replace it, maybe "outcome"? Or "judgement"?

HippolyteM:
I am ashamed, I didn't remember that we already had an exact wording.:/
Then, what about adding to A7a : "After the judge's call, and if the competitor agree with it, the puzzle is considered inspected."
And A6e could refer to this rule. (Even if, I agree, it would be better to find a way to define "inspected" before its first use.

As for Laura's question, I also think there is something to add, as it is not clear what we exactly want the judge to tell. According to (A7a to A7c), it is indeed the judgment, but then there is no need for the judge to tell the actual time (recorded time if okay, recorded time plus penalties if needed), which is quite important, I think we all agree on that.
Here is a suggestion:
A7a) The judge tells the competitor their judgment:
    A7a1) If the judge finds that the puzzle is solved, they call "OKAY".
    A7a2) If the judge assigns any penalties, they call "PENALTY".
    A7a3) If the result is DNF, the judge calls "DNF".
A7b) The judge tells the competitor the final time. They specify the time displayed on the timer, the exact penalties, and the final score.
(or more simple:
A7b) The judge tell the competitor exactly what he will write on the score sheet)

Hippolyte

Laura-O:
I just discussed this with @AlbertoPdRF and we both think that this is a good idea, but it will not work in competitions.

Even when judges are taught to say the result after an attempt, this is still done very inconsistently. I also think competitors are used to this today. Therefore, if we add this regulation and apply it in a strict way, we would see a lot of DNFs.

We can continue to discuss this, but I am closing this for now as there hasn't been any activity since more than a year here.

PR #386, opened 2016-09-28
A4b) Change a slightly ambgiguous "them" to explicitly reference the competitor.

lgarron:
@keemy pointed out to me that "them" can otherwise be interpreted as "palms", which would not carry the intended meaning. (We don't mean to say that the left palm must be on the left side and the right palm must be on the right side.)

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -383,7 +383,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - A3d2) When 8 seconds of inspection have elapsed, the judge calls "8 SECONDS".
         - A3d3) When 12 seconds of inspection have elapsed, the judge calls "12 SECONDS".
 - A4) Starting the solve:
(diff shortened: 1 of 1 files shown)


jfly:
Haha, so I've been misreading this regulation. What would it mean for me to put my palms on the side of the timer that is not closer to me?

lgarron:
> What would it mean for me to put my palms on the side of the timer that is not closer to me?

Uh, it should mean the opposite, I think?

jfly:
> Uh, it should mean the opposite, I think?

I'm so confused I don't even know how to respond to that. Maybe let's try another approach? What is an example of something that `A4b` does not allow?

lgarron:
Oh, sorry, I read an implicit comma after "mean".

> What is an example of something that A4b does not allow?

Putting your palms on the side of the timer that is not closer to you. ;-)

lgarron:
> Putting your palms on the side of the timer that is not closer to you. ;-)

But joking aside, the main use of A4b in practice is to prevent competitors from putting their palms _on the timer itself_.

@keemy's reading is a possible loophole that lets you claim that A4b was saying something else instead.

PR #387, opened 2016-09-28
3j++) Change "identifiable" to "distinguishable" (more exact).

lgarron:
"identifiable" is a slightly more general word; I think it's better to use a word with a more exact meaning.


lgarron:
@Laura-O, I believe you wrote the current wording. Is this refinement alright with you?

PR #395, opened 2016-11-01
Updated the latest textual revision.

viroulep:
As per @lgarron's suggestion:

> We should follow the old convention and add "Last textual revision":
> https://www.worldcubeassociation.org/regulations/history/files/regulations2010.html
>

@jfly's fixes for contact link and a typo have actually been up since they were merged in `draft`, so we may as well just add this and merge current `draft` in `official`.

### DIFF
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -5,6 +5,8 @@
 
 ## Notes
 
+Latest textual revision: December 12, 2016
+
 ### WCA Regulations
 
 The WCA Guidelines supplement the [WCA Regulations](regulations:top). Please see the Regulations for more information about the WCA.
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -5,6 +5,8 @@
 
 ## Notes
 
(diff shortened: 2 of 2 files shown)


viroulep:
I've created the "textual-revision" label, and tagged all the relevant PRs (open or closed) since `draft` diverged from `official`.

We just have to merge #398 and decide about if we want to do #397, and then I'm willing to update this one and merge the tagged PRs to both `draft` and `official`.

viroulep:
Unless someone objects by Monday, I'll merge [all the PRs with tag `textual-revision`](https://github.com/thewca/wca-regulations/issues?q=label%3Atextual-revision) into `official` and deploy them along with the new TNoodle version.

lgarron:
If we're going to merge #398, I don't think this should count as just a textual revision.
I think we should make it a full revision, and make (small email, non-stickied post) announcements.

viroulep:
Right it does change something, I've removed the label from this one!

PR #396, opened 2016-11-08
Tweaked language to allow for multiple official versions of the scrambling software.

jfly:
The motivation for this change is that new TNoodle versions with new features always get delayed for months because we don't like to mark new releases of TNoodle as official if they don't change the way scrambles are generated. However, we wrote TNoodle and the WCA API with this problem in mind, and as far as all our software is concerend, it is possible to have multiple official versions of TNoodle.

The current language still implies that there is only one official scrambling software program... maybe in some future we'll have multiple offical scrambling softwares, but I think it would be best to address that when it comes.

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -161,7 +161,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 4d3) Square-1 is scrambled starting with the darker colour on front (out of the 2 possible scrambling orientations).
     - 4d4) Clock is scrambled starting with either side in front and 12 o'clock pointing up.
     - 4d5) Skewb is scrambled starting with the white face (if not possible, then the lightest face) on top and the green face (if not possible, then the darkest adjacent face) on the front-left.
-- 4f) Competition scramble sequences must be generated using the current official version of the official WCA scramble program (available [via the WCA website](https://www.worldcubeassociation.org/regulations/scrambles/)).
+- 4f) Competition scramble sequences must be generated using a current official version of the official WCA scramble program (available [via the WCA website](https://www.worldcubeassociation.org/regulations/scrambles/)).
 - 4g) After scrambling a puzzle, the scrambler must verify that the puzzle is scrambled correctly. If the puzzle state is wrong, the scrambler must correct it (e.g. by solving the puzzle and applying the scramble sequence again).
     - 4g1) Exception: For the 6x6x6 Cube, 7x7x7 Cube, and Megaminx, it is not necessary to correct the puzzle state, at the discretion of the WCA Delegate.
 


lgarron:
> The motivation for this change is that new TNoodle versions with new features always get delayed for months because we don't like to release new versions of TNoodle that don't change the way scrambles are generated.

If by "don't like" you mean "it's literally not allowed for Delegates to use those". ;-)

I think it's perfectly reasonable to allow multiple official scramble programs, but all of them must comply with the Regulations.

wca-regulations.md
```diff
@@ -161,7 +161,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 4d3) Square-1 is scrambled starting with the darker colour on front (out of the 2 possible scrambling orientations).
     - 4d4) Clock is scrambled starting with either side in front and 12 o'clock pointing up.
     - 4d5) Skewb is scrambled starting with the white face (if not possible, then the lightest face) on top and the green face (if not possible, then the darkest adjacent face) on the front-left.
-- 4f) Competition scramble sequences must be generated using the current official version of the official WCA scramble program (available [via the WCA website](https://www.worldcubeassociation.org/regulations/scrambles/)).
+- 4f) Competition scramble sequences must be generated using one of the current official versions of the official WCA scramble program (available [via the WCA website](https://www.worldcubeassociation.org/regulations/scrambles/)).
```
lgarron: I would simplify this to "a current official version"

jfly: Done!

jfly:
> If by "don't like" you mean "it's literally not allowed for Delegates to use those". ;-)

Sorry, I abused the vocabulary here. Let me rephrase: "we don't like to make new versions of TNoodle official if they don't change the way scrambles are generated"

> I think it's perfectly reasonable to allow multiple official scramble programs, but all of them must comply with the Regulations.

I also have no problem with multiple official scramble programs, I just could not think of a good rephrasing that allows for this, so I gave up. Do you have a suggestion?

lgarron:
> I also have no problem with multiple official scramble programs, I just could not think of a good rephrasing that allows for this, so I gave up. Do you have a suggestion?

I don't think this needs further coverage. It seems obvious to me that Delegates should not be allowed to use scramble programs that don't follow the Regulations.

lgarron:
Anyhow, I'm happy to merge this. @Laura-O, @Claster, @viroulep can I get a second approval from one of you to merge this?

jfly:
> It seems obvious to me that Delegates should not be allowed to use scramble programs that don't follow the Regulations

I think that is clear, and is not the issue I'm trying to address. What is unclear is if there were multiple scramble programs that _do_ follow the regulations, would the wca regulations allow for declaring both of them official?

lgarron:
> if there were multiple scramble programs that do follow the regulations, would the wca regulations allow for declaring both of them official?

Oh, you mean different programs (like TNoodle and Mark 2) rather than different versions of the "same" program?
I don't see a problem with writing "a current official version of an official WCA scramble program"

Laura-O:
@lgarron: as already said in the mail thread: let's get this done. 👍

jfly:
Darn it! I just made some relevant changes.

jfly:
New PR incoming.

(edit: See #398)

lgarron:
[Reverted. Let's continue on #398.]

PR #397, opened 2016-11-08
Trim trailing whitespace from the regulations.

jfly:
Just opening this for discussion. I'd like to implement https://github.com/thewca/wca-regulations-compiler/issues/8 before we merge this up just yet.

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -7,15 +7,15 @@
 
 ### WCA Regulations and Guidelines
 
-The WCA Regulations contain the full set of Regulations that apply to all official competitions sanctioned by the World Cube Association.    
+The WCA Regulations contain the full set of Regulations that apply to all official competitions sanctioned by the World Cube Association.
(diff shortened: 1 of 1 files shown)


viroulep:
I'm curious about what other @thewca/wrc-team members think about that :)
I support the change: the 4 trailing whitespaces were there because pandoc implemented them as a `<br/>`. I think we are all used to GFM in which one `\n` is considered as a `<br/>` and `\n\n` as a paragraph break. thewca/wca-regulations-compiler#9 is ready to be merged with these changes (also making trailing whitespaces illegal) if we are willing to make them happen.

Note: we also need to update the translations, most of them have the same trailing whitespaces ;)

Laura-O:
I also support the change.
Trailing whitespaces are often confusing people and I don't see a reason why we should keep them.

lgarron:
I've been following the implementation comments casually, and I believe the plan is to treat single newlines as line breaks and double new lines as new paragraphs, right?

I've always thought that is a better default, and some Markdown implementations even use that instead of the standard behaviour.

PR #413, opened 2017-03-13
Change Article Z preamble to avoid ambiguity. Fixes #412.

### LINKED ISSUE #412: Z-regulations: ambiguous preamble
"The WCA Board must approve ..." can be misunderstood as if the Board may not disapprove any application of Z-regulations.

lgarron:


Laura-O (commented):
I like the idea in general, but I am not sure if this is 100% clear.
You could understand that the Board decides if you are allowed to apply any Z-regulation, so you do not need an approval for specific cases.

AlbertoPdRF:
That's true, but I can't come up with a solution other than adding ", at the discretion of the WCA Board." after every Zn).

Claster:
Organisation teams may adopt optional regulations to facilitate the administration of the competition.
They must get an approval from the WCA Board for every optional regulation for a competition.

AlbertoPdRF:
LGTM!

PR #414, opened 2017-03-13
Fix subject-verb agreement in 3d1b+.

lgarron:
@Laura-O, any objection if I merge grammar fixes like this into the `fixes` branch? I have a few lying around: https://github.com/thewca/wca-regulations/pulls

### DIFF
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -62,7 +62,7 @@ To be more informative, each Guideline is classified using one of the following 
 - 3a++) [ADDITION] By default, a competitor should use the same puzzle for consecutive attempts in a speed solving round. A competitor may switch puzzles between attempts, at the discretion of the judge or WCA Delegate.
 - 3a+++) [CLARIFICATION] Competitors may borrow puzzles from other competitors privately, to use in competition.
 - 3a1+) [CLARIFICATION] Competitors may be disqualified if they do not have a puzzle ready to submit soon after they are called (e.g. if they planned to use a puzzle that another competitor is currently using, and therefore cannot submit a puzzle at the moment).
-- 3d1b+) [CLARIFICATION] Patterns do not need to be present on every face, but only on faces where a distinct colour cannot be found. Pieces must not have any feature (e.g. textures, pattern irregularities) that significantly distinguishes them from similar pieces.
+- 3d1b+) [CLARIFICATION] Patterns do not need to be present on every face, but only on faces where a distinct colour cannot be found. Pieces must not have any features (e.g. textures, pattern irregularities) that significantly distinguish them from similar pieces.
 - 3h+) [CLARIFICATION] Puzzles may be refined internally by sanding or lubricating.
 - 3h++) [EXAMPLE] Examples of enhancements include: new moves are possible, normal moves are impossible, more pieces or faces are visible, colours on the backside of the puzzle are visible, moves are done automatically, or the puzzles has more/different solved states.
 - 3h2+) [CLARIFICATION] In the past, "stickerless" puzzles were not permitted. Such puzzles are now permitted.


viroulep:
If this is a kind of "textual revision" that should land in `official` too, I would propose doing it the opposite way: make/merge the change first against `official`, then merge it back into `draft`.

I ran into a similar situation when merging the previous purely textual changes, and realized that if you get it merged into `draft` first you:
- couldn't merge the same commit into `official` (you would bring all the parent commits in `draft`, which may bring in non-textual revision)
- would have to cherry-pick the commit on `official`, creating two commit with the same diff (wouldn't be an issue when merging `draft` for the next release though)

Hope my explanation makes sense!

lgarron:
I don't mind having non-critical textual revisions in `draft` or `fixes` for a while – that was the original purpose of `fixes`. However, I don't mind updating `official` more regularly as long as the website:

1) makes it easy for a non-technical user to get all official revision markers and their dates.
2) shows the correct date for the latest major change and the latest textual revision.

Does our current system make that easy?

Laura-O:
No objection.

> 2. shows the correct date for the latest major change and the latest textual revision.

We already have the date of the "Latest textual revision" on the website.

lgarron:
> We already have the date of the "Latest textual revision" on the website.

But if we push the official version with every small fix, that requires remembering to manually update the line with every commit, right?

Laura-O:
Yes, indeed, that's not very handy.
Does it make sense to open an issue for that in the website repository?

Laura-O:
Pushing this a bit... I have added *textual-revision* labels to all relevant pull requests.
Could you all please double-check if [these pull requests](https://github.com/thewca/wca-regulations/pulls?q=is%3Aopen+is%3Apr+label%3Atextual-revision) are just text fixes, which do not change anything?

If nobody contradicts, can we merge and deploy this in a few days, @viroulep @lgarron?

viroulep:
> Does it make sense to open an issue for that in the website repository?

I honestly don't know where this issue belongs...
The website doesn't know how the Regulations are generated, only the [wrc](https://github.com/thewca/wca-regulations-compiler) does. I'll try to come up with something clever for this.

> Could you all please double-check if these pull requests are just text fixes, which do not change anything?
If nobody contradicts, can we merge and deploy this in a few days, @viroulep @lgarron?

As I said I think textual-revision should be initially made against `official` and then ported to `draft` to avoid merging non textual-revision the commit merged could depend on.
(That's also logical: the purpose of a textual-revision is a small change of the current official version, regardless of what could be going on in the draft version)

For those willing to put a picture on why I'm insisting on it (and because sometimes a picture is better than a lot of words :p), let's look at this sample graph of commits:
git-before-merge

Assuming "F" is our current `official` branch, and "D" our current `draft` branch; "C" and "D" are brand new Regulations that should be part of a "full" release.
If a textual-revision change is made on top of "D" (let's call it "G"), *merging* the PR into `official`(F) will actually bring C and D!
A solution to this would be to *cherry-pick* G from draft on top of official, actually creating a second commit G' with the same changes (so we would have two different commits for the same changes, one on each branch).
This is a bit awkward, and making our textual-revision change on top of F would have made our life easier: we definitely don't mind merging changes that are already in `official` into `draft`! 

I quickly checked the [current diff between draft and official](https://github.com/thewca/wca-regulations/compare/official...draft?expand=1) and luckily it appears that we've only merge textual revisions so far in draft!
So in the current state we can just merge all these and merge draft in official.

AlbertoPdRF:
It seems like you've got a point here.

I think that we should merge all the PR (after revision) into draft, then draft into official, and from that point on start making the textual revisions changes on top of official, as you propose :)

viroulep:
(same here, last commit fixed conflicts)

PR #447, opened 2017-06-28
Change A7 to ensure the judge signs before the competitor.

### LINKED ISSUE #427: DNS for signing empty results after an attempt
Quoting @SAuroux:
>When reading Nikkhil's reply, I first thought "no way, this is only DNS if the competitor signs before starting the attempt".
>But reading A7c2, that paragraph really says that the situation where a competitor solves normally, but then signs before the judge writes down the time is a DNS as well. 
>Problem: This contradicts with the definitions of DNS and DNF:
>9f4) The result of an attempt is recorded as DNF (Did Not Finish) if the attempt is disqualified or unsolved/unfinished.
>9f5) The result of an attempt is recorded as DNS (Did Not Start) if the competitor is eligible for an attempt but declines it.

My initial though was that changing this to DNF wouldn't be much better as this would override the call of the judge. However, the situation is similar to a competitor resetting the timer before the judge has written down the result, so DNF should be fine.

### LINKED ISSUE #443: Missing signatures on score sheets
After some delegates asked for the consequences of competitors or judges not signing on the score sheets, the WRC has decided to add to Article A ([A7)](https://www.worldcubeassociation.org/regulations/#A7) specifically) of the Regulations something like:

> If the competitor and/or judge's signature is found to be missing after the score sheet has been delivered to the score taker, the result is considered a DNF. Exception: If the judge and/or competitor can be tracked down to vouch for the recorded time before the end of the competition day, then the recorded time may be considered valid, at the discretion of the WCA delegate.

pedrosino:
Fixes #443, fixes #427.

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -413,11 +413,11 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - A7a1) If the judge finds that the puzzle is solved, they call "OKAY".
         - A7a2) If the judge assigns any penalties, they call "PENALTY".
         - A7a3) If the result is DNF, the judge calls "DNF".
-    - A7b) The judge records the result on a score sheet.
+    - A7b) The judge records the result on a score sheet and signs next to the result.
         - A7b1) If penalties are assigned, the judge records the original recorded time displayed on the timer, along with any penalties. The format should be "X + T + Y = F", where X represents the sum of time penalties before/starting the solve, T represents the time displayed on the timer (the "original recorded time"), Y represents a sum of time penalties during/after the solve, and F represents the final result (e.g. 2 + 17.65 + 2 = 21.65). If X and/or Y is 0, the 0 terms are omitted (e.g. 17.65 + 2 = 19.65).
-    - A7c) The judge and competitor must each check the recorded result, and sign (or initial) the score sheet to acknowledge that the result is correct, complete, properly formatted, and clearly readable. This finishes the attempt.
+    - A7c) The competitor must check the recorded result, and sign (or initial) the score sheet to acknowledge that the result is correct, complete, properly formatted, and clearly readable. This finishes the attempt.
         - A7c1) If the competitor or the judge refuses to accept and sign the score sheet, the WCA Delegate must resolve the dispute.
-        - A7c2) The competitor must not sign an attempt on the score sheet before the judge has recorded the final result for that attempt. If a competitor signs (or otherwise marks) an attempt before they have begun it, or before the judge has finished recording the result, they forfeit the attempt and the result will be considered DNS.
+        - A7c2) The competitor must not sign an attempt on the score sheet before the judge has recorded the final result for that attempt and signed. If a competitor signs (or otherwise marks) an attempt before they have begun it, they forfeit the attempt and the result will be considered DNS. If the competitor signs (or otherwise marks) an attempt after starting the attempt but before the judge has finished recording the result, it will be considered DNF.
     - A7f) When a competitor's score sheet for a round is complete, the judge delivers the score sheet to the score taker.
     - A7g) At the discretion of the WCA Delegate, an incident or penalty caused by a new competitor's inexperience may be replaced with an extra attempt.
 


wca-regulations.md
```diff
@@ -413,11 +413,11 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - A7a1) If the judge finds that the puzzle is solved, they call "OKAY".
         - A7a2) If the judge assigns any penalties, they call "PENALTY".
         - A7a3) If the result is DNF, the judge calls "DNF".
-    - A7b) The judge records the result on a score sheet.
+    - A7b) The judge records the result on a score sheet and signs next to the result.
```
Laura-O, line 416: On the cubecomps scoresheets judges sign in the rightmost column, so it's not strictly "next to the result".
Maybe we can find a better wording for that?

wca-regulations.md
```diff
@@ -413,11 +413,11 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
...
+    - A7c) The competitor must check the recorded result, and sign (or initial) the score sheet to acknowledge that the result is correct, complete, properly formatted, and clearly readable. This finishes the attempt.
         - A7c1) If the competitor or the judge refuses to accept and sign the score sheet, the WCA Delegate must resolve the dispute.
-        - A7c2) The competitor must not sign an attempt on the score sheet before the judge has recorded the final result for that attempt. If a competitor signs (or otherwise marks) an attempt before they have begun it, or before the judge has finished recording the result, they forfeit the attempt and the result will be considered DNS.
+        - A7c2) The competitor must not sign an attempt on the score sheet before the judge has recorded the final result for that attempt and signed. If a competitor signs (or otherwise marks) an attempt before they have begun it, or before the judge has finished recording the result, they forfeit the attempt and the result will be considered DNS.
```
lgarron: This interacts with https://github.com/thewca/wca-regulations/pull/448 , but this should be DNF if the attempt has already started. We should probably include that change here to make sure it stays consistent.

wca-regulations.md
pedrosino: I'd say we kindly ask Luis to change the scoresheet :)

I asked him once but can't remember the reasons he gave back then.
@Luis-J-Ianez ?

wca-regulations.md
pedrosino: Yeah, I thought of that. I'm fine merging both PRs if necessary.

wca-regulations.md
Luis-J-Ianez: I can swap the signature boxes if this helps.

pedrosino: That would help a lot, I believe. It is more natural that the judge records the result and signs right away in the next box.

Luis-J-Ianez: OK, I can do this if WRC agrees. I wonder if I should wait for CubingUSA Nats and WC to be over before doing such a change.

lgarron: Alternatively: Use them as a chance to train judges from all over the US/world?

Laura-O: Well, I don't know how the staff at US Nationals is chosen, but I doubt the staff at worlds are those judges that need training on this. :-)

I think it's better to wait until this busy time is over, let @Luis-J-Ianez make an announcement and then make the change.

lgarron: How so? The judge is currently allowed to sign after the competitor, and I don't see a reason to expect everyone at Worlds staff to sign before the competitor?

Laura-O: Are we talking about the same thing? I was talking about changing the design of the score sheets.

I am sure this will confuse some people: judges will sign in the wrong column or competitors will ask why judges signed in the wrong column.
I think it's bad to change it right before the largest and the third (?) largest competition ever, as it will only generate more issues and maybe even incidents.

jfly: @timreyn, is us nationals even going to use cubecomps scorecards? As I recall, it never has in the past. In fact, the scorecards Tim generates for us Nats has the judge signature to the left of the competitor signature, so I think it would be great to remedy this discrepancy.

timreyn: Nope -- we're using cubingusa scorecards for CubingUSA Nationals, with group numbers printed on the scorecard, since we want to have everything tightly integrated with our scheduling and group-assigning software.

(of course, this is now the only competition left using cubingusa scorecards since all US competitions are migrated to the WCA site)

+1 to not changing things immediately before cubingusa nationals and worlds, either regulations or scorecard layouts.

wca-regulations.md
Laura-O: Yes, merging both would be great. Can you do that @pedrosino ?

Laura-O:
@pedrosino Could you merge this PR with #448, please?
If you are busy now, please let us know and we will do that.

pedrosino:
Done, sorry about the delay.
I'll close the other PR.

wca-regulations.md
```diff
@@ -413,11 +413,11 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
...
+    - A7c) The competitor must check the recorded result, and sign (or initial) the score sheet to acknowledge that the result is correct, complete, properly formatted, and clearly readable. This finishes the attempt.
         - A7c1) If the competitor or the judge refuses to accept and sign the score sheet, the WCA Delegate must resolve the dispute.
-        - A7c2) The competitor must not sign an attempt on the score sheet before the judge has recorded the final result for that attempt. If a competitor signs (or otherwise marks) an attempt before they have begun it, or before the judge has finished recording the result, they forfeit the attempt and the result will be considered DNS.
+        - A7c2) The competitor must not sign an attempt on the score sheet before the judge has recorded the final result for that attempt and signed. If a competitor signs (or otherwise marks) an attempt before they have begun it, they forfeit the attempt and the result will be considered DNS. If the competitor signs (or otherwise marks) an attempt after stopping the solve but before the judge has finished recording the result, it will be considered DNF..
```
lgarron: "after stopping the solve" should probably be "after starting the attempt"

Otherwise, there is a loophole where the competitor could sign during the solve (ignoring the possible interpretation that this qualifies as "assistance" per A5b).

Laura-O:
LGTM!
Thanks for updating @pedrosino!

PR #448, opened 2017-06-28
DNS/DNF for signing empty results

### LINKED ISSUE #427: DNS for signing empty results after an attempt

pedrosino:
Fixes #427


wca-regulations.md
```diff
@@ -417,7 +417,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
...
     - A7c) The judge and competitor must each check the recorded result, and sign (or initial) the score sheet to acknowledge that the result is correct, complete, properly formatted, and clearly readable. This finishes the attempt.
         - A7c1) If the competitor or the judge refuses to accept and sign the score sheet, the WCA Delegate must resolve the dispute.
-        - A7c2) The competitor must not sign an attempt on the score sheet before the judge has recorded the final result for that attempt. If a competitor signs (or otherwise marks) an attempt before they have begun it, or before the judge has finished recording the result, they forfeit the attempt and the result will be considered DNS.
+        - A7c2) The competitor must not sign an attempt on the score sheet before the judge has recorded the final result for that attempt. If a competitor signs (or otherwise marks) an attempt before they have begun it, they forfeit the attempt and the result will be considered DNS. If the competitor signs (or otherwise marks) an attempt before the judge has finished recording the result, it will be considered DNF.
```
lgarron: I think this is not quite sufficient. Technically, "If the competitor signs (or otherwise marks) an attempt before the judge has finished recording the result, it will be considered DNF." also applies before the attempt has started.

I think the best way to make this clear would be to split it into `A7c2` (signing before the attempt) and `A7c3` (signing after the start of the attempt, but before the judge has written the result.

pedrosino: Oh, sorry. This got lost in the process of making the PR. See my proposal at the original issue.
I will update the PR.

wca-regulations.md
```diff
@@ -417,7 +417,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
...
     - A7c) The judge and competitor must each check the recorded result, and sign (or initial) the score sheet to acknowledge that the result is correct, complete, properly formatted, and clearly readable. This finishes the attempt.
         - A7c1) If the competitor or the judge refuses to accept and sign the score sheet, the WCA Delegate must resolve the dispute.
-        - A7c2) The competitor must not sign an attempt on the score sheet before the judge has recorded the final result for that attempt. If a competitor signs (or otherwise marks) an attempt before they have begun it, or before the judge has finished recording the result, they forfeit the attempt and the result will be considered DNS.
+        - A7c2) The competitor must not sign an attempt on the score sheet before the judge has recorded the final result for that attempt. If a competitor signs (or otherwise marks) an attempt before they have begun it, they forfeit the attempt and the result will be considered DNS. If the competitor signs (or otherwise marks) an attempt after solving but before the judge has finished recording the result, it will be considered DNF.
```
Laura-O, line 420: We do not use "solving" in this context Maybe _"... after stopping the solve but before the judge has finished recording the result ..."_ would be better?

pedrosino:
This was merged to #447

PR #449, opened 2017-06-28
Remove ranking-based cutoff for combined rounds

### LINKED ISSUE #425: Get rid of ranking based combined round cutoffs
[9g2](https://www.worldcubeassociation.org/regulations/#9g2) says:

> 9g2) Whether a competitor proceeds to next phase of a Combined round, must be decided by ranking (best x competitors) or by result (all competitors with a best result under x) of the first phase.

I propose removing the part about ranking and changing this to:

> 9g2) Whether a competitor proceeds to next phase of a Combined round, must be decided by result (all competitors with a best result under x) of the first phase.

I just chatted with @Laura-O about this. We both have never seen a ranking based cutoff actually used, and in practice, it would be a big hassle to enforce. As Laura just said to me "it doesn't really make sense... enter the results first and then decided who may proceed to the next phase... yay :stuck_out_tongue:"

As the WST works on the WCIF spec and adding support for announcing round cutoffs and time limits, we're intentionally *not* going to support ranking based cutoffs.

pedrosino:
Fixes #425


Claster:
Looks good.

lgarron:
When was the last time this was used?

If it's been a few years, then I support this (like removing qualification rounds).

Claster:
This was used at Polish Championship 2016: #425

AlbertoPdRF:
This is already fixed on the branch Fixes-2017, so I'm closing this one.

PR #450, opened 2017-06-28
Clarify borrowing puzzles

### LINKED ISSUE #390: 3a+++ wording
Pedantic/pointless change but anyway:

3a+++) CLARIFICATION Competitors may borrow puzzles from other competitors privately, to use in competition.

A competitor is also allowed to borrow puzzles from others who are not competing, right?

pedrosino:
Fixes #390


lgarron:
On reflection, I think it would be good to be explicit. How about "others (including other competitors in the same round)"

pedrosino:
I'm fine with that as well. Will wait for someone else to comment

Laura-O:
Yep, I think Lucas' recommendation is better. Can you change your PR please @pedrosino ?

pedrosino:
Ok, done.

PR #451, opened 2017-06-28
Update A1 to clarify attempt and solve

### LINKED ISSUE #349: Conflict between A1a and A1a1 (attempt x solve)
> A1) Speed Solving attempts must abide by the following procedures.
>    A1a) The organisation team must enforce time limits for attempts and/or rounds.
>        A1a1) The default time limit per solve is 10 minutes, though the organisation team may announce a higher or lower time limit.
>        A1a2) Cumulative time limits may be enforced (e.g. 3 attempts with a cumulative time limit of 20 minutes). The time elapsed in a DNF result counts towards the cumulative time limit.

A1a says the limits should be enforced on "attempts and/or rounds", but A1a1 says the default limit is "per solve".
I think the latter is correct, since the limit is actually on the "solve" portion of the attempt, and not the whole attempt, which includes inspection, checking the puzzle and recording the result.

This also relates to the proposed change here: https://github.com/cubing/wca-regulations/commit/c7ab1122ec9f848a8e1c8741bc6d344a0cc8be32
It should be changed to "solve" instead of "result"

pedrosino:
(Tries to) fixes #349


lgarron:
Looks alright to me.

wca-regulations.md
```diff
@@ -349,9 +349,9 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
...
         - A1a1) The default time limit per solve is 10 minutes, though the organization team may announce a higher or lower time limit.
-        - A1a2) Cumulative time limits may be enforced (e.g. 3 results with a cumulative time limit of 20 minutes). The time elapsed in a DNF result counts towards the cumulative time limit.
+        - A1a2) Cumulative time limits may be enforced (e.g. 3 solves with a cumulative time limit of 20 minutes - see [Regulation A1a5](regulations:regulation:A1a5)). The time elapsed in a DNF result counts towards the cumulative time limit.
         - A1a3) For each round, any time limits must be announced before the round starts, and should not be changed after it has begun. Changes must be made at the discretion of the WCA Delegate, who must carefully consider the fairness of the change.
         - A1a4) The competitor must end each solve within the time limit. If a competitor reaches the time limit for a solve/round, the judge stops the attempt immediately and records the result as DNF. Exception: Multiple Blindfolded Solving (see [Regulation H1b1](regulations:regulation:H1b1)).
```
Claster: >judge stops the attempt

judge stops the solve?

pedrosino: Oh, good catch. I'll update.

pedrosino:
(I'm amending commits to keep a single one in the PR...let me know if this is fine or you would prefer both commits to appear)

AlbertoPdRF:
LGTM!

Issue #429, opened 2017-04-21
Regulation 9p2: What is a "cutoff time"?


jfly:
[Regulation 9p2 says](https://www.worldcubeassociation.org/regulations/#9p2):

> 9p2) The competitors who advance to the next round must be determined by either a cutoff ranking or a cutoff time in the preceding elimination round.

I was just chatting with @Laura-O about this, and we realized that it's not very clear what a "cutoff time" is. There are at least 2 problems with this regulation as written:

1. It forgets about events where your result is not just a time, such a 333fm and 333mbf. The way this is addressed in other regulations is to talk about "results", rather than "times".
2. It totally forgets about rounds where people are ranked by average, rather than single. It seems like right now 9p2 would allow you to run a round of 333 ave5 and allow competitors to advance if they got a best solve < 10 seconds. This is certainly not the intention of the regulation!
  - Note: Rewording 9p2 to address this is going to be tricky: remember that not all rounds are ranked by average, and 333bf is ranked by single, but an average is computed for the event.

AlbertoPdRF:
Problem 1 is solved with the rewording suggested on #423 

I would solve problem 2 removing the possibility of advancing to a next round by result, as this is barely used and also has some problems: for instance if more than 75% competitors beat said result, some competitors won't proceed to the next round even when they met the requirements to do it. If my example wasn't clear, I'll try to explain it better.

Also, I think that this issue should have been a comment on #423 in order to keep the discussion about this regulation all in the same place.

@thewca/wrc-team Can we discuss how to deal with that regulation? I can make the PR when we are done :)

pedrosino:
@AlbertoPdRF I think there is no need to remove completely the possibility of advancing by result. Your proposal in #423 solves this, I think, and it may be useful if you only want people under a certain result to advance (because you don't have a lot of time for the next round).

AlbertoPdRF:
I think that adding a clarification (Guideline) stating that by result we mean the single or average that determines the ranking in the respective Round would fix problem 2 that @jfly pointed out. Thoughts?

jfly:
@albertopdrf, I don't love that we'd need a new guideline to make this clear, but I don't right now have a better suggestion. Could you send in a PR implementing what you suggest and we can comment on it there?

AlbertoPdRF:
I also don't love it, but I can't see an easier or more elegant solution at the moment. I'll open the PR once we organize all of our branches for the upcoming changes :)

PR #467, opened 2017-10-09
Cleanup about cutoffs.

### LINKED ISSUE #423: Cleanup the regulations about cutoffs
I just did a careful read of how the regulations talk about cutoffs (with the intent of documenting these things in our competition JSON format (see https://github.com/thewca/worldcubeassociation.org/issues/946)). There are some ambiguities and inconsistencies with how the relevant regulations are worded that I think we could clean up.

The regulations define 2 kinds of cutoffs:

- [9g2](https://www.worldcubeassociation.org/regulations/#9g2): Advancing to the next phase of a combined round ("decided by ranking (best x competitors) or by result").
- [9p2](https://www.worldcubeassociation.org/regulations/#9p2): Advancing to the next round of an event ("determined by either a cutoff ranking or a cutoff time in the preceding elimination round").

## My complaints

1. The regulations do not define good terms to distinguish between these 2 concepts. As things stand right now, it's ambiguous what "cutoff" refers to. IMO, this ambiguity is part of why it has been so difficult to stamp out usage of the phrases "hard cutoff" and "soft cutoff".
2. 9p2 uses the phrase "cutoff time", which seems to exclude 333mbf and 333fm, events which are not ranked by time (see #429).
3. 9p2 uses the phrase "elimination round", which shows up nowhere else in the regulations. I know what it's referring to, but I think the word "elimination" only serves to increase cognitive load, forcing the reader to ask: "how does an elimination round differ from a regular round?"

## How I'd address my complaints

1. Invent new terms to distinguish between "next phase cutoff" and "next round cutoff" and remove all references to vanilla "cutoff" from the regulations. Here are some suggestions:
    - 9g2: "next phase cutoff", "phase cutoff", "combined round cutoff", "combined cutoff", or "soft cutoff"
    - 9p2: "next round cutoff", "round cutoff", "advancing cutoff"
2. Change the phrasing of 9p2 to be exactly the same as the phrasing of 9g2: "decided by ranking (best x competitors) or by result".
3. Remove the word "elimination", I don't believe it adds any value here.

I'm happy to send in a PR with some or all of these changes if you like these ideas.

### LINKED ISSUE #426: Clarification on how exactly 9g2 cutoffs by result work
Emphasis mine:

> [9g2](https://www.worldcubeassociation.org/regulations/#9g2)) Whether a competitor proceeds to next phase of a Combined round, must be decided by ranking (best x competitors) or **by result (all competitors with a best result under x)** of the first phase.

Just to be clear about this, we have 3 "kinds" of results, depending on the event:

- FMC: Integer (number of turns) or DNF or DNS
- Multi bld: 2 integers + centiseconds or DNF or DNS
- Every other event: Centiseconds or DNF or DNS

A liberal interpretation of regulation 9g2 makes it sound like you could theoretically have a combined round of FMC mean of 3 where people must solve the first attempt in less than or equal to 30 moves in order to be allowed to participate in the second and third attempts.

Things get weird for multibld. [9f12c](https://www.worldcubeassociation.org/regulations/#9f12c) defines what a "result" is for mbld, and it says that "a greater difference [result] is better". This, combined with the fact that 9g2 says "all competitors with a best result **under** x" results in a funny conclusion: you'd have to do "worse" to advance to the next phase a combined round of mbld. (aplogies if I got this wrong, I read the relevant regulations *many* times before writing this).

Is there any reason to ever hold a combined round of FMC or a combined round of multibld? Supporting this will make our software significantly more complicated than I feel it is worth doing. Similarly with #425, in practice I bet these never happen, in part because cubecomps just doesn't support it.

I propose extending 9g2 to only allow combined rounds for events where results are measured in centiseconds.

AlbertoPdRF:
Fixes #423 and partly #429.

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -263,7 +263,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 - 9o) Combined rounds count as one round when counting the number of rounds per event.
 - 9p) If an event has multiple rounds, then:
     - 9p1) At least 25% of competitors must be eliminated between consecutive rounds of the same event.
-    - 9p2) The competitors who advance to the next round must be determined by ranking (best x competitors) or by result (all competitors with a better result than x) in the preceding round.
+    - 9p2) The competitors who advance to the next round must be determined by result (all competitors with a better result than x) in the preceding round.
     - 9p3) If a qualifying competitor withdraws from a round, they may be replaced by the best-ranked non-qualifying competitor from the preceding round.
 - 9s) Each round of each event must have a time limit (see [Regulation A1a](regulations:regulation:A1a)).
 
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -248,7 +248,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 9f14) For "Mean of 3" and "Average of 5" rounds, if two or more competitors achieve identical average/mean results, rankings are assessed based on the best attempt per competitor, where "better" is defined as the shorter time.
     - 9f15) Competitors who achieve the same result in a round receive an identical ranking for the round.
 - 9g) A Combined Round consists of two phases of attempts, where competitors advance to the second phase if they meet a designated cutoff during the first phase.
-    - 9g2) Whether a competitor proceeds to next phase of a Combined round, must be decided by ranking (best x competitors) or by result (all competitors with a best result under x) of the first phase.
+    - 9g2) Whether a competitor proceeds to next phase of a Combined round, must be decided by ranking (best x competitors) or by result (all competitors with a better result than x) of the first phase.
 - 9i) Results of official WCA competitions must be listed on the WCA world rankings.
     - 9i1) The WCA recognizes the following types of regional records: national records, continental records, and world records.
     - 9i2) All the results of a round are considered to take place on the last calendar date of the round. If a regional record is broken multiple times on the same calendar date, only the best result is recognized as breaking that regional record.
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -263,8 +263,8 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 - 9o) Combined rounds count as one round when counting the number of rounds per event.
 - 9p) If an event has multiple rounds, then:
     - 9p1) At least 25% of competitors must be eliminated between consecutive rounds of the same event.
-    - 9p2) The competitors who advance to the next round must be determined by either a cutoff ranking or a cutoff time in the preceding elimination round.
-    - 9p3) If a qualifying competitor withdraws from a round, they may be replaced by the best-ranked competitor below the cutoff from the preceding round.
+    - 9p2) The competitors who advance to the next round must be determined by ranking (best x competitors) or by result (all competitors with a better result than x) in the preceding round.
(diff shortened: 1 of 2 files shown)


AlbertoPdRF:
With the latest commit, both #423 and #429 are addressed. Let's discuss if there are better solutions :)

EDIT: I wrote 426 instead of 423 in the first commit, so I'll address that issue in this PR too. After all, those 3 issues are related one with another :)

wca-regulations.md
```diff
@@ -248,7 +248,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 9f14) For "Mean of 3" and "Average of 5" rounds, if two or more competitors achieve identical average/mean results, rankings are assessed based on the best attempt per competitor, where "better" is defined as the shorter time.
     - 9f15) Competitors who achieve the same result in a round receive an identical ranking for the round.
 - 9g) A Combined Round consists of two phases of attempts, where competitors advance to the second phase if they meet a designated cutoff during the first phase.
-    - 9g2) Whether a competitor proceeds to next phase of a Combined round, must be decided by ranking (best x competitors) or by result (all competitors with a best result under x) of the first phase.
+    - 9g2) Whether a competitor proceeds to next phase of a Combined round, must be decided by ranking (best x competitors) or by result (all competitors with a better result than x) of the first phase.
```
jfly, line 251: Might it make sense to link to [9f12](https://www.worldcubeassociation.org/regulations/#9f12) here after the use of the word "better"?

jfly (commented):
This looks great to me!

wca-regulations.md
AlbertoPdRF: @thewca/wrc-team Thoughts? I personally think it's a good idea!

Laura-O:
LGTM!

wca-regulations.md
jfly: Bump, this looks like it got merged without this chance happening. Thoughts?

Laura-O: Yeah, I completely missed that before merging. Thanks for checking, @jfly

AlbertoPdRF: Addressed on #478

pedrosino:
I believe there's an error here:
Regulation
> 9p2) The competitors who advance to the next round must be determined by ranking (best x competitors) or by result (all competitors with a better result than x) in the preceding round.

Guideline
> 9p2+) [CLARIFICATION] Best result refers to the single or average that determines the ranking of the corresponding round.

There's no "best result" in the regulation anymore (it used to be, but at 9g2), so the guideline doesn't make sense. Also, "better result than" sounds weird. I talked to @jfly and he agrees that "result better than" is nicer.

Actually, there's no need to talk about a "best result" in either place. You need A result better than X.

Here's a suggestion:
> 9g2) Whether a competitor proceeds to next phase of a Combined round must be decided by result (all competitors with a **result better than** x) of the first phase (see [Regulation 9f12](regulations:regulation:9f12)).

> 9p2) The competitors who advance to the next round must be determined by ranking (best x competitors) or by result (all competitors with a **result better than** x) in the preceding round.

> 9p2+) Result refers to the single or average that determines the ranking of the corresponding round.


EDIT

Comparison of current text and my suggestion:
https://www.diffchecker.com/v1CdIXKv

PR #471, opened 2017-10-22
Remove combined round by ranking #425

Laura-O:


AlbertoPdRF:
We should change 

> better result under x

to 

> better result than x

This is already addressed on #467, so maybe we should just get rid of the combined round by ranking there.

Laura-O:
> This is already addressed on #467, so maybe we should just get rid of the combined round by ranking there.

Yep, can you change your PR accordingly?

AlbertoPdRF:
Done! So I'm closing this one.

PR #474, opened 2017-10-23
3d3: switch order tiles and stickers

### LINKED ISSUE #280: 3d3) "tiles or stickers" or "stickers or tiles"
In statement 3d, "stickers" and "tiles" appear in this order, in other words, "tiles" is described after "stickers".
It would be more natural if we change statement 3d3 to

```
If the coloured parts are stickers or tiles,
```

instead of

```
If the coloured parts are tiles or stickers,
```

Laura-O:
Fixes #280


lgarron:
LGTM

PR #477, opened 2017-10-28
A1a2++: events -> rounds #457

### LINKED ISSUE #457: A1a2++ Clarification on possible combinations for combined cumulative time limits
[A1a2++](https://www.worldcubeassociation.org/regulations/guidelines.html#A1a2++) is not clear about what can be combined with a cumulative time limit: 
> Cumulative time limits may be enforced for a single event or a combination of events.

Time limits are enforced for rounds or a combination of rounds, not for events. This should be clarified.
Additionally, we should decide whether it should be possible to use cumulative limits for a combination of rounds of the same event (like "15 minutes for BLD attempts in round 1 and 2").

I don't have a strong opinion, but I think it's not the intention of this regulation to combine limits of several rounds of one event. Furthermore, this could cause weird cases/results if competitors have reached the combined time limit and still proceed to the next round.

Laura-O:
Fixes #457

* event(s) was changed to round(s)
* Cumulative time limits for a combination of events cannot be enforced for more than one round of an event (e.g. not for round 1+2 of 3BLD)

### DIFF
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -149,7 +149,7 @@ To be more informative, each Guideline is classified using one of the following 
 ## <article-A><speedsolving><speedsolving> Article A: Speed Solving
 
 - A1a2+) [ADDITION] In case of a cumulative time limit, the judge records the original recorded time for a DNF on the score sheet in parentheses, e.g. "DNF (1:02.27)".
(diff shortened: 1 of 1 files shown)


Claster:
Looks good.

PR #478, opened 2017-10-29
Fix #467 mistake.

AlbertoPdRF:
On #467 I made a mistake: instead of getting rid of ranking based cutoffs, I got rid of the possibility to advance to the next round by ranking of the preceding round.


jfly:
Wow, sorry I missed that when reviewing your change! This looks good to me.

AlbertoPdRF:
Don't worry, it was my fault :P 

I'll link to 9f12 after the use of the word "better" on 9g2) on this PR also.

Laura-O:
LGTM!
Thanks for noticing and fixing this. :-)

PR #479, opened 2017-10-29
Update 9b

### LINKED ISSUE #299: Change format for Feet to "Average of 5"
Adam Polkowski (from [this discussion](https://groups.google.com/d/msg/wca-delegates/eZU7sSa8h2s/1f3W6c9fBwAJ)):

> After the first day of Kaliska Open 2015 we had a general meeting of Polish Cubing Association. 
> We had a lot of interior topics, but also one which I must send to you. 
> In first round of 3x3x3 with feet, Jakub Kipa had problem with timer. 
> Timer didn't start. I know that is his mistake, but he didn't proceed. The skill level of this event around the world is growing up very fast.Let's change the format for averages of 5.
> What is your opinion?

### LINKED ISSUE #453: Changing possible formats for events?
Following discussion about a recent case of 2x2 being held as Best of 1, Oleg said:

```
Currently it seems to be possible to have some weird cut-off formats like best of 3 / average of 5 for 3x3, or best of 1 / best of 3 for 3x3 bld.

How about the following?
1) For events from 9b1 only "avg 5" and "bo2 / avg2" are possible;
2) For events from 9b2 only "bo1", "bo2", "mean3" and "bo1 / mean3" are possible;
3) For events from 9b3 only "bo1", "bo2", "bo3" and "bo2 / bo3" are possible.

What could be valid reasons to use anything else?
```

about 1) I don't think we have any "avg2" format. I guess you mean "bo2 / avg5"
about 3) I would prefer "bo1 / bo3" if we did combined rounds limitations. See below

I don't see a problem with the regulations allowing different formats for each event, and also regulating how combined rounds must be can get complicated.

I think there may be valid reasons to hold best of X instead of Avg of 5 for some "popular" events. Not that I like the best of 1 for 2x2, for many reasons :P 

(EDITED because my numbering was changed automatically to 1) and 2) instead of the desired 1) and 3)

Laura-O:
First draft for an implementation of #453.
It only changes the format of events in 9b1). I think these are the only events which require default formats, but I am open to suggestions/changes.

Fixes #453, fixes #299.

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -221,13 +221,13 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 
 - 9a) The WCA governs competitions for mechanical puzzles that are operated by twisting groups of pieces, commonly known as "twisty puzzles".
 - 9b) The official events of the WCA are:
-    - 9b1) Rubik's Cube, 2x2x2 Cube, 4x4x4 Cube, 5x5x5 Cube, 3x3x3 One-Handed, 3x3x3 With Feet, Megaminx, Pyraminx, Rubik's Clock, Skewb, and Square-1.
+    - 9b1) 3x3x3 Cube, 2x2x2 Cube, 4x4x4 Cube, 5x5x5 Cube, 3x3x3 One-Handed, 3x3x3 With Feet, Clock, Megaminx, Pyraminx, Skewb, and Square-1.
         - 9b1a) Round format for these events is: "Average of 5". For a Combined Round the round format is "Best of 2" for the first phase and "Average of 5" for the second phase. Exception: In special cases, the round format for these events can be "Best of X" (where X is 1, 2, or 3), at the discretion of the WCA Board.
     - 9b2) 3x3x3 Fewest Moves, 6x6x6 Cube, and 7x7x7 Cube.
         - 9b2a) Round formats for these events are: "Best of X" (where X is 1 or 2) and "Mean of 3".
     - 9b3) 3x3x3 Blindfolded, 4x4x4 Blindfolded, 5x5x5 Blindfolded, and 3x3x3 Multi-Blind.
         - 9b3a) Round formats for these events are: "Best of X" (where X is 1, 2, or 3).
-        - 9b3b) For Rubik's Cube: Blindfolded, the WCA also recognizes "Mean of 3" rankings and records based on the times from "Best of 3" rounds.
+        - 9b3b) For 3x3x3 Blindfolded, the WCA also recognizes "Mean of 3" rankings and records based on the times from "Best of 3" rounds.
 - 9f) The results of a round are measured as follows:
     - 9f1) All timed results under 10 minutes are measured and truncated to the nearest hundredth of a second. All timed averages and means under 10 minutes are measured and rounded to the nearest hundredth of a second.
     - 9f2) All timed results, averages, and means over 10 minutes are measured and rounded to the nearest second (e.g. x.4 becomes x, x.5 becomes x+1).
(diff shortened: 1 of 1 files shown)


Claster:
I would say 9b2 and 9b3 need defaults as well.

AlbertoPdRF:
Yes, I agree with @Claster, although it isn't as usual to hold BigBLD Bo3 as it is to hold 3BLD Bo3. Same for FMC and the other events from 9b2).

Laura-O:
Yes, then let's discuss how they should look like. :-)

Suggestion:
* Move Feet to 9b1)

* 9b2) FMC, 6x6x6 Cube, and 7x7x7 Cube: Bo1, Bo2, Mo3 or Combined Bo1 and Mo3.
=> Compared to the current regulations, this would only forbid a combined round with Bo2/Mo3.

* 9b3) Rubik's Cube: Blindfolded, 4x4x4 Cube: Blindfolded, 5x5x5 Cube: Blindfolded, Rubik's Cube: Multiple Blindfolded: Bo1, Bo2, Bo3 and Combined Round Bo1/Bo2, Bo1/Bo3 and Bo2/Bo3
=> Compared to the current regulations, this would only forbid: nothing.

I think it would not be good to be more restrictive in 9b3). MultiBLD is sometimes held as Combined Round Bo1/Bo2 (for example at Euro 2016) and I think that's reasonable. In general, MultiBLD, 4BLD, and 5BLD often have a special role as there are fewer competitors and they require special treatment (in terms of having enough judges, no noise, etc.), so they are usually held at the beginning/end of a competition day or in parallel. I think it's difficult to define standard formats for them without making exceptions frequently.

jfly:
My understanding is that this does not affect our new edit events page, does it?

AlbertoPdRF:
@jfly If Feet is moved to 9b1), the answer is yes.

jfly:
Ok, please send in a PR or file an issue with whatever changes we need to make to the WCA website. Ideally we can coordinate so the change goes live when the regulations go into effect.

Laura-O:
Yes, sure!

AlbertoPdRF:
After thinking a bit more, I would prefer not to set defaults for 9b2) and 9b3) as they are right now, as the Board would need to make exceptions for almost every competition.

AlbertoPdRF:
This also fixes #299.

viroulep:
This PR needs to be rebased because of #488, but I like the changes!

AlbertoPdRF:
Before merging this one we should move 3x3 WF to 9b1).

Also, after the "Exception: " the "In" should have a lowercase "i" I think.

Claster:
Ok, I'm convinced about 9b2 and 9b3. Looks good!

Claster (approved):

pedrosino:
I created #510 this morning, and I'm fine with closing it in favour of this one.
However, there's still "Rubik's Cube" in some places, and Rubik's Clock, which we would prefer not to have.

Also, my PR has an update to 9a1 to mention the Rubik's Cube directly, which would be the only mention in the regulations.

Laura-O:
I fixed the order of the events here: the side events were ordered alphabetically, which wasn't correct anymore after removing "Rubik's".

viroulep:
I think we're all good here!

jfly:
@viroulep, besides changing Feet average of 5, did this change the allowed cutoffs in a way such that our edit events page should change? See https://github.com/thewca/worldcubeassociation.org/pull/2392 where @coder13 is making certain cutoffs unselectable.

viroulep:
Format-wise, feet now behaves just like 3x3, so I'd say @coder13 changes are just fine for feet!

viroulep:
After a quick chat with Jeremy I realized I totally missed his point.
Only Bo2 is possible for the first part of a combined round for all the events in 9b1, including feet.

I'll comment over the website PR.

Issue #405, opened 2017-01-03
Reorder and Rename events in 9b


Claster:
Recently the order of events was changed on rankings and competitions pages. For consistency, we should change them in Regulations (9b) as well.

The new order of events:

Rubik's Cube
2x2x2 Cube
4x4x4 Cube
5x5x5 Cube
6x6x6 Cube
7x7x7 Cube
3x3x3 Blindfolded
3x3x3 Fewest Moves
3x3x3 One-Handed
3x3x3 With Feet
Megaminx
Pyraminx
Rubik's Clock
Skewb
Square-1
4x4x4 Blindfolded
5x5x5 Blindfolded 
3x3x3 Multi-Blind

This implies that we should change 9b1 to:
9b1) Rubik's Cube, 2x2x2 Cube, 4x4x4 Cube, 5x5x5 Cube, Rubik's Cube: One-Handed, Megaminx, Pyraminx, Clock, Skewb, and Square-1.

9b2 to:
9b2) 6x6x6 Cube, 7x7x7 Cube, Rubik's Cube: Fewest Moves, and Rubik's Cube: With Feet.

No changes to 9b3.

Claster:
Oh wait, we should also rename events (probably that is the most important part :P) .

9b1) Rubik's Cube, 2x2x2 Cube, 4x4x4 Cube, 5x5x5 Cube, 3x3x3 One-Handed, Megaminx, Pyraminx, Clock, Skewb, and Square-1.

9b2) 6x6x6 Cube, 7x7x7 Cube, 3x3x3 Fewest Moves, and 3x3x3 With Feet.

9b3) 3x3x3 Blindfolded, 4x4x4 Blindfolded, 5x5x5 Blindfolded, and 3x3x3 Multi-Blind.

Laura-O:
I think this is fixed by renaming and reordering the events in 9b*.

We could change the headings of C-F, but I think that's not necessary and it might be confusing to have 4 paragraphs starting with "3x3x3...".

AlbertoPdRF:
This was fixed in #479

PR #481, opened 2017-10-29
Disallow time limit + cumulative time limit. Fix #374.

### LINKED ISSUE #374: Clarify whether a regular and cumulative limits can exist together
Relevant Regulations:

> A1a) The organisation team must enforce time limits for attempts and/or rounds.
> A1a2) Cumulative time limits may be enforced (e.g. 3 results with a cumulative time limit of 20 minutes). The time elapsed in a DNF result counts towards the cumulative time limit.

So it is not clear whether cumilative limits exist instead of regular ones or together with them. In practice I didn't see organizers set them both (though one could say that if an organizer sets a cumulative limit and doesn't set a regular one, the regular limit is 10 minutes by default).

If it is allowed to use them together, as a side effect it is allowed to use cumulative limit for multi-bld.

Inspired by French championship 2016 report.

Claster:
Fixes #374.


wca-guidelines.md
```diff
@@ -151,6 +151,8 @@ To be more informative, each Guideline is classified using one of the following
 - A1a2+) [ADDITION] In case of a cumulative time limit, the judge records the original recorded time for a DNF on the score sheet in parentheses, e.g. "DNF (1:02.27)".
 - A1a2++) [CLARIFICATION] Cumulative time limits may be enforced for a single round or a combination of rounds of different events.
 - A1a2+++) [CLARIFICATION] In case of a cumulative time limit, the result of each remaining attempt in the round is recorded as DNS (Did Not Start) after a competitor reaches the time limit.
+- A1a2++++) [ADDITION] If cumulative time limits for a single event or a combination of events are enforced, such events don't have regular time limits per solve.
+- A1a2+++++) [CLARIFICATION] Events where time limit per solve is defined by Regulations (Fewest Moves Solving and Multiple Blindfolded Solving) can't have cumulative time limits.
```
Laura-O: I think this should be either "a time limit" or "time limits".

AlbertoPdRF: Also, we should take into account #405 here and name the events correctly.

EDIT: Or just fix #405 after merging this one.

lgarron: Nit: "cannot" instead of "can't".

Laura-O:
@Claster Can you please implement the suggestions so that we can merge this PR?

AlbertoPdRF:
LGTM!

PR #483, opened 2017-10-29
Clarify use of headphones/earbuds

Laura-O:
I think we discussed this at WC2017.
Regulations are quite unclear about the use of wireless headphones/earbuds and I think we should be stricter here.


AlbertoPdRF:
LGTM!

PR #485, opened 2017-10-29
DNF for remaining attempts after disqualification. Fix #392.

### LINKED ISSUE #392: What happens with results if a competitor is disqualified?
Say, a round has format average of 5, and a competitor is disqualified after the first attempt according to A1c. What should be the remaining results? 4 empty results, 4 DNS or 4 DNF?

Laura's reply:

> he [delegate] has applied A1c), which says that either the attempt can be disqualified (DNF) or the competitor can be disqualified from the event.
> 
> For the latter the regulations refer to 2j):
> 2j) The WCA Delegate may disqualify a competitor from a specific event.
> 2j1) If a competitor is disqualified from an event for any reason, they are not eligible for any more attempts in the event.
> 
> And then there is this guideline:
> 9f5+) CLARIFICATION The result for an attempt is DNS if the competitor was eligible for the attempt and did not start it (see Regulation A3b2). If the competitor did not qualify for an attempt (e.g. in a combined round), they have no result for the attempt.
> 
> So in this case, the result would be just one DNF and not 4 additional DNS.
> 
> If [delegate's] intentions was to disqualify the attempt only, the competitor was eligible to do the remaining 4 attempts and as these were declined, the result should be DNF, DNS, DNS, DNS, DNS. This decision doesn't make much sense here, but I think it's possible.
> DNFing all attempts does not seem to be a reasonable option here: in prior discussions (I think that was about cumulative time limits) we concluded that an attempt can only be DNFed when the competitor has started the attempt.

Claster:
Motivation for DNF:
1) We have several cases in the past where results were recorded as DNF.
2) Introducing brand new result DQ can break every piece of code in the world that relies on DNF and DNS as the only ones being less than zero.

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -91,6 +91,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - 2i2b) The competitor must not interact with (e.g. operate, hold, wear) any active camera. Exception: the competitor may wear a camera mounted on their head, as long as it is out of their sight and it is clear that they are not interacting with it (apart from wearing it).
 - 2j) The WCA Delegate may disqualify a competitor from a specific event.
     - 2j1) If a competitor is disqualified from an event for any reason, they are not eligible for any more attempts in the event.
+        - 2j1a) The results of all remaining attempts in this event are recorded as DNF.
     - 2j2) If a competitor is disqualified during the course of an event, their earlier results remain valid. Exception: cheating or defrauding (see [Regulation 2k2a](regulations:regulation:2k2a)).
 - 2k) At the discretion of the WCA Delegate, a competitor may be disqualified from some events (a single event, multiple events, or all events) if the competitor:
     - 2k1) Fails to check in or register in time for the competition.
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -91,7 +91,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - 2i2b) The competitor must not interact with (e.g. operate, hold, wear) any active camera. Exception: the competitor may wear a camera mounted on their head, as long as it is out of their sight and it is clear that they are not interacting with it (apart from wearing it).
 - 2j) The WCA Delegate may disqualify a competitor from a specific event.
     - 2j1) If a competitor is disqualified from an event for any reason, they are not eligible for any more attempts in the event.
-        - 2j1a) The results of all remaining attempts in this event are recorded as DNF.
+        - 2j1a) The results of all remaining attempts in the event are recorded as DNF.
     - 2j2) If a competitor is disqualified during the course of an event, their earlier results remain valid. Exception: cheating or defrauding (see [Regulation 2k2a](regulations:regulation:2k2a)).
 - 2k) At the discretion of the WCA Delegate, a competitor may be disqualified from some events (a single event, multiple events, or all events) if the competitor:
     - 2k1) Fails to check in or register in time for the competition.


wca-regulations.md
```diff
@@ -91,6 +91,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - 2i2b) The competitor must not interact with (e.g. operate, hold, wear) any active camera. Exception: the competitor may wear a camera mounted on their head, as long as it is out of their sight and it is clear that they are not interacting with it (apart from wearing it).
 - 2j) The WCA Delegate may disqualify a competitor from a specific event.
     - 2j1) If a competitor is disqualified from an event for any reason, they are not eligible for any more attempts in the event.
+        - 2j1a) The results of all remaining attempts in this event are recorded as DNF.
```
AlbertoPdRF: This or the? :P We use the above, and this below.

Laura-O: We rarely use "this" in the regulations, so "the event" sounds better for me.

lgarron:
> Motivation for DNF:
> 
> We have several cases in the past where results were recorded as DNF.
> Introducing brand new result DQ can break every piece of code in the world that relies on DNF and DNS as the only ones being less than zero.

I do not consider these to be good reasons.

Past cases are unambiguous mistakes based on the current Regulations/Guidelines: https://www.worldcubeassociation.org/regulations/guidelines.html#9f5+

If we think DQ is clearer, I don't think that we should use DNF instead of DQ just because existing software doesn't handle it yet.

We can update the few pieces of important existing software (CubeComps, WCA site), and then adopt this afterwards.

jfly:
:+1: to @lgarron's comments! I don't understand the specifics of this particular discussion, but in general, we cannot decide to do the wrong thing because we are afraid to change our software systems.

Laura-O:
Some thoughts on this:
In the past, disqualifications have been published anonymously by the WDC if the competitor was under 18 years old. Using "DQ" for disqualified attempts would add this information permanently to a WCA profile, so the information that the competitor was disqualified would be public.

Do we want that and is this in line with the spirit of the WCA?

lgarron:
> In the past, disqualifications have been published anonymously by the WDC if the competitor was under 18 years old. Using "DQ" for disqualified attempts would add this information permanently to a WCA profile, so the information that the competitor was disqualified would be public.

Isn't it already possible to cross-reference the posts in some case?
I just looked up a random case or two, and it's easy to find out whom https://www.worldcubeassociation.org/archive/forum_topics/2008 is referring to.

Is your concern that this would be visible to people who are not looking for it, and that DQ will carry a stigma?

Laura-O:
Yes, there is a huge difference between reading an old post and being able to track the competitor mentioned in this post, and visiting the profile of specific competitors and knowing that they were disqualified x years ago.

WCA profiles are also ranked high in search results for your name, so it's likely that people who know nothing about speedcubing visit the profile and see disqualified results.

Claster:
1) Keeping DNF is the least painful solution for everybody;
2) According to 9f5+, disqualified results should be empty. This, however, will break checks in Workbook Assistant and scripts WRT uses. So as is, this regulation is not applied, so why not to change it?

Laura-O (approved):

Claster:
Anybody else wants to comment or approve? ;)

AlbertoPdRF:
It's fine as it is I think.

LGTM!

lgarron:
I think Laura made a fair point, so I'm also fine with DNF. :-)

wca-regulations.md
pedrosino: Is it still possible to suggest here? I would propose that this phrase was part of 2j1, inside ( ):

2j1) If a competitor is disqualified from an event for any reason, they are not eligible for any more attempts in the event (i.e. the results of all remaining attempts in this event are recorded as DNF).

Laura-O: Do you have a reason for us? :-)

I think it's not "i.e.", otherwise we would not need this addition?
Also, it's quite an important regulation (considering that it was decided inconsistently in the past), so hiding it in parentheses might not be a good idea.

pedrosino: The reason is that this seems like a continuation or an explanation of the previous sentence.

"they can't do more attempts, and those remaining attempts are marked as DNF"

I don't feel too strong about this...if kept separately, I would suggest to add "In this case" to the beginning of 2j1a

Jambrose777:
2j1a makes it to where a competitor does not begin an attempt and will still receive a DNF. Is this fundamentally correct? Shouldn't their remaining solves be DNS since they have not attempted them?

SAuroux:
Yes, it is correct, as this is incorporated into the definition of DNFs. Meanwhile, a DNS would be incorrect (also according to its definition). See [9f4](https://www.worldcubeassociation.org/regulations/#9f4) and [9f5](https://www.worldcubeassociation.org/regulations/#9f5).

PR #486, opened 2017-10-30
Allow competitors to talk to delegates Fixes #297

### LINKED ISSUE #297: Competitor can communicate with the delegate during an attempt?
From https://groups.google.com/d/msgid/wca-delegates/CAD6HhcRtr%3DsJYwa6bV7pyNQmW%2BRqYp3vGwPAAmRN-rG%2BpeuPvg%40mail.gmail.com?utm_medium=email&utm_source=footer

I believe communicating with the delegate or judge is fine, so my proposal would be to simply add "or the delegate" to A5a

Laura-O:
After reading the discussion again, I think the most reasonable fix for this is to add the delegate.
Explicitly mentioning questions regarding regulations could be misinterpreted as a responsibility of the judge, so let's keep this simple.

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -392,7 +392,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - A4d3) If a stopwatch is in use, the judge starts the stopwatch as soon as the competitor starts the solve.
     - A4e) Time penalties for starting the solve are cumulative.
 - A5) During the solve:
-    - A5a) While inspecting or solving the puzzle, the competitor must not communicate with anyone other than the judge. Penalty: disqualification of the attempt (DNF).
+    - A5a) While inspecting or solving the puzzle, the competitor must not communicate with anyone other than the judge or the WCA Delegate. Penalty: disqualification of the attempt (DNF).
     - A5b) While inspecting or solving the puzzle, the competitor must not receive assistance from anyone or any object other than the surface (also see [Regulation 2i](regulations:regulation:2i)). Penalty: disqualification of the attempt (DNF).
 - A6) Stopping the solve:
     - A6a) The competitor stops the solve by releasing the puzzle and then stopping the timer. If a stopwatch is in use, the judge stops the stopwatch as soon as the competitor does this.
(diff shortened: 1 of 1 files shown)


wca-regulations.md
```diff
@@ -392,7 +392,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - A4d3) If a stopwatch is in use, the judge starts the stopwatch as soon as the competitor starts the solve.
     - A4e) Time penalties for starting the solve are cumulative.
 - A5) During the solve:
-    - A5a) While inspecting or solving the puzzle, the competitor must not communicate with anyone other than the judge. Penalty: disqualification of the attempt (DNF).
+    - A5a) While inspecting or solving the puzzle, the competitor must not communicate with anyone other than the judge or the WCA Delegate. Penalty: disqualification of the attempt (DNF).
```
jfly: nit: Competitions sometimes have multiple delegates listed as delegates for the competition (so referring to "the WCA delegate" can be nonsensical), and they also have delegates present who are not listed as delegates for the competition. A possible rephrasing that I'd be okay with would be "**a** WCA Delegate".

Laura-O: Thanks again!
It's fixed now.

AlbertoPdRF:
LGTM!

lgarron:
Hmm, this is already merged.

I think we should be careful because the Delegate can delegate responsibilities to others:
https://www.worldcubeassociation.org/regulations/#1c

The intent here should be "you can talk to someone who can respond on behalf of the Delegate".

We also don't use "a/any Delegate" elsewhere in the Regulations right now, and I'd prefer to avoid that, because:
- It's unclear if this refers to an official Delegate for the competition, or just any Delegate
- It's unclear if "a Delegate" here means something else than any other place we have Delegate discretion.

AlbertoPdRF:
Regarding the first part: I don't get what you are trying to say - - any Delegate can delegate its responsibilities. If a Delegate is not delegating a competition, it doesn't have any responsibilities.

Regarding the second part: if you change "the" for "a", all the set 1c) is still perfectly clear, as it starts with: "A WCA Delegate for a competition is responsible for:"

Issue #252, opened 2015-03-30
Penalties/fees for pre-registering without attending


lgarron:
Discussed in the [Melbourne Autumn 2015 Delegate's Report](https://groups.google.com/forum/#!msg/wca-delegates/8qjoKW87Lrs/CDQAhARiQ1EJ) thread.

AlbertoPdRF:
This is fixed with #489

PR #489, opened 2017-10-31
Pending/withdrawn registration fees #380

### LINKED ISSUE #380: Pending/withdrawn registration fees
The Board asked to add a regulation that allows to refuse registrations of competitors who owe any fees to the WCA and/or organizers

> Pedro suggested this as a possible idea:
> "I'd propose a 2c1a) Registration may include payment of due fees from past competitions."

AlbertoPdRF:
Fixes #380.

This is a proposal, let's discuss it :)

### DIFF
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -47,7 +47,7 @@ To be more informative, each Guideline is classified using one of the following 
 
 - 2c+) [ADDITION] First-time competitors should register using their legal name. They may register using a reasonable nickname, at the discretion of the WCA Delegate.
 - 2c++) [ADDITION] Competitors must not provide intentionally misleading information, and returning competitors should provide information consistent with past information (e.g. exact name and WCA ID).
-- 2c1+) [EXAMPLE] For example, the organization team may require a confirmation e-mail to be answered before a particular date, pre-payment of the registration fee, or paying due registration fees of past competitions.
+- 2c1+) [EXAMPLE] For example, the organization team may require a confirmation e-mail to be answered before a particular date, pre-payment of the registration fee, or paying due registration fees of past competitions. The organization team should carefully consider the fairness of the requirements.
 - 2d+) [ADDITION] Date of birth and contact information should be especially secured.
 - 2d++) [RECOMMENDATION] If a third party (e.g. journalist) asks the organization team to be put in contact with any competitor(s), the competitor(s) should first be asked for consent.
 - 2e+) [CLARIFICATION] If a competitor holds multiple citizenship, they may choose which country to represent at their first competition.
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -48,6 +48,7 @@ To be more informative, each Guideline is classified using one of the following 
 - 2c+) [ADDITION] First-time competitors should register using their legal name. They may register using a reasonable nickname, at the discretion of the WCA Delegate.
 - 2c++) [ADDITION] Competitors must not provide intentionally misleading information, and returning competitors should provide information consistent with past information (e.g. exact name and WCA ID).
 - 2c1+) [EXAMPLE] For example, the organization team may require a confirmation e-mail to be answered before a particular date, pre-payment of the registration fee, or paying due registration fees of past competitions. The organization team should carefully consider the fairness of the requirements.
+- 2c1++) [REMINDER] The requirements for a completed registration must be clearly stated on the competition website.
 - 2d+) [ADDITION] Date of birth and contact information should be especially secured.
 - 2d++) [RECOMMENDATION] If a third party (e.g. journalist) asks the organization team to be put in contact with any competitor(s), the competitor(s) should first be asked for consent.
 - 2e+) [CLARIFICATION] If a competitor holds multiple citizenship, they may choose which country to represent at their first competition.
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -47,8 +47,8 @@ To be more informative, each Guideline is classified using one of the following 
 
 - 2c+) [ADDITION] First-time competitors should register using their legal name. They may register using a reasonable nickname, at the discretion of the WCA Delegate.
 - 2c++) [ADDITION] Competitors must not provide intentionally misleading information, and returning competitors should provide information consistent with past information (e.g. exact name and WCA ID).
-- 2c1+) [EXAMPLE] For example, the organization team may require a confirmation e-mail to be answered before a particular date, pre-payment of the registration fee, or paying due registration fees of past competitions. The organization team should carefully consider the fairness of the requirements.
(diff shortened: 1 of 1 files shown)


Laura-O:
I like the idea in general, but I am not sure about the "paying due registration fees of past competitions" part as it might cause unfair treatment.

Replying to emails and paying the registration fee for the competition I am attending is usually a replicable action: I can show the email in my mail account and I have a receipt from Paypal (or whatever service is used).

Paying outstanding registration fees is more complicated: what happens when an organizer just claims a competitor didn't pay and there is no proof (e.g. because it was paid on-site). What happens if this organizer informed other organizers and they do not let the competitor compete?

jfly:
Could it work to turn things upside down? After a competition, Jeremy the delegate can go through some process with the WDC to have the results of Joe (a competitor with outstanding registrations fees) annulled if Jeremy can prove that Joe had outstanding fees and that Jeremy made a good faith effort to inform Joe of this and ask Joe for fees, but he never did. This would give Joe an opportunity to represent his case to the WDC, and if he chooses not to defend himself, that's his own fault.

AlbertoPdRF:
Yes @Laura-O, I agree with you. Maybe @jfly's suggestion could work, but I'm not quite sure how to implement it.

SAuroux:
I don't really see the issue to be honest: 

1. this is just an example.
2. this already happens.

So adding this guideline would not change anything, only depict what is already done in practice.

Laura-O:
> So adding this guideline would not change anything, only depict what is already done in practice.

What makes you think so? If that doesn't change anything, why should we add it?
I also don't think this is done by many organizers (I know of only one or two). Adding it as an example in the guidelines could change that.

The problem I see is that the organizer determines who is allowed to compete. 2c1 does not describe a clear approach how organizers should deal with that: whether they should just delete a registration, not approve it or try to the issue until the day of the competition.
As a delegate, I would definitely not like to see an organizers delete a registration after they heard that a registered competitor didn't pay the registration fee at a different competition because "the guidelines say I can do that".

Maybe a good way to deal with this is to change the "as determined by the organization team." in 2c1 to "at the discretion of the delegate"?

AlbertoPdRF:
I also wouldn't like to see an organizer to delete a registration because they can, but I don't think that at the discretion of the Delegate in 2c1) fixes that. Maybe another Guideline in addition to the one proposed here where all about the pending fees is explained, and adding there that this is at the discretion of the Delegate is better?

Also, this PR is related to #252 too.

AlbertoPdRF:
Any more opinions here?

Laura-O:
Maybe we could add an additional sentence like (to this guideline or in another one):

> The organization team should carefully consider the fairness of the requirements.

This is also a clarification that the other examples should be fair (for example, it would not be a fair requirement if the organizer expects replies within 12 hours).

wca-guidelines.md
```diff
@@ -47,6 +47,7 @@ To be more informative, each Guideline is classified using one of the following
 
 - 2c+) [ADDITION] First-time competitors should register using their legal name. They may register using a reasonable nickname, at the discretion of the WCA Delegate.
 - 2c++) [ADDITION] Competitors must not provide intentionally misleading information, and returning competitors should provide information consistent with past information (e.g. exact name and WCA ID).
+- 2c1+) [EXAMPLE] For example, the organization team may require a confirmation e-mail to be answered before a particular date, pre-payment of the registration fee, or paying due registration fees of past competitions. The organization team should carefully consider the fairness of the requirements.
```
viroulep: Should we mention that these requirements should be clearly stated on the competition's website?
It's fairly obvious to us, but we wouldn't want some organizers changing the requirements without the competitors knowing!

AlbertoPdRF: The organizers can't change the information tab of the WCA website once the competition is confirmed, so maybe that's the place where they should clearly state those requirements?

Claster: Yes, a good idea to clarify that such requirements must be clearly stated on the comp website.

viroulep:
Hum maybe directly add a specific "registration-requirements" field?
Since it's going to be a specific regulation it's worth to have a dedicated field, such as we have for the competitors limit.

We could give some examples as hints, such as "registration on WCA website", "registration on X website", "registration and payment must be done on the WCA website to be accepted".
I would give freedom to the user on how to format the message, as long as the Board understand it.

I'll mention @jfly for a second opinion, but I think adding such a field would be pretty straightforward and not much time consuming.

wca-guidelines.md
viroulep: I initially meant to reply here but obviously messed it up, so please see my reply on the main discussion :p

AlbertoPdRF:
I like the idea!

Laura-O:
This would be a field which has to be filled out when the competition is created and which cannot be changed by the organizer, correct?

Could this be a problem if organizers decide to ask registered competitors for an additional confirmation later?
I think it's fine to make organizer decide about this before a competition is announced, but they should be aware that this can't be changed later.

AlbertoPdRF:
> This would be a field which has to be filled out when the competition is created and which cannot be changed by the organizer, correct?

Correct.

> Could this be a problem if organizers decide to ask registered competitors for an additional confirmation later?
I think it's fine to make organizer decide about this before a competition is announced, but they should be aware that this can't be changed later.

I think that a warning below the field explaining that this can't be edited once the competition is confirmed should be enough.

Should we open an issue on the worldcubeassociation.org repository for someone to work on this? @jfly @viroulep

AlbertoPdRF:
I added another Guideline. Two questions:

-Is the wording ok (specially the [REMINDER] tag)?
-Should we switch the order of these new Guidelines?

viroulep:
> Should we open an issue on the worldcubeassociation.org repository for someone to work on this? @jfly @viroulep

I think so, at least to have something to track it there, as we may merge this before it's ready on the website.

> -Should we switch the order of these new Guidelines?

It feels better to me to switch the current order, but I don't think it's mandatory.

> -Is the wording ok (specially the [REMINDER] tag)?

The wording looks good to me, however I don't think we have this information elsewhere in the Regulations, so I'd probably not go for a reminder but probably an "ADDITION" tag.

AlbertoPdRF:
> I think so, at least to have something to track it there, as we may merge this before it's ready on the website.

Ok, done.

> It feels better to me to switch the current order, but I don't think it's mandatory.

Ok, if no one has a different opinion I'll switch the order.

> The wording looks good to me, however I don't think we have this information elsewhere in the Regulations, so I'd probably not go for a reminder but probably an "ADDITION" tag.

Yes, maybe this makes more sense.

Laura-O:
Yes, I agree, using the *ADDITION*-label makes the most sense here.

Laura-O:
LGTM!

PR #490, opened 2017-11-01
Advance conditions have to be announced, fixes #391

### LINKED ISSUE #257: Require competitions to announce conditions for advancing from each round, before the round has begun
These decisions are often made after the round as finished, which is not really accountable, and a bit unfair (it's easy to shift the cutoff around based on how nice you want to be to people near the cutoff... which are often your friends).

Proposal: require organizers to specify a cutoff (number of competitors or time limit) for advancing from each non-final round. This cutoff must be selected before the rounds start, and should preferably be listed on the website before the competition.

In case there are not enough competitors for a round, the Delegate can use his/her judgment. A simple policy is to allow the maximum number of competitors in that case.

### LINKED ISSUE #391: Condition for proceeding to the next round should be announced beforehand
Clarify that organizers should announce cutoff ranking or cutoff time (see 9p2) before the round starts. This criterion shouldn't change (similar to A1a3+). A good reason to change the criterion is that according to the current one more than 75% competitors pass to the next round; in this case organizers should let 75% pass.

Changing the condition or announcing it after the current round ends could be interpreted as organizers/delegates favor themselves or their friends, which may or may not be true.

Laura-O:
Copied from A1a3 and changed.
Fixes #257.

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -265,6 +265,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 - 9p) If an event has multiple rounds, then:
     - 9p1) At least 25% of competitors must be eliminated between consecutive rounds of the same event.
     - 9p2) The competitors who advance to the next round must be determined by ranking (best x competitors) or by result (all competitors with a better result than x) in the preceding round.
+        - 9p2a) For each round, advancement conditions must be announced before the round starts, and should not be changed after it has begun. Changes must be made at the discretion of the WCA Delegate, who must carefully consider the fairness of the change.
     - 9p3) If a qualifying competitor withdraws from a round, they may be replaced by the best-ranked non-qualifying competitor from the preceding round.
 - 9s) Each round of each event must have a time limit (see [Regulation A1a](regulations:regulation:A1a)).
 


wca-regulations.md
```diff
@@ -265,6 +265,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 - 9p) If an event has multiple rounds, then:
     - 9p1) At least 25% of competitors must be eliminated between consecutive rounds of the same event.
     - 9p2) The competitors who advance to the next round must be determined by ranking (best x competitors) or by result (all competitors with a better result than x) in the preceding round.
+        - 9p2a) For each round, advance conditions must be announced before the round starts, and should not be changed after it has begun. Changes must be made at the discretion of the WCA Delegate, who must carefully consider the fairness of the change.
```
jfly: **super nitpick and long explanation, sorry!**

I think "advancement conditions" sounds better to me "advance conditions".

Things get interesting if you Google ["advance vs advancement"](https://www.google.com/search?q=advance+vs+advancement&oq=advance+vs+advancement&aqs=chrome..69i57j0l3.4060j0j4&sourceid=chrome&ie=UTF-8), but here's my take on it as a native English speaker:

- advance can be both a noun and a verb, and when I see it, I think first think it's a verb.
- advancement can only be a noun

Given the above, when reading "advance conditions", my brain first parses "advance" as a verb, and then I get to "conditions" and the phrase makes no sense, so then I have to backtrack and interpret the phrase differently with advance as a noun. "advancement conditions" doesn't have this problem for me.

jfly: Urg, on further thought, I just realized that in the phrase "advance conditions", "advance" is probably an adjective, not a noun... probably best not to listen to me on any grammar stuff, but I do maintain that "advancement" sounds nicer to me here.

lgarron: I agree with "advancement". "Advancing" could also work, but I prefer it less.

Laura-O: Ok, I will change the PR tomorrow.
Thanks for the explanation @jfly!

AlbertoPdRF:
This also fixes #257.

Laura-O:
I updated the PR.

lgarron:
LGTM.

I would like to see us make the requirement more strict in the future, but this is a great start for now.

lgarron (approved):

PR #492, opened 2017-11-02
Regulation 2e) only "forces" delegates to verify citizenship #432

### LINKED ISSUE #432: Regulation 2e) only forces delegates to verify citizenship
Regulation [2e)](https://www.worldcubeassociation.org/regulations/#2e):

> Competitors must represent a country of which they hold citizenship. The WCA Delegate should verify citizenship standing by means of documents (e.g. a passport). If a competitor is found ineligible to represent the country under which they have registered, the competitor may be disqualified retroactively and/or suspended, at the discretion of the WCA Board.

It's the only regulation that forces a delegate to check legal documents against registration data at the competition's registrations, and only for citizenship. I think we should include a more general regulation in [Article 2](https://www.worldcubeassociation.org/regulations/#article-2-competitors) to force delegates to check legal documents against all the registration data (name, dob, etc.).

What do you think?

AlbertoPdRF:
Fixes #432.

I'm not convinced with the proposed wording, so suggestions are appreciated :).


wca-guidelines.md
```diff
@@ -50,6 +50,7 @@ To be more informative, each Guideline is classified using one of the following
 - 2d+) [ADDITION] Date of birth and contact information should be especially secured.
 - 2d++) [RECOMMENDATION] If a third party (e.g. journalist) asks the organization team to be put in contact with any competitor(s), the competitor(s) should first be asked for consent.
 - 2e+) [CLARIFICATION] If a competitor holds multiple citizenship, they may choose which country to represent at their first competition.
+- 2e++) [ADDITION] The WCA Delegate should also verify the name and date of birth standing by means of documents (e.g. a passport) of every new competitor at their first competition.
```
jfly: I just realized that [`2e`](https://www.worldcubeassociation.org/regulations/#2e) uses the phrase "citizenship standing". What does that mean? And what does "name and date of birth standing" mean in this new guideline?

AlbertoPdRF: My guess is it means that you have the documents that prove your citizenship, name, and dob, but I'm not sure at all. Also, I think we should do the same that we did on #486 and change "The WCA Delegate" for "A WCA Delegate", both here and in 2e).

jfly: I agree with changing "The WCA Delegate" to "A WCA Delegate".

I think the word "standing" should be removed from this guideline and `2e`, because it's only confusing me.

lgarron: We should probably stick with "the WCA delegate":

https://www.worldcubeassociation.org/regulations/#1c

AlbertoPdRF: We already say "a competition" instead of "the competition", why shouldn't we say "a WCA Delegate" instead of "the WCA Delegate"?

Laura-O: I get Lucas' point, but there are several regulations which suggest that a competition can only have one delegate (like [1c](https://www.worldcubeassociation.org/regulations/#1c)), so this is an essential inconsistency.

A workaround could be to use "the WCA Delegate" here and add a clarification for 1c that _the WCA Delegate_ refers to all WCA delegates of the competition.

SAuroux: I''m not sure why you call this a "workaround" @Laura-O, because this sounds like a very clean solution for this inconsistency.

AlbertoPdRF: If we go with Laura's proposal, it would be better to change the wording implemented in #486 for consistency, right?

lgarron:
> If we go with Laura's proposal, it would be better to change the wording implemented in #486 for consistency, right?

Yeah. I'll send a PR for that so we don't lose track.

wca-guidelines.md
lgarron: I've created https://github.com/thewca/wca-regulations/pull/498 and https://github.com/thewca/wca-regulations/issues/499

Laura-O:
@jfly suggested removing "standing" here. That sounds reasonable to me and makes the sentence easier to understand. Any more opinions on this?

AlbertoPdRF:
This one should be good to go now, unless any of you have a better wording in mind.

Laura-O:
Yep, LGTM!

PR #493, opened 2017-11-02
[Guidelines] 5b5+++ and 5b5++++ consistency #370

### LINKED ISSUE #370: [Guidelines] 5b5+++ and 5b5++++ consistency
5b5+ and 5b5++ start exactly the same way (Examples of parts physically detached, resulting in a", but this is not the case for 5b5+++ and 5b5++++

```
5b5+++) EXAMPLE Example of parts not fully attached, resulting in a solved state: a 5x5x5 centre piece twisted in its spot.
5b5++++) EXAMPLE Example of parts physically attached but not fully placed, resulting in a "DNF": a 3x3x3 edge piece slightly popped out.
```

In my understanding, they should start the same way, only with the result being different.

AlbertoPdRF:
WIP: there's still a change to make on the Regulations file.

Fixes #370.


AlbertoPdRF:
This is done now, let me know what you think!

Laura-O:
Just to make sure: this is only a change for consistency, it does not change the way we handle these cases, right?

AlbertoPdRF:
Yes, I did it with that intention. And apart from giving consistency, I think the meaning is clearer this way.

lgarron:
LGTM

PR #494, opened 2017-11-02
Guideline 4b4+ (exceeds?) #369

### LINKED ISSUE #369: Guideline 4b4+ (exceeds?)
4b4+) CLARIFICATION All attempts that have been started within the time frame follow the standard solving procedure (i.e. they are not stopped when the time frame of the relevant scramble sequence **exceeds**).

Maybe it should be "expires" instead?

AlbertoPdRF:
Fixes #369.


Laura-O:
LGTM!

PR #495, opened 2017-11-02
Clarification of A1b1 #375

### LINKED ISSUE #163: When is the mat required/optional?
Also see issues #90 and #162.

### LINKED ISSUE #375: Clarification of A1b1
Right now A1b1 states:

> A1b1) For events with inspection, a Stackmat must be used in addition to the stopwatch. For events without inspection where the competitor is expected to exceed 10 minutes, use of the Stackmat is optional.

As pointed out by @cubizh, @Claster and Daniel, one scenario is not covered:

> events with inspection => Stackmat must be used
> events without inspection where the competitor is expected to exceed 10 minutes => optional
> events without inspection where the competitor is not expected to exceed 10 minutes => ???

Also, there is the edge case where the competitor is expected to maybe exceed 10 minutes, but has a chance to solve under 10 minutes. I don't think it's too much trouble to use a Stackmat just in case.

I propose to change A1b1 to:

> A1b1) A Stackmat must be used in addition to the stopwatch. For events without inspection where the competitor is expected to exceed 10 minutes, use of the Stackmat is optional.

And add a clarification about unsure expected results, where a Stackmat should be used in case the result happens to be under 10 minutes.
If it sounds impractical to organizers used to hold big blind, please let me know!

Also, we have H1e:

> H1e) Use of the Stackmat timer is optional for timekeeping.

The interaction with the "new" A1b1 would be weird in case the competitor is expected to solve their multiblind attempt under 10 minutes: Stackmat would be mandatory but stackmat timer optional?
It seems to go in the opposite direction of the previous discussion: https://github.com/cubing/wca-regulations/issues/273, where it was going towards making the mats optional for multiblind.

So we could either simply remove H1e to make the whole Stackmat mandatory for multiblind expected to last less than 10 minutes (and optional otherwise), or keep H1e and make the whole Stackmat optional for every multiblind attempt.
(Or something else?)

Related report and videos leading to this issue:
https://groups.google.com/forum/#!topic/wca-delegates/FTU06Vg4xJ4
https://www.youtube.com/watch?v=MSLVXQg_qs0
https://www.youtube.com/watch?v=tmWxSZtqOkw

AlbertoPdRF:
Fixes #375, fixes #163.

I went with my own suggestion instead of those listed on the issue. Let me know what you think!

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -355,7 +355,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - A1a4) The competitor must end each solve within the time limit. If a competitor reaches the time limit for a solve/round, the judge stops the solve immediately and records the result as DNF. Exception: Multiple Blindfolded Solving (see [Regulation H1b1](regulations:regulation:H1b1)).
         - A1a5) A solve is considered to meet the time limit if and only if the final result, after any time penalties are applied, is less than the time limit. Exception: Multiple Blindfolded Solving (see [Regulation H1b1](regulations:regulation:H1b1)).
     - A1b) If the time limit for a solve is greater than 10 minutes, a stopwatch must be used for timekeeping.
-        - A1b1) For events with inspection, a Stackmat must be used in addition to the stopwatch. For events without inspection where the competitor is expected to exceed 10 minutes, use of the Stackmat is optional.
+        - A1b1) A Stackmat must be used in addition to the stopwatch. Exception: for events without inspection where the competitor is expected to exceed 10 minutes, use of the Stackmat is optional.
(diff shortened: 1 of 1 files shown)


Laura-O:
Let's wait a bit if someone else has something to add, but that sounds like a very neat solution for this issue!

Laura-O:
I think this also fixes #163.

AlbertoPdRF:
Indeed! Didn't see that one.

lgarron:
LGTM; I like that this PR uses "Exception:" in a similar way to other Regulations.

lgarron (approved):

viroulep (approved):
LGTM!
I edited the initial message to include the other issue in "fixes", so that it closes both of them when merged in draft.

Issue #460, opened 2017-09-07
D1b being vague


a-Maass:
D1b) During the attempt, the competitor must use only their feet and the surface to manipulate the puzzle. Penalty: disqualification of the attempt (DNF).

apparently I've heard of a few cases where the delegate thought that this didn't prohibit inspecting the cube with your hands, which seems false to me, I think this regulation should be rewritten to make it clear that you DO NOT use your hands for feet inspection

Randomno:
I don't think it's vague, it's just that a lot of people don't know what an attempt is and how it differs from a solve.

Definitely worth clarification/reminder in Guidelines, considering there's one for 2H inspection in OH

lgarron:
We're currently discussing this in a WRC email thread. My suggestion was to add a Guideline:

> D1b++) The competitor must not touch the puzzle with any part of their body other than their feet.

Laura-O:
Another solution would be to allow inspecting with hands.

- We already allow using both hands in OH inspection, so using both hands in feet would be consistent.
- It removes confusion, DNFs and general regulation violations.
- Although inspection with hands is easier, I am not sure if that's really a large advantage. It probably takes 2-3 seconds to take the cube and put it back, so you have less "pure" inspection time.

PR #496, opened 2017-11-03
D1b: allow inspection with hands

Laura-O:
Nobody complained in #460, so I am going ahead with my idea:
I changed D1b so that competitors must use their feet only after starting the timer, so inspection with hands is allowed.

As far as I can see, this change doesn't generate any loopholes:
- The inspection starts like in any other speedsolving event, competitors can pick up their cube or leave it on the mat and inspect with their feet like they could do in any other event.
- Starting the timer is defined in D3 and has to be done by placing the feet on the timer.
- Stopping is in D4.

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -467,7 +467,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 
 - D1) Standard speed solving procedures are followed, as described in [Article A](regulations:article:A) (Speed Solving). Additional regulations that supersede the corresponding procedures in [Article A](regulations:article:A) are described below.
     - D1a) During the attempt, the competitor must sit in a chair, sit on the surface, or stand.
-    - D1b) After starting the timer, the competitor must use only their feet and the surface to manipulate the puzzle. Penalty: disqualification of the attempt (DNF).
+    - D1b) During the solve, the competitor must use only their feet and the surface to manipulate the puzzle. Penalty: disqualification of the attempt (DNF).
 - D3) Starting the solve:
     - D3a) The competitor places their feet onto the timer sensors.
     - D3b) The competitor removes their feet from the timer sensors to start the solve.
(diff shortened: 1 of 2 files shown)


lgarron:
I personally still don't like this change. I still think OH should have one-handed inspection. :-P
However, I think it's the right thing to do.

Could you also include a guideline similar to https://www.worldcubeassociation.org/regulations/guidelines.html#3h2+ ? Suggestion:

> D1b++) CLARIFICATION In the past, competitors were required to inspect the puzzle using their feet for Solving with Feet. They are now permitted to touch the puzzle with their hands or any other part of their body during inspection (see Guideline A5b+).

Laura-O:
I added the guideline, that's a good idea!

Lesson learned: if you want to use the [wrc](https://github.com/thewca/wca-regulations-compiler) and you installed the "Wine Resource Compiler" you might get weird error messages. :-(

wca-regulations.md
```diff
@@ -467,7 +467,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 
 - D1) Standard speed solving procedures are followed, as described in [Article A](regulations:article:A) (Speed Solving). Additional regulations that supersede the corresponding procedures in [Article A](regulations:article:A) are described below.
     - D1a) During the attempt, the competitor must sit in a chair, sit on the surface, or stand.
-    - D1b) During the attempt, the competitor must use only their feet and the surface to manipulate the puzzle. Penalty: disqualification of the attempt (DNF).
+    - D1b) After starting the timer, the competitor must use only their feet and the surface to manipulate the puzzle. Penalty: disqualification of the attempt (DNF).
```
AlbertoPdRF: For consistency with [C1b)](https://www.worldcubeassociation.org/regulations/#C1b) I would change

> After starting the timer,

for

> During the solve,

Laura-O: Good catch. I can't remember why I changed this, but your suggestion is definitely better.

AlbertoPdRF:
LGTM!

kingmathyall:
LGTM too!

PR #498, opened 2017-11-07
A5a) Change "a WCA Delegate" to "The WCA Delegate" for consistency. #297 #486

lgarron:
See https://github.com/thewca/wca-regulations/pull/492/files#r149062523


jfly:
Didn't we agree to change clarify what "the WCA Delegate" means? Is that change happening somewhere else?

lgarron:
> Didn't we agree to change clarify what "the WCA Delegate" means? Is that change happening somewhere else?

Yes, it will have to.

Laura-O:
I will merge this now. Please don't forget to add a fix for #499 @lgarron.

PR #500, opened 2017-11-07
Clarify the (non)-interaction between A6c and A6e.

### LINKED ISSUE #373: Clarify A6c => A6e implication
Currently every violation of A6c automatically implies a violation of A6e, as a competitor touches a puzzle before a judge inspected it. Thus, violation of A6c is rewarded with +4 (or DNF), strickly speaking. We could clarify that if A6c is violated, but no moves applied, it is not counted towards A6e (though a second touch triggers A6e).

> A6c) The competitor must fully release the puzzle before stopping the solve. Penalty: time penalty (+2 seconds).
> A6e) The competitor must not touch or move the puzzle until the judge has inspected the puzzle. Penalty: disqualification of the attempt (DNF). Exception: If no moves have been applied, a time penalty (+2 seconds) may be assigned instead, at the discretion of the judge.

Inspired by 2nd Guwahati Open 2016 report.

viroulep:
Fixes #373.

I liked @SAuroux's suggestion but @notkevin888 made a good point about a possible loophole.
A6c and A6e refer to two distincts actions during the attempt that should lead to the same penalties.

So I clarified the separation of both actions with Sébastien's suggestion, and used A6e's wording in A6c to fix Michael point.
Let me know what you think.


wca-regulations.md
```diff
@@ -401,9 +401,9 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - A6b) The competitor is responsible for stopping the Stackmat timer correctly.
         - A6b1) If the timer stops before the end of the solve and the timer shows a time strictly below 0.06 seconds, then the attempt is replaced by an extra attempt. A competitor forfeits the right to the additional attempt if the WCA Delegate determines that the timer was stopped deliberately.
         - A6b2) If the timer stops before the end of the solve and displays a time of 0.06 seconds or higher, then the attempt is disqualified (DNF). Exception: if the competitor can demonstrate that the timer malfunctioned, they may receive an extra attempt, at the discretion of the WCA Delegate.
-    - A6c) The competitor must fully release the puzzle before stopping the solve. Penalty: time penalty (+2 seconds).
+    - A6c) The competitor must fully release the puzzle before stopping the solve. Penalty: disqualification of the attempt (DNF). Exception: If no moves have been applied after stopping the solve and before releasing the puzzle, a time penalty (+2 seconds) may be assigned instead, at the discretion of the judge.
```
notkevin888: Given the phrasing of A6a (`The competitor stops the solve by releasing the puzzle and then stopping the timer`), maybe change references to `stopping the solve` to `stopping the timer`?  I could see possible confusion where (a very pedantic) someone gets stuck in a recursive loop :P

notkevin888 (commented):
Clean fix!  Just one comment to reduce ambiguity even further :)

wca-regulations.md
viroulep: How would you loop?

I was about to disagree with the suggestion, but while writing the explanation below I realized it's actually better to use "stop the timer".

The way I see it, the attempt has 3 states: inspection, solve, recording.
Article A (3 to 7) describes what you can do during these states, and how transitioning between them works.

"Stopping the solve" is the transition between "solve" and "recording", and consist of two actions: "releasing the puzzle" and "stopping the timer".
A6c describes what happens when you mess up the order of the actions **within** the transition, so it doesn't make much sense to refer to "after/before stopping the solve", when you're precisely describing what is "stopping the solve".

So basically I would indeed replace occurrences of "stopping the solve" by "stopping the timer" within A6c.

Laura-O:
LGTM!

PR #501, opened 2017-11-08
Add restrictions to Z4, #469

### LINKED ISSUE #469: Adding stricter regulations for managing registrations
Because of reoccurring issues with competitor limits and the handling of waiting lists, I propose to add the following restrictions to the regulations:

- If there is a competitor limit and the competitor limit was reached before the competition, on-site registrations are not allowed.
- Competitor limits can only be changed with Board approval and only for a legitimate reason (e.g., when the venue was changed).
- Registrations may not be opened again once they were closed.

Laura-O:
First draft to fix #469 

I think the first two additions (Z4a + Z4b) are quite clear.

To explain the third a bit more:
It's a common approach to have two registration periods in China. Both of them are announced when the competition is announced. The second period is used to fill the spots of competitors who registered in the first period but canceled their registration.
I think that's not a bad approach as it makes the registration process transparent: people know that they will get a second chance to register and they know when this will happen.

For that reason, I decided against adding a stricter regulation reopening the registration.
In general, the main problem is that additional registration periods can currently be done in secret if an organizer just shares the news with his friends and no announcement is made. An announcement for 48 hours does not guarantee that everyone who is interested in registering knows of the additional registration period, but it gives everyone at least a chance.

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -521,5 +521,8 @@ Organization teams may adopt optional regulations to facilitate the administrati
 - Z1) The organization team may require competitors to submit puzzles during registration.
 - Z2) The organization team may limit the number of events per competitor.
 - Z3) The organization team may select competitors who directly qualify for certain rounds of certain events based on the results of specific previous competitions.
-- Z4) The organization team may limit the number of competitors per event or per competition, on either a "first come first serve" basis or based upon qualification results or rankings in the WCA world rankings of a previously announced calendar date.
+- Z4) The organization team may limit the number of competitors per event or per competition, on either a "first come first serve" basis or based upon qualification results or rankings in the WCA world rankings of a previously announced calendar date. If the number of competitors per event or per competition is limited, the following regulations apply:
+    - Z4a) If the limit for a competition or an event is reached before the competition, the organization team must not accept on-site registrations for the competition or the event.
+    - Z4b) Competitor limits should be announced together with the competition. Any changes to competitor limits are at the discretion of the WCA Board.
+    - Z4c) If there is more than one registration period, the organization team must announce each additional registration period at least 48 hours before it starts.
 - Z5) The organization team may prohibit competitors from participating in specific combinations of events.
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -206,3 +206,8 @@ To be more informative, each Guideline is classified using one of the following 
 - H1b1++) [ADDITION] The judge may permit the competitor to continue the attempt unofficially, but the attempt must be stopped and judged first, in full accordance with the Regulations.
 - H1d+) [EXAMPLE] Example: Suppose a competitor attempts 10 cubes, stops the solve with a time of 59:57, and has two time penalties. The final result is 59:57 + 2*2 = 60:01 (also see [Regulation A1a5](regulations:regulation:A1a5)).
 - H1d++) [EXAMPLE] Example: Suppose a competitor attempts 10 cubes, the judge stops the competitor at 60 minutes, and the attempt has two time penalties. The final result is 60:00 + 2*2 = 60:04.
+
+
+## <article-Z><optional><optional> Article Z: Optional Regulations
+
+- Z4b+) [EXAMPLE] Examples for reasons which justify a change of the competitor limit after the competition was announced: the venue was changed or the number of registrations is higher than expected.


wca-regulations.md
```diff
@@ -521,5 +521,8 @@ Organization teams may adopt optional regulations to facilitate the administrati
...
 - Z2) The organization team may limit the number of events per competitor.
 - Z3) The organization team may select competitors who directly qualify for certain rounds of certain events based on the results of specific previous competitions.
-- Z4) The organization team may limit the number of competitors per event or per competition, on either a "first come first serve" basis or based upon qualification results or rankings in the WCA world rankings of a previously announced calendar date.
+- Z4) The organization team may limit the number of competitors per event or per competition, on either a "first come first serve" basis or based upon qualification results or rankings in the WCA world rankings of a previously announced calendar date. If the number of competitors per event or per competition is limited, the following regulations apply:
+    - Z4a) If the limit for a competition or an event is reached before the competition, the organization team must not accept on-site registrations for the competition or the event.
+    - Z4b) All limits must be approved by the WCA Board and may only be changed for legitimate reasons (e.g. when the venue was changed).
```
lgarron: "legitimate" is wishy-washy here, since it really just means "whatever the WCA board accepts".

How about:
1) Competitor limits should be announced together with the competition. Any changes to competitor limits are at the discretion of the WCA Board.

I would put the example in the Guidelines, along with one or two other reasonable ones.  (What are common reasons that are accepted? If there aren't any other, can we require venue change as the reason?)

Laura-O: From what I know the most common reasons are venue changes and/or much more people registering than expected. I have added a guideline for these two examples.

wca-regulations.md
```diff
@@ -521,5 +521,8 @@ Organization teams may adopt optional regulations to facilitate the administrati
...
-- Z4) The organization team may limit the number of competitors per event or per competition, on either a "first come first serve" basis or based upon qualification results or rankings in the WCA world rankings of a previously announced calendar date.
+- Z4) The organization team may limit the number of competitors per event or per competition, on either a "first come first serve" basis or based upon qualification results or rankings in the WCA world rankings of a previously announced calendar date. If the number of competitors per event or per competition is limited, the following regulations apply:
+    - Z4a) If the limit for a competition or an event is reached before the competition, the organization team must not accept on-site registrations for the competition or the event.
+    - Z4b) Competitor limits should be announced together with the competition. Any changes to competitor limits are at the discretion of the WCA Board.
+    - Z4c) If there is more than one registration period, the organization team must announce each additional registration period at least 48 hours before it starts.
```
lgarron, line 527: This implies that you can announce on-site registrations up to 48 hours in advance, right?

lgarron (approved):

wca-regulations.md
Laura-O: Yes, right.

AlbertoPdRF: And how should that announcement happen? I think that the vast majority of people won't see it regardless of the method used.

I think that the Chinese system is clear and fair, so we shouldn't forbid it, but I don't feel like leaving the door open to re-opening registrations with an announcement 48 hours before is the right thing to do. My suggestion is to change Z4c) to something like:

> If there is more than one registration period, the organization team must clearly state when each additional registration period will happen on the competition website before the competition is announced.

viroulep: I don't have a strong opinion between the two options.
I think the main issue(s) with reopening registrations were stated by Laura in the PR description, and either way I feel like it seriously prevent abuse.

Laura-O: Yes, I considered something like this but I think it's too strict. There can be valid reasons to reopen the registration although it wasn't planned before. It's sometimes difficult to predict registrations/cancellations, so this is not necessarily bad planning.

Additionally, forbidding any additional registration periods would also cause problems when the competitor limit is raised after the registration was closed.

AlbertoPdRF: Ok, I'm fine with leaving that door open.

Now that I read Z4b) again, shouldn't it be "competitors limits" instead of "competitor limits"?

Laura-O: Why do you think so?
It's also "Competitor Limit" on the competition pages.

Claster: I don't get it: you can advance on-site registration 48 hours in advance, but then Z4a says it is not allowed. Isn't it a contradiction here?

Laura-O: Can you explain where you see the problem? Where is the contradiction?

Z4a does only forbid on-site registrations when the competitor limit was reached.

Claster: Ah, I see. And if the limit was once reached, but then several competitors dropped out, the limit is not considered as reached anymore, right? (The Chinese system)
Sounds good then.

viroulep:
It seems the main discussed point was agreed on, and the changes look good to me too, so I think we can go ahead and merge this.
@Laura-O it looks like the changes conflict with a previous PR, could you please fix it before someone can merge the PR?

Laura-O:
> @Laura-O it looks like the changes conflict with a previous PR, could you please fix it before someone can merge the PR?

That's weird. I already did this 5 days ago and it doesn't show any conflicts for me?

viroulep:
I don't get it, it doesn't now either for me, but yesterday clearly a conflict was showing...
Maybe my browser had this in cache and didn't properly update the page.

Merging then!

bruchemrfe:
I think we should split the first part of Z4. There are two topics covered in one regulation:
1) can there be competitor limits and which rules apply?
2) can there be qualification times?
Now there is no "first come, first served" for people who meet the qualification limit.
For example in World Championship there are 110 competitors who meet the qualification limit. But the competitor limit is 100. So there should also be first come first served for people who meet the qualification limit.
This can be even more complicated if I am already accepted, but then someone who registered earlier than me also made the qualification limit.

I think we should cover the qualifications separately. It is normal to have competitor limits. It is rare to have qualification limits.

bruchemrfe:
it is still not 100% clear in Z4a regarding the difference between competitors and registered persons. Some registered persons may not show up, so that would almost always mean fewer competitors than the limit. Can we accept more registered persons or should we consider all registered persons a competitor?

bruchemrfe:
Z4b) alternative text:
Competitor limits should have been published when announcing the competition.

bruchemrfe:
what happened to this one?
Z4c) If the registration for a competition has been closed while the competitor limit was reached, the registration should not be reopened. 
We had a discussion on this one and you convinced me it is unfair to reopen.

PR #503, opened 2017-11-10
Does Article Z allow for multiple qualifying times per event? #428

### LINKED ISSUE #428: Does Article Z allow for multiple qualifying times per event?
For example, US Nationals 2017 [announced the following qualification for FMC](https://www.cubingusa.com/nationals2017/qualifying.php):

```
FMC	35 single OR 45 mean
```

Is this actually allowed by [Z4](https://www.worldcubeassociation.org/regulations/#Z4)?

AlbertoPdRF:
Fixes #428.

This is my proposal, let's discuss it :)

### DIFF
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -210,4 +210,4 @@ To be more informative, each Guideline is classified using one of the following 
 
 ## <article-Z><optional><optional> Article Z: Optional Regulations
 
-- Z4+) [CLARIFICATION] In case of a competitors limit based upon qualification results or rankings in the WCA world rankings of a previously announced calendar date, the organization team must use only one type of result (i.e. single, mean, or average).
+- Z4+) [CLARIFICATION] In case of a competitors limit based upon qualification results or rankings in the WCA world rankings of a previously announced calendar date, the organization team must use only one type of result (i.e. single or average/mean).
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -206,3 +206,8 @@ To be more informative, each Guideline is classified using one of the following 
 - H1b1++) [ADDITION] The judge may permit the competitor to continue the attempt unofficially, but the attempt must be stopped and judged first, in full accordance with the Regulations.
 - H1d+) [EXAMPLE] Example: Suppose a competitor attempts 10 cubes, stops the solve with a time of 59:57, and has two time penalties. The final result is 59:57 + 2*2 = 60:01 (also see [Regulation A1a5](regulations:regulation:A1a5)).
 - H1d++) [EXAMPLE] Example: Suppose a competitor attempts 10 cubes, the judge stops the competitor at 60 minutes, and the attempt has two time penalties. The final result is 60:00 + 2*2 = 60:04.
+
+
+## <article-Z><optional><optional> Article Z: Optional Regulations
+
+- Z4+) [CLARIFICATION] In case of a competitors limit based upon qualification results or rankings in the WCA world rankings of a previously announced calendar date, the organization team must use only one type of result (i.e. single, mean, or average).


wca-guidelines.md
```diff
@@ -206,3 +206,8 @@ To be more informative, each Guideline is classified using one of the following
 - H1b1++) [ADDITION] The judge may permit the competitor to continue the attempt unofficially, but the attempt must be stopped and judged first, in full accordance with the Regulations.
 - H1d+) [EXAMPLE] Example: Suppose a competitor attempts 10 cubes, stops the solve with a time of 59:57, and has two time penalties. The final result is 59:57 + 2*2 = 60:01 (also see [Regulation A1a5](regulations:regulation:A1a5)).
 - H1d++) [EXAMPLE] Example: Suppose a competitor attempts 10 cubes, the judge stops the competitor at 60 minutes, and the attempt has two time penalties. The final result is 60:00 + 2*2 = 60:04.
+
+
+## <article-Z><optional><optional> Article Z: Optional Regulations
+
+- Z4+) [CLARIFICATION] In case of a competitors limit based upon qualification results or rankings in the WCA world rankings of a previously announced calendar date, the organization team must use only one type of result (i.e. single, mean, or average).
```
lgarron: Nit: What if an event in the past has had both "mean" and "average". Can a competition accept either?

AlbertoPdRF: I would like to only accept the current format, although for the first months after a format change for an event maybe that's a problem.

lgarron: Right, this will affect feet. How about a guideline specifically about that?

jfly: What about keeping things simple and saying you can only set qualifications based on the preferred format? IMO, organizers shouldn't be setting qualification times that are unreasonable for people to achieve (and setting a qualification time for feet that is an average 5 of ridiculous because nobody has had a chance to do that yet).

viroulep:
The idea behind US nats having a single or mean qualification limit (for FMC) was to ensure even people who didn't have the opportunity of doing a mean could qualify.

The regional FMC competition are getting popular and held basically across all continents now, so most probably the need for the "single or mean" qualification will probably decrease (supporting the idea of having only one type of qualification times).

I'd love to get some thoughts from the US nats team to see if they have some other reasons to keep it.
Maybe @KitClement, @timreyn can comment on this.

And I'm all for removing the "single and mean/average" qualifications.

wca-guidelines.md
Laura-O: Why not just create two categories: single OR average/mean?

Qualification limits for means and averages of an event would probably not be different anyway.

jfly: Upon further reflection, I like @Laura-O's suggestion to just let the delegates choose exactly one of single, average, or mean.

timreyn: I'm in favor of treating average and mean as a single bucket -- as @Laura-O said, single OR average/mean.  The WCA rankings treat average and mean as interchangeable: people are still in the rankings for 4x4 average with Mo3, and when feet changes to Ao5, the world record will still be a Mo3 for a while.

This would be consistent with the rest of the way that the WCA treats averages versus means: treat already-completed means the same as already-completed averages, but restrict which formats can be used moving forward.

timreyn:
Regarding US Nats, this was Kit's suggestion to have separate qualification times, so I'll defer to him for the official stance.  My personal opinion is that the frequency of FMC-only competitions makes this use case not worth supporting, which may not have been the case 1 year ago.

Claster (approved):

Laura-O:
LGTM!

viroulep:
Awesome :)

AlbertoPdRF:
Oh, sh*t. This has "competitors limit" instead of "competitor limit". I'll make another PR.

Issue #458, opened 2017-08-24
Pillowed cubes revisit


Mollerz:
https://www.worldcubeassociation.org/regulations/#3h1

I don't see why this is a problem anymore. They were originally only disallowed because people believed they shared the same or similar traits to stickerless cubes, however these are allowed now. I don't see any reason why we shouldn't allow pillowed puzzles? We have concave puzzles, for example the wingy skewb, the concave megaminx. We allow bubbled stickerless cubes like the YuHu megaminx. Removing this regulation doesn't seem like a detriment.

Laura-O:
Duplicate of #195, please continue the discussion there.

PR #504, opened 2017-11-10
Allow pillowed puzzles Fixes #195

### LINKED ISSUE #195: Pillowed Puzzles
Before 2009, pillowed puzzles were not allowed, and most (if not all) manufactured puzzles were cubic regardless.  With the introduction of V-Cubes in 2008 and the addition of 7x7x7 as an event in 2009, the pillowed V7 was the only puzzle available for this event.  Today, pillowed versions of virtually every cubic puzzle exist, and these are not allowed in competition, with the longstanding exception of 7x7x7.  
## Reasons for Allowing
- While the majority of cubers prefer the new cubic Shengshou 7x7x7 cube, many cubers (including the current single WR holder, Lin Chen) prefer to use the pillowed V-Cube.  Not allowing any pillowed puzzles will have a significant competitive impact on these 7x7x7 cubers.
- The advantages of a pillowed puzzle over a cubic one are arguably minimal.  These include:
  - A visual advantage (parts of 5 sides potentially visible)
  - A gripping advantage (potentially easier to hold around a pillowed puzzle compared to a cubic one)
- The new MoYu 7x7x7 that is soon to be released will be pillowed.  Given MoYu's track record with puzzle quality, this will be a widely popular puzzle, and banning pillowed puzzles will be widely unpopular.
- Consistent regulations - allowing some cubic puzzles to be pillowed while others must be cubic makes the spirit of this regulation rather unclear.
## Reasons for Disallowing
- While it seems technical, we call our puzzles "cubes," and pillowed puzzles do not look like cubes.  Practically, this may give those who prefer pillowed puzzles an advantage visually as well as with their grip.
- An "anything goes" policy on pillowed cubes would open the door to many other puzzles that have never been WCA legal - including fully spherical puzzles.  Do we need to draw a line for how much pillowing is allowed?
- As stated in the reasons for allowing, disallowing all pillowed puzzles would also be consistent.

Laura-O:
That's another issue that has not been solved for many years now.

@Mollerz revived this discussion in #458: we allow concave puzzles and several other modifications, so I think we are close to an "anything goes"-policy. I think forbidding pillowed puzzles is not reasonable anymore.

@thewca/wrc-team Let me know what you think.

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -122,9 +122,8 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - 3d1a) Blind competitors may use textured puzzles with different textures on different faces. Each face should have a distinct color, to aid in scrambling and judging.
         - 3d1b) Color blind competitors who cannot distinguish between the necessary number of colors may use colored parts with patterns, if it has been explicitly approved according to [Regulation 2s](regulations:regulation:2s). Patterns may come from stickers or be drawn.
     - 3d2) The colors of the colored parts must be solid, with one uniform color per face. Each color must be clearly distinct from the other colors.
-    - 3d3) If the colored parts are stickers or tiles, they must not be thicker than 1.5mm. Exception: Megaminx tiles may be slightly thicker than 1.5mm, at the discretion of the WCA Delegate.
 - 3h) Modifications that enhance the basic concept of a puzzle are not permitted. Modified versions of puzzles are permitted only if the modification does not make any additional information available to the competitor (e.g. orientation or identity of pieces), compared to an unmodified version of the same puzzle.
-    - 3h1) "Pillowed" puzzles are not permitted. Exception: Pillowed 7x7x7 Cubes are permitted.
+    - 3h1) "Pillowed" puzzles are permitted.
     - 3h2) Puzzles whose colored plastic is visible inside the puzzle (e.g. "stickerless" puzzles) are permitted. This does not include the following puzzles:
         - 3h2a) Puzzles with transparent parts. Exception: An overlay sticker (see [Regulation 3l](regulations:regulation:3l)).
     - 3h3) Any modifications to a puzzle that result in poor performance by a competitor are not grounds for additional attempts.


AlbertoPdRF:
I would prefer to ban all pillowed cubes instead, but I don't have a strong opinion here.

SAuroux:
I support allowing pillowed puzzles as this (1) further simplifies the regulations and (2) we all know that pillowed puzzles do not provide any real advantage (unlike stickerless cubes, which we already allow).

I like @Laura-O 's phrasing: forbidding pillowed puzzles is not reasonable anymore.

lgarron:
> I support allowing pillowed puzzles as this (1) further simplifies the regulations and (2) we all know that pillowed puzzles do not provide any real advantage (unlike stickerless cubes, which we already allow).

A pillowed cube can allow seeing all 4 sides of PLL from the top, which feels like we should not enable – but I'm coming around to the idea that this doesn't fundamentally change the puzzle and most competitors will not use it that way (just like people don't intentionally look for stickers on the back of stickerless cubes, nor set outrageously faster times on them – and magnets haven't fundamentally changed the game either).

My interpretation is that will also allow the [sphere](http://twistypuzzles.com/cgi-bin/puzzle.cgi?pkey=579):


I agree with trying this change for now. We can always respond if we see crazy new puzzles on the market.

Optional ideas:
- Add a Guideline of the form "3h1+) Pillowed puzzles were not permitted in the past. They are permitted now.", at least for a year.
- Provide a guideline for how pillowed is reasonable, e.g. "each face must not vary more than 1cm from a plane".

kingmathyall:
I agree that pillowed cubes should be allowed as I truly don't think it gives an advantage and if it does, it's as minuscule as the "advantage" that comes with stickerless cubes.

However, if we allow pillowed puzzles, would what this mean for Regulation 3d3 referring to the thickness of tiles. My interpretation of this regulation is so that stickers aren't thick enough to the point where you can see the color in the same way you might see the color of a side on a pillowed puzzle.

Laura-O:
> Provide a guideline for how pillowed is reasonable, e.g. "each face must not vary more than 1cm from a plane".

I don't think this is even necessary. Even if someone wants to use spheric puzzles, advantages are so low compared to the disadvantages.

I think this regulation change is mainly interesting for new competitors, who think the pillowed puzzles look nicer. I can remember a handful of situations when newcomers wanted to use pillowed 2x2-4x4 V-Cubes here (the German distributor only sells pillowed puzzles...). For the rest of the competitors, this is probably just not relevant.
But in general, I think it's a good moment to show that the WCA is open to all kind of puzzles. :)

> However, if we allow pillowed puzzles, would what this mean for Regulation 3d3 referring to the thickness of tiles. My interpretation of this regulation is so that stickers aren't thick enough to the point where you can see the color in the same way you might see the color of a side on a pillowed puzzle.

I'm completely fine with removing that as well. :-)

AlbertoPdRF:
> I'm completely fine with removing that as well. :-)

Should we do that on this PR too?

Claster:
I would prefer to ban pillowed 7x7 as well, but don't have a strong preference.

I think if we just delete 3h1, they WILL be banned though, as they "make any additional information available to the competitor", compared to cubic puzzles, thus they violate 3h. So if you want to allow them, you need to clarify that explicitly.

kingmathyall:
So, shall we get rid of 3d3 regarding to tile/sticker thickness and change 3h1 to "Pillowed puzzles are permitted"?

Laura-O:
Yes, I think that's the way to go, it's updated.

Some more reasons for removing 3d3:

- Puzzles with real tiles (thick stickers) are not popular anymore. I don't think there were any popular speedcubes with tiles in the last 2-3 years.
- Thick tiles, which could make additional information available, are still forbidden because of [3h](https://www.worldcubeassociation.org/regulations/#3h)

Goodbye 3d3, you were one of the reasons why I started to comment on WRC issues in 2014. :cry:

Claster (approved):

AlbertoPdRF:
LGTM!

Issue #484, opened 2017-10-29
9a1: What are "Rubik's puzzles"?


jfly:
Regulations [9a1](https://www.worldcubeassociation.org/regulations/#9a1) says:

> 9a) The WCA governs competitions for: 9a1) Puzzles known as Rubik's puzzles.

What are "Rubik's puzzles"? There are a number of puzzles/toys out there that have "Rubik" in the name that are not actually governed by the WCA, such as: magic, master magic, snake, and probably other things I don't know about. Do we think this regulation actually has any value as it currently is?

Laura-O:
Good point!

To fix this we could summarise both bullets in 9a:  _The WCA governs competitions fo puzzles that are manipulated by twisting the sides, commonly known as "twisty puzzles"_

The mission of the WCA is pretty similar btw: "_The World Cube Association governs competitions for all puzzles labelled as Rubik puzzles, and all other puzzles that are played by twisting the sides, so-called 'twisty puzzles'_.

jfly:
Oh wow! I had no memory of "Rubik puzzles" being in the mission.

In fact, if you look at [WCA Motion: 1.2017.1 - Spirit and Mission](https://docs.google.com/document/d/1VHGJHFYA5QM31_aPExRavzfWFqmI-Jn0a3jxWHoUS9A/edit), you'll see that the mission of the WCA is simply "The Mission of the World Cube Association is to have more competitions in more countries with more people and more fun, under fair and equal conditions.". Our website is apparently out of date here: https://www.worldcubeassociation.org/about, because it has a 4 paragraph mission with a "goal" and a "spirit". I'll ping @pedrosino to ask him to take a look as a Board member.

As something of an aside, it's kind of annoying that our mission on the website talks about "Rubik Puzzles", and our regulations talk about "Rubik's Puzzles". At the very least, we should make them consistent with each other.

pedrosino:
I believe we should replaced all mentions to Rubik's in the regulations and other documents.

We currently have a lot of them:

>9a) The WCA governs competitions for:
>__9a1) Puzzles known as **Rubik's** puzzles.
>__9a2) Other puzzles that are manipulated by twisting the sides, commonly known as "twisty puzzles".
>9b) The official puzzles and event formats of the WCA are:
[...]

PR #510, opened 2017-11-17
Remove "Rubik's" in various places.

pedrosino:
The Board has discussed this, and we think it's best if we keep Rubik's Cube only at 9a1, as a direct mention to the puzzle.


Laura-O:
@pedrosino This confuses me. As I mentioned in #484, we already have #479, which fixes #405.
Simply speaking: we already have a PR which fixes the event names according to our official event names.
If we add another change, this should be done in that PR, so I am closing this now.

pedrosino:
Sorry, I didn't remember that one.
Will comment there, as there are some differences which would need to change.

PR #511, opened 2017-11-20
Clarify C1b Fixes #464

### LINKED ISSUE #464: Rule C1 might be too strict.
Good evening. 
I would like to make a complaint about rule C1c, due to an incident that occurred to me in the final round of the 3x3 one-handed event at Verona open (September 24th 2017). I was on my third solve and after a really good start I dropped my cube due to a lockup; the cube fell on my right hand (I solve with left) which was placed still on the stackmat timer. I then proceeded in picking up the cube with my left hand and kept on solving; it ended up being the second 13s solve in the average (this would have given me a really good shot on getting NR average which is 14.8s). The judge appealed to rule C1c that states: "During the solve, once a competitor touches the puzzle with one hand, they must not touch the puzzle with the other hand. Penalty: disqualification of the attempt (DNF)." Because of my right hand, the cube rolled further away from me (giving me a major disadvantage in time waste); I didn't attempt to use my right hand to stop the cube because I thought that C1c concerned only intentional gestures, and I didn't want the penalty to occur. I let everything happen.
I then asked the delegates' help (there were Lorenzo Vigani Poli, Matteo provasi and Simone Cantarelli) to solve this problem: Lorenzo seemed to agree with my claims, but was reminded by Matteo that rule C1c doesn't allow exceptions and that in past occasions similar cases were ruled as DNFs (this was later confirmed by old reports), so he told me "even if you were right, these are the rules, so we must play by them"; Simone wasn't even summoned because we felt there was no need. My solve was converted into a DNF: because of 2n3 I accepted the decision and went on with the average, very upset and distracted, and ended up getting very bad times.
I think C1c should be revisited in order to prevent incidents like this from happening in the future: I didn't attempt to cheat, nor did I get any unfair advantage form it. Rule C3b1 states that "If a puzzle defect occurs, and parts of the puzzle briefly come in contact with other body parts without the competitor's intention, this is not considered touching the puzzle, at the discretion of the judge." Isn't a very bad lockup a puzzle defect? maybe if I had been solving with two hands, the same lockup would have ended up in a pop (two hands, double the force and more intense the grip); I didn't have any way to prevent the puzzle form hitting me. Moreover, rule 5a gives "Examples of puzzle defects", not all of them, so a cube drop might as well be taken into consideration. I feel that if the rules are there to make competitions fair for everyone, things like this should not happen: I felt robbed and cheated (not by someone in particular), I hope we can make one handed better and more enjoyable for everyone. There are very specific rules in the WCA Regulations, that were shaped with time; I feel like this one should be developed as well since it can be misleading and can appear to be irrational.
I'm sorry if I made any mistakes or if I sounded inappropriate: english is not my main language. I hope you can take this into consideration.
Best regards.

Tommaso Raposio (WCAID 2014RAPO01)

Laura-O:
I propose to not add C1b4) because I think the situation is already covered by C1b3.
We could add another guideline, but I think that's not necessary.

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -458,10 +458,10 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 ## <article-C><one-handed><onehandedsolving> Article C: One-Handed Solving
 
 - C1) Standard speed solving procedures are followed, as described in [Article A](regulations:article:A) (Speed Solving). Additional regulations that supersede the corresponding procedures in [Article A](regulations:article:A) are described below.
-    - C1b) During the solve, the competitor must use only one hand to touch the puzzle. Penalty: disqualification of the attempt (DNF).
+    - C1b) During the solve, the competitor must use only one hand to operate the puzzle. Penalty: disqualification of the attempt (DNF).
         - C1b2) If a puzzle defect occurs, and the competitor chooses to repair it, they must repair it using only the solving hand. Penalty: disqualification of the attempt (DNF).
-        - C1b3) If a puzzle defect occurs, and parts of the puzzle briefly come in contact with other body parts without the competitor's intention, this is not considered touching the puzzle, at the discretion of the judge.
-    - C1c) During the solve, once a competitor touches the puzzle with one hand, they must not touch the puzzle with the other hand. Penalty: disqualification of the attempt (DNF).
+        - C1b3) If other body parts of the competitor come in contact with the puzzle or parts of the puzzle without the competitor's intention and without applying moves, this is not considered operating the puzzle, at the discretion of the judge.
+    - C1c) During the solve, once a competitor operates the puzzle with one hand, they must not operate the puzzle with the other hand. Penalty: disqualification of the attempt (DNF).
 
 
 ## <article-D><feet><solvingwithfeet> Article D: Solving With Feet


wca-regulations.md
```diff
@@ -458,10 +458,10 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 ## <article-C><one-handed><onehandedsolving> Article C: One-Handed Solving
 
 - C1) Standard speed solving procedures are followed, as described in [Article A](regulations:article:A) (Speed Solving). Additional regulations that supersede the corresponding procedures in [Article A](regulations:article:A) are described below.
-    - C1b) During the solve, the competitor must use only one hand to touch the puzzle. Penalty: disqualification of the attempt (DNF).
+    - C1b) During the solve, the competitor must use only one hand to manipulate the puzzle. Penalty: disqualification of the attempt (DNF).
```
lgarron: I strongly prefer "operate" to "manipulate" (as in [9a](https://github.com/thewca/wca-regulations/commit/2a8bb4bab952cecad30f4a74d5d0e90defdb7930)). How would you feel about using that throughout this PR?

wca-regulations.md
```diff
@@ -458,10 +458,10 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
...
         - C1b2) If a puzzle defect occurs, and the competitor chooses to repair it, they must repair it using only the solving hand. Penalty: disqualification of the attempt (DNF).
-        - C1b3) If a puzzle defect occurs, and parts of the puzzle briefly come in contact with other body parts without the competitor's intention, this is not considered touching the puzzle, at the discretion of the judge.
-    - C1c) During the solve, once a competitor touches the puzzle with one hand, they must not touch the puzzle with the other hand. Penalty: disqualification of the attempt (DNF).
+        - C1b3) If the puzzle or parts of the puzzle briefly come in contact with other body parts without the competitor's intention and without applying moves, this is not considered manipulating the puzzle, at the discretion of the judge.
```
lgarron: Nit: grammatically, the puzzle is the subject of this sentence, and a puzzle doesn't apply moves. Taht can be fixed by changing to active voice:

>  If other body parts of the competitor come in contact with the puzzle or parts of the puzzle without the competitor's intention and without applying moves, this is not considered operating the puzzle, at the discretion of the judge.

lgarron (changes requested):

wca-regulations.md
Laura-O: As @kingmathyall mentioned in the related issue, "manipulate" is also used in other rules. So I think we should either change all occurrences or leave it as it is.

Laura-O:
This is fixed and I changed manipulate to operate.
I will create another PR to change "manipulate" in the Feet regulations.

Claster (approved):

viroulep:
I'm fine with these changes, @lgarron can we go ahead and merge this?

Laura-O:
@lgarron?

Claster:
Probably Lucas was so impressed by the perfect wording that he remains speechless :D
Merging.

PR #512, opened 2017-11-20
Clarify similar Clock pieces, fixes #409

### LINKED ISSUE #409: Revisit "similar pieces" on Clock
Currently we have:

> 3j) Puzzles must be clean, and must not have any markings, elevated pieces, damage, or other differences that significantly distinguish any piece from a similar piece. Exception: a logo (see Regulation 3l).
> 3j2+) CLARIFICATION On Clock, all eighteen inner clock faces are considered similar pieces.

That implies that all 18 inner clocks must be the same color. This, however, not always the case (probably almost never the case?): all clocks on one side have one color, but these two sides differ.

It would make sense to consider inner clocks only on one side as similar pieces. Or get rid of it completely, allowing inner clocks to have different colors. It becomes easier to distinguish top side and to avoid a common mistake when competitors "solve" a Clock with all clocks pointing at 6 or 3 or 9 instead of 12. However, this advantage can be gained with currently legal ways, just by writing "this is up" on an insert.

Laura-O:


viroulep:
LGTM!

Claster:
Sorry for being picky, but several regulations (4d4, 3j++) refer to a face of a clock as a "side"; and 10h1 calls inner clocks "inner clock faces".

> On Clock, the nine inner **clock faces** on one **side** are considered similar pieces.

Laura-O:
Yep, that was dumb c&p, sorry about that. It's fixed now.

Claster:
lgtm, merging

PR #513, opened 2017-11-20
Clarify 'the delegate', fixes #499

### LINKED ISSUE #499: Clarify that "the WCA Delegate" applies refers to all official Delegates of the competition.
From Regulation 1c:

> The WCA Delegate **may delegate responsibilities** to other members of the organization team, but **is ultimately accountable** for how these responsibilities are carried out. 

In fact, "the WCA Delegate" here and elsewhere refers to anyone fulfilling the Delegate role in a particular capacity. By default, this includes just the official Delegates of the competition, but many responsibilities (e.g. checking for legal puzzles) are formally or informally delegated to other staff members at many competitions.

From https://github.com/thewca/wca-regulations/pull/492/files#r148676866 :

> @Laura-O:
> I get Lucas' point, but there are several regulations which suggest that a competition can only have one delegate (like 1c), so this is an essential inconsistency.
> A workaround could be to use "the WCA Delegate" here and add a clarification for 1c that the WCA Delegate refers to all WCA delegates of the competition.

Another option would be to use another way to refer to "someone acting with the Delegate's responsibility" or "someone authorized by the Delegate" every time. But I think I would prefer to keep the clarity of the "the WCA Delegate" and add a Guideline.

Also see https://github.com/thewca/wca-regulations/pull/498

Laura-O:
Do we need a further clarification that this is not only a clarification for 1c?
I think it's fine as 1c defines the responsibilities, so any other mention refers to 1c, but I'm also fine with adding one if someone has a good idea of the wording.

### DIFF
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -32,6 +32,7 @@ To be more informative, each Guideline is classified using one of the following 
 
 ## <article-1><officials><officials> Article 1: Officials
 
+- 1c+) [CLARIFICATION] "The WCA Delegate" refers to anyone fulfilling the role of a WCA Delegate at a competition.
 - 1c3+) [RECOMMENDATION] Results should be ready at the end of the last day of competition.
 - 1c3++) [ADDITION] Physical records (e.g. score sheets, written solutions from Fewest Moves Solving) must be retained for a minimum of one month. Digital records (e.g. scramble sequences, results) must be retained permanently.
 - 1c3b+) [CLARIFICATION] If there are multiple groups, it is not necessary to identify which competitor was in which groups.


wca-guidelines.md
```diff
@@ -32,6 +32,7 @@ To be more informative, each Guideline is classified using one of the following
 
 ## <article-1><officials><officials> Article 1: Officials
 
+- 1c+) [CLARIFICATION] "The WCA Delegate" refers to anyone fulfilling the role of the WCA Delegate at a competition.
```
viroulep: Maybe remove the last "the", so that it's "fulfilling the role of WCA Delegate"?
To me "the WCA Delegate" still sounded like there is only one Delegate at a competition :p

Laura-O: Good point, it's changed now.

viroulep (approved):

Claster (approved):

PR #516, opened 2017-11-23
Forbid logos in blindfolded events and clarify 3k

Laura-O:
As discussed in our email thread, here is a first draft:

- Changed 3l so that all logos for blindfolded events are forbidden.
- 3l2: Remove the exception for blindfolded events.
- 3k: 'must' -> 'should' to weaken the role of the delegate
- 3k1: Remove "Earlier results remain valid, at the discretion of the WCA Delegate."
- Add 3k2: define a penalty for attempts done with illegal cubes (DNF) and add an exception to grant extra attempts during a round.

Why?
The "distinguishable by feel"-exception has always been interpreted differently. Some delegates allow specific cubes, some don't. For example, the logo of the RSC has been discussed many times after WC17... 

Furthermore, it has always been interpreted as the responsibility of the delegate to check and approve puzzles. The regulations do not define a penalty for the use of illegal puzzles, so it might be not clear for competitors that they can be punished for using illegal puzzles. This makes it also difficult and painful to decide about incidents as the responsibility is shared between delegate and competitor. A clear definition of the penalty will hopefully solve this.

The WCA has been moving towards an "anything goes"-policy in the last 2-3 years and this change is pretty much the opposite. However, with the currently popular hardware that should not be a huge problem:
- Most puzzles have overlay stickers, which had to be removed with the prior regulations anyway.
- There are some stickerless puzzles with printed/embossed logos. These logos can be distinguished by feel, so they were also not allowed before.
- This is indeed a problem for competitors using individual stickers/logos, but this is clearly a minority.

### DIFF
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -75,6 +75,7 @@ To be more informative, each Guideline is classified using one of the following 
 - 3j++) [CLARIFICATION] On Clock, pins must not be distinguishable from any other pin of the same side.
 - 3j2+) [CLARIFICATION] On Clock, all eighteen inner clock faces are considered similar pieces.
 - 3l+) [ADDITION] Logos may feature any reasonable design that does not give a conspicuous advantage (e.g. encoding information that could be used to cheat), and leave the colored part clearly recognizable. Unconventional logos are only permitted at the discretion of the WCA Delegate.
+- 3l++) [CLARIFICATION] In the past, puzzles with logos have been permitted for blindfolded events. Such puzzles are no longer permitted.
 
 
 ## <article-4><scrambling><scrambling> Article 4: Scrambling
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -133,8 +133,9 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 3j1) Puzzles are permitted to have reasonable wear, at the discretion of the WCA Delegate.
     - 3j2) Definition: Two pieces are similar to each other if they are identical in shape and size, or mirrored in shape and identical in size.
     - 3j3) Corrugated/textured parts which permit the orientation of pieces to be distinguished by feel are not permitted for blindfolded events.
-- 3k) Puzzles must be approved by the WCA Delegate before use in the competition.
-    - 3k1) If a puzzle is found to be not permitted during a round, the competitor must submit a replacement. Earlier results remain valid, at the discretion of the WCA Delegate.
+- 3k) Puzzles should be approved by the WCA Delegate before use in the competition.
+    - 3k1) If a puzzle is found to be not permitted during a round, the competitor must submit a replacement.
+    - 3k2) Penalty for attempts done with not permitted puzzles: disqualification of the attempt (DNF). Exception: If a puzzle is found to be not permitted during a round, earlier results may be replaced with an extra attempt, at the discretion of the WCA Delegate.
 - 3l) A puzzle may have a logo on a colored part. If it does, it must have at most one colored part with a logo. Exception: For blindfolded events, a puzzle must not have a logo.
     - 3l1) The logo must be placed on a center piece. Exceptions for puzzles that do not have center pieces:
         - 3l1a) For Pyraminx and 2x2x2, the logo may be on any piece.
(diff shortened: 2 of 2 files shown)


Claster:
Generally it looks good to me.

1) I would add a clarification to Guidelines: "Previously logos were allowed for bld, but not now :P". We could live without it as well.
2) I find the condition "if found during a round" a bit harsh. I would relax it to "if found until the start of the next round or until the end of the comp if the round was last".

Laura-O:
> I would add a clarification to Guidelines: "Previously logos were allowed for bld, but not now :P". We could live without it as well.

That's a good idea, added.

> I find the condition "if found during a round" a bit harsh. I would relax it to "if found until the start of the next round or until the end of the comp if the round was last".

Yes, I thought about this part as well.
But what is "the next round"? The next round after the current one? Or the next round of the event?
The next round after the current wouldn't make a difference if it's not the last round of the day. Sometimes that's also difficult to define when rounds are scheduled in parallel or overlapping.
The next round of the event makes things more complicated as the competitor would have to do the extra attempts during a different event, the delegate has to generate new scrambles, etc.

I also think most of these incidents are discovered during the round in progress and these are probably also the cases for which this regulation is particularly useful (e.g. inexperienced competitors using a puzzle with two stickers or competitors unintentionally using a cube with an overlay sticker in BLD).

@thewca/wrc-team What's your opinion?

kingmathyall:
I like all of these ideas! Gets rid of any confusion whatsoever with logos!

Claster:
I meant "until the start of the next round of the current blindfolded event". Well, I don't have a strong preference and am fine with "during current round" as well.

LGTM.

viroulep:
"During the current round" seems to match what happened to the competition I've been to, so I'm fine with this!

(and obviously I'm fine with the idea of this PR in general, as we discussed it a bit on the WRC mailing list)

Claster:
Merging.

Jambrose777:
> Add 3k2: define a penalty for attempts done with illegal cubes (DNF) and add an exception to grant extra attempts during a round.

Does this leave the door open for abuse? The example Bob gave was that if someone triple DNFed, then present the cube with an overlay to the delegate to be awarded 3 more attempts?

Baiqiang:
I have a question. If a competitor uses something like nail polish to wash a printed logo (e.g. valk 3), there may be some scratches on it. Do we allow it?

Claster:
@Jambrose777, good question. This is similar to repeated scrambles, something that can be caused by an official mistake that should not happen in the first place, and something that can be exploited by malicious competitors. I think, delegates should carefully apply "at the discretion of the delegate" part, in particularly evaluating whether they trust that competitor enough.

@Baiqiang, if it can be felt then no. This is applied to every sticker, not just the one that had a logo. This is covered by 3j.

pedrosino:
I think the text on 3k is not consistent with the rest of the regulations.
For all penalties, we have something like:
> n) This should happen. Penalty: something

While for this we have a separate sub-article with just the penalty.
I'm trying to think of a rewriting to make it more aligned with the others.

Laura-O:
> While for this we have a separate sub-article with just the penalty.

Yes, that's because the penalty is not for 3k, but for a consequence (3k1).

The only difference is that we don't use an introductory sentence here. But as it should be clear without one, I think that's not a problem.

pedrosino:
I have two suggestions. They are basically an attempt to make the text format consistent with the rest:

First:

> 3a) Competitors must provide their own puzzles for the competition.
.   3a1) Competitors must be ready to submit their puzzles when they are called (see Regulation 2u).
.   3a2) Puzzles must be fully operational, such that normal scrambling is possible.
.   3a3) Polyhedral puzzles must use a color scheme with one unique color per face in the solved state. Each puzzle variation must have moves, states, and solutions functionally identical to the original puzzle.
.   _**3a4) Puzzles must be compliant with the conditions defined further in this article. Penalty for attempts done with not permitted puzzles: disqualification of the attempt (DNF).**_
(...)
3k) Puzzles should be approved by the WCA Delegate before use in the competition.
.   _3k1) If a puzzle is found to be not permitted during a round, the competitor must submit a replacement. In this case, earlier results may be replaced with an extra attempt, at the discretion of the WCA Delegate._


Second:

> 3k) _**Competitors are responsible for providing [or “must provide”] puzzles that are compliant with the conditions defined in this article. Penalty for attempts done with not permitted puzzles: disqualification of the attempt (DNF).**_
.   3k1) Puzzles should be approved by the WCA Delegate before use in the competition.
.   _3k2) If a puzzle is found to be not permitted during a round, the competitor must submit a replacement. In this case, earlier results may be replaced with an extra attempt, at the discretion of the WCA Delegate._

PR #517, opened 2017-11-27
Add the Translation disclaimer

### LINKED ISSUE #362: Include the translation disclaimer in the source Regulations directly
Per the [translations instructions](https://github.com/cubing/wca-regulations-translations#translation-instructions), translators are supposed to add the following paragraph:

> Please note that this translation is not an official version (even though it is provided on the WCA website): it may be useful, but it does not come with any guarantees. If there is a difference between a translation and the (current official) English version, the English version must be used.

I've been meaning to place this paragraph into the appropriate place in the Regulations itself, so that translators can translate it just like the Regs. We can mark it so that it is not included in the English version.

Alternatively, we can change the phrasing so that it also makes sense in English, e.g. to refer to translations in general rather than "this translations".

viroulep:
Fixes #362.

I've made the necessary changes in the Regulations compiler to hide the "Translation" subsection, I'll mention this PR there.

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -5,6 +5,11 @@
 
 ## Notes
 
+### Translation
+
+Please note that translations are not official versions, even though they are provided on the WCA website for convenience.
+If there is a difference between a translation and the (current official) English version, the English version must be used.
+
 ### WCA Regulations and Guidelines
 
 The WCA Regulations contain the full set of Regulations that apply to all official competitions sanctioned by the World Cube Association.


wca-regulations.md
```diff
@@ -5,6 +5,11 @@
 
 ## Notes
 
+### Translation
+
+Please note that translations are not official versions (even though they are provided on the WCA website): they may be useful, but they do not come with any guarantees.
```
lgarron: Nit: I know I wrote the original version of this, but I think we can avoid parentheses, e.g.:

How's this?

> Please note that translations are not official versions, even though they are provided on the WCA website for convenience. If there is a difference between a translation and the (current official) English version, the English version must be used.

lgarron (approved):
LGTM, although I would prefer a wording update.

wca-regulations.md
viroulep: Updated :)

viroulep:
I updated the wording and the changes to ignore the subsection are now live in the last version of `wrc`, so this is ready!

PR #518, opened 2017-11-29
Clarify moves definition for square one.

### LINKED ISSUE #364: What metric do we use for Square One?
(Moving the discussion over from https://github.com/cubing/tnoodle/issues/153.)

See regulation [4b3d](https://www.worldcubeassociation.org/regulations/#4b3d), which calls for at least 11 moves on Square One and [12c](https://www.worldcubeassociation.org/regulations/#12c), which defines notation, but doesn't tell you what is considered a move.

I suppose right now, TNoodle is the de-facto standard for this. You can see the cost computation code here: https://github.com/cubing/tnoodle/blob/c6c0f6650e10d4ae76467532deb19a2504de9b67/scrambles/src/puzzle/SquareOnePuzzle.java#L216-L233, which says that something like `(5, 2)` is cost 1. However, Chen Shuang's Square One solver actually considers `(5, 2)` to be 2 separate moves. See https://github.com/cubing/tnoodle/blob/c6c0f6650e10d4ae76467532deb19a2504de9b67/sq12phase/src/cs/sq12phase/FullCube.java#L78-L84.

We could speed TNoodle Square One generation up quite a bit if we resolve this discrepancy. See TNoodle comment [here](https://github.com/cubing/tnoodle/blob/c6c0f6650e10d4ae76467532deb19a2504de9b67/scrambles/src/puzzle/SquareOnePuzzle.java#L31-L36).

viroulep:
Fixes #364.

The aim of this change is twofold:
 - Explicitly state the distinction between a "penalty" move and a "scrambling" move
 - Define what a "scrambling move" is

Feel free to comment both on the general implementation and nitpicking the wording!
I tried a version where I put 12c4 as two Guidelines (one for 12c2, one for 12c3), but I didn't like it, mainly because metric is defined directly in the Regulations for other puzzles.

Since 12c2+ is targeting "penalty" moves, I moved it around and made it a clarification, as 10e1+10f4 makes it clear that the misalignment refers to two adjacent faces.


wca-regulations.md
```diff
@@ -324,6 +324,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 12c1) Moves are applied with one of the two smallest surfaces of the equatorial slice on the left side of the front face.
     - 12c2) (x, y) means: turn upper layer x times 30 degrees clockwise, turn bottom layer y times 30 degrees clockwise.
     - 12c3) "/" means: turn the right half of the puzzle 180 degrees.
+    - 12c4) Metric for Square-1: (x, y) counts as one move, "/" counts as one move.
```
jfly: Uber nit: Do we care about `(0, 0)`?

jfly (commented):
This looks pretty good to me!

wca-regulations.md
viroulep: We can actually make 12c2 more precise, by specifying that x and y must be in [-5; 6], and that x and y cannot be both equal to 0.

Laura-O: Yes, clarifying this in 12c2 sounds good!

Laura-O:
Apart from the small improvement Jeremy suggested this LGTM

Claster:
LGTM now.

Claster:
Ok, I guess we have two LGTMs from WRC members, merging.

PR #522, opened 2017-12-16
A1a2++++: cannot -> must not

Laura-O:
Quick word fix


AlbertoPdRF:
LGTM!

viroulep:
I agree with the change as well, definitely better per the "can"/"must" definition we use!

PR #523, opened 2017-12-17
A7c2: make 2nd sentence consistent

Laura-O:
As suggested in the discussion on the delegate list, the second sentence is inconsistent and should be fixed.


AlbertoPdRF:
LGTM!

wca-regulations.md
```diff
@@ -425,7 +425,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
...
         - A7c1) If the competitor or the judge refuses to accept and sign the score sheet, the WCA Delegate must resolve the dispute.
-        - A7c2) The competitor must not sign an attempt on the score sheet before the judge has recorded the final result for that attempt and signed. If a competitor signs (or otherwise marks) an attempt before they have begun it, they forfeit the attempt and the result will be considered DNS. If the competitor signs (or otherwise marks) an attempt after starting the attempt but before the judge has finished recording the result, it will be considered DNF.
+        - A7c2) The competitor must not sign an attempt on the score sheet before the judge has recorded the final result for that attempt and signed. If a competitor signs (or otherwise marks) an attempt before they have begun it, they forfeit the attempt and the result will be considered DNS. If the competitor signs (or otherwise marks) an attempt after starting the attempt but before the judge has finished recording the result and signed, it will be considered DNF.
```
jfly, line 428: This construction feels a bit awkward (lack of parallelism between "recording" and "signed") and potentially ambiguous to me. Here's a rephrasing I consider bother easier to read and unambiguous:

> "... before the judge has finished recording and signing the result..."

Alternatively, here's a rephrasing with even fewer words that I also consider easy to read and unambiguous:

> ... before the judge has recorded and signed, ..."

jfly (commented):
Some comments for you on the English. Just to be clear, this still allows for an inexperience (or even malicious judge) to record a result, give the scorecard to the competitor, and then DNF the competitor because the competitor signed before the judge did, right?

I know for a fact that inexperienced judges at my competitions forget to sign, or sign after competitors. When I'm a competitor for them, I always ask them to sign before I sign, but I know that competitors in my region don't always do that.

pedrosino:
I agree with Jeremy.

Practical example:
A scorecard arrives at the score taking desk, and is found to be missing a judge's signature. You no longer can find the judge and confirm the time, and ask them to sign, because in this new wording the competitor broke the regulation, so it's an automatic DNF. I find that overly harsh and not what was meant in the first place.

PR #524, opened 2017-12-17
Improve A7c*

Laura-O:
That's an attempt to solve the problems with A7c and especially to implement what we wanted to achieve without any bad side-effects.

This change was partly developed by/with @SAuroux, thanks!


wca-guidelines.md
```diff
@@ -168,6 +168,7 @@ To be more informative, each Guideline is classified using one of the following
 - A6e+) [CLARIFICATION] If the puzzle falls into the lap of the competitor after stopping the timer, this is considered touching the puzzle.
 - A6g+) [ADDITION] While the judge is determining whether to assign a penalty for misalignment, they should not touch the puzzle.
 - A7c+) [ADDITION] If a result signed by a competitor is found to be incomplete or unreadable, it should be interpreted the worst reasonable way (e.g. "1:05" will be considered 1:05.99, "25.X3" will be considered 25.73 if X could be either a 1 or a 7). Given that the Stackmat timer displays a time of X:0Y.ss as X:Y.ss, a missing tens digit should be considered more likely than a missing ones digit (e.g. "1:2.27" will be considered 1:02.27). The judge who signed an ambiguous result may be requested to clarify it, at the discretion of the score taker.
+- A7c4+) [ADDITION] Missing signatures may be rectified if the responsible judge and/or competitor can be tracked, at the discretion of the score taker.
```
AlbertoPdRF, line 171: I would prefer "at the discretion of the Delegate" here.

Laura-O: Yes, we have discussed that as well. The guideline is pretty much following the approach in A7c+) which also leaves this to the score taker.
I also think that makes sense as the score taker is the person who finds these mistakes and has to take actions.

Claster: Should we keep the "tracked during the competition" part, as we communicated it to the Delegates some time ago?
Technically the Delegate who checks scorecards after the competition acts also as "score taker".

Laura-O: We also accepted results which were found after the competition (especially if there was a video proof). This would not be possible with a "during the competition"-restriction.

So what actually happens at the end of the competition which changes the situation? For me, restricting this to during the competition seems quite random.

Claster:
Ok then, LGTM.

viroulep:
Ok so if I sum up the changes here is some way to describe it to the Delegates:

 - A7b clarifies the whole point of signing
 - A7c2 and A7c3 describe what happens if a competitor signs before the judge has written a time (both are needed to not have a situation where we don't know what to do)
 - A7c4 defines a strongly preferred order for signature, put the responsibility of checking signatures are there on the competitor, and thus clarifies that any missing signature leads to DNF.
 - A7c4+ adds some flexibility in case competitors and judges can be tracked and could still sign the scorecard.

If I got it right then LGTM!

Thanks Laura and Sébastien!

Laura-O:
Yep, your summary LGTM!

PR #538, opened 2018-02-05
Explicitly forbid causing intentional puzzle defects.

lgarron:
Based on an email by Callum Goodyear to the WRC:

https://groups.google.com/d/msg/wca-wrc/Plyqnzwi90M/dkjdxnFyBAAJ

### DIFF
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -94,6 +94,7 @@ To be more informative, each Guideline is classified using one of the following 
 
 ## <article-5><puzzle-defects><puzzledefects> Article 5: Puzzle Defects
 
+- 5b+) [ADDITION] The competitor is not permitted to cause a puzzle defect intentionally, unless correcting a previous puzzle defect (see [Regulation 5b3b](regulations:regulation:5b3b) and [Regulation 5b3c](regulations:regulation:5b3c)). Examples of intentionally causing puzzle defects: removing an attached center cap, rotating a corner piece, peeling a sticker.
 - 5b5+) [EXAMPLE] Examples of parts physically detached, resulting in a solved state: one 3x3x3 center cap, one big cube center piece, an inner piece of a big cube.
 - 5b5++) [EXAMPLE] Examples of parts physically detached, resulting in a "DNF": two 3x3x3 center caps, two big cube center pieces, a 3x3x3 edge piece, a 4x4x4 edge piece ("wing"), any combination of two pieces that both have colored parts.
 - 5b5+++) [EXAMPLE] Example of parts physically attached but not fully placed, resulting in a solved state: a 5x5x5 center piece twisted in its spot.


abunickabhi:
Agree on this change, although analysing the intention is a tough task, if the competitor's cube is on the verge of being defective.

xsrvmy:
Does "correcting a previous defect" include fixing a lockup? (For example: fixing a jammed 7x7 center by taking it out of the puzzle)

lgarron:
> Does "correcting a previous defect" include fixing a lockup? (For example: fixing a jammed 7x7 center by taking it out of the puzzle)

That is covered by 5a, 5b1, and 5b3.

lgarron:
@AlbertoPdRF, could you review? :-D

wca-guidelines.md
```diff
@@ -92,6 +92,7 @@ To be more informative, each Guideline is classified using one of the following
 
 ## <article-5><puzzle-defects><puzzledefects> Article 5: Puzzle Defects
 
+- 5b+) [ADDITION] The competitor is not permitted to cause a puzzle defect intentionally, unless correcting a previous puzzle defect (see [Regulation 5b3b](regulations:regulation:5b3b) and [Regulation 5b3c](regulations:regulation:5b3c)). Examples: removing an attached center cap, rotating a corner piece, peeling a sticker.
```
AlbertoPdRF: To be consistent with other Guidelines, I think we should change "Examples:" to something like: "Examples of puzzle defects that can be caused intentionally: removing ...". What do you think?

lgarron: Done! Could you review again?

AlbertoPdRF: LGTM now :)

lgarron: Thanks! 😄

Claster:
Is peeling a sticker also prohibited in Fewest Moves?

lgarron:
> Is peeling a sticker also prohibited in Fewest Moves?

That is a great question. You can already use stickers in FMC to violate quite a few Regulations, though.

xsrvmy:
I think FMC should be excluded. The judge cannot walk around to check for illegal corner twists :P

lgarron:
> I think FMC should be excluded. The judge cannot walk around to check for illegal corner twists :P

There are many things the judge can't practically walk around to check for. 🤷‍♀️

xsrvmy:
I think the better question is, is there any reason to disqualify a competitor from causing a puzzle defect in FMC? I do not see a reason why, say, twisting a corner should have any effect on fmc results, other than being able to solve the puzzle maybe very slightly faster when scrambled.
Also has regulation 5 ever been used in FMC ever?

alexmaasswca:
Really disagree with this applying to FMC, because it has been allowed in the past, so you're causing a time disadvantage for the competitor for no good reason. 

Let's say this becomes a regulation, and Worlds 2019 happens. If you want to enforce this, you have to watch ALL 180 competitors during the WHOLE hour long attempt, for a competitor to twist a corner, which takes less than a second to do. Do you expect that to be feasible to do during the whole entire attempt? And also the two other ones that happen.

In different events you have a judge watching you pretty much the whole solve, which is not the case with FMC, so this is really unenforceable. Why add a reg that that is really unenforceable?

PR #544, opened 2018-03-08
Remove "combined rounds" in favor of an optional cutoff format. Closes #543

### LINKED ISSUE #543: Get rid of the term "combined round"
I've heard many competitors and organizers confused by the terms "combined round" and "combined final".  One common misunderstanding (among both newer competitors and people who have been around a while) is that "combined final" means an event with only one round, so everything is combined into the final.

I don't really know what the benefit of the term is.  A combined final is just a final that has a cutoff.  But I don't know why our software and regulations need to treat the two separately.  It's redundant to say that a round is a combined round AND list a cutoff -- saying that the round is combined doesn't add any information, and clutters the schedule and list of events.

I propose, as an alternative, to just say that rounds may have cutoffs, and that people must meet these cutoffs after 2 solves (for an average of 5) or 1 solve (for a mean of 3) in order to finish their average.

This would not be a functional change, just a terminology change.

lgarron:


wca-regulations.md
```diff
@@ -252,8 +252,8 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
...
     - 9f15) Competitors who achieve the same result in a round receive an identical ranking for the round.
-- 9g) A Combined Round consists of two phases of attempts, where competitors advance to the second phase if they meet a designated cutoff during the first phase.
-    - 9g2) Whether a competitor proceeds to next phase of a Combined round must be decided by result (all competitors with a better result than x) of the first phase (see [Regulation 9f12](regulations:regulation:9f12)).
+- 9g) If the round format is "Average of 5", a round may have a "Best of 2" cutoff phase where the competitor must meet a given cutoff for either of the first two solves (e.g. a result better than X seconds) in order to be eligible for the remaining three attempts.
+    - 9g1) Any cutoff must be announced before the round starts, and should not be changed after it has begun. Changes must be made at the discretion of the WCA Delegate, who must carefully consider the fairness of the change.
```
lgarron, line 256: @thewca/wrc-team I believe that cutoffs are sometimes changed in practice, so I'm copying the regulation for time limits. Let me know if you think we need to lock down cutoffs (but not time limits) more than trusting the judge's discretion.

lgarron:
@timreyn, @jly: How does this look to you?

lgarron:
My interpretation of [9b)](https://www.worldcubeassociation.org/regulations/#9b) in the current Regulations is that they imply that only avg5 rounds can have a cutoff, but I may be wrong.

@jfly, @SAuroux have there ever been any other kind of cutoff rounds?

wca-guidelines.md
```diff
@@ -156,7 +158,7 @@ To be more informative, each Guideline is classified using one of the following
 - A1a2+++) [CLARIFICATION] In case of a cumulative time limit, the result of each remaining attempt in the round is recorded as DNS (Did Not Start) after a competitor reaches the time limit.
 - A1a2++++) [ADDITION] If a cumulative time limit for a single event or a combination of events is enforced, such events don't have a regular time limit per solve.
 - A1a2+++++) [CLARIFICATION] Events where a time limit per attempt is defined by the Regulations (3x3x3 Fewest Moves and 3x3x3 Multi-Blind) must not have a cumulative time limit.
-- A1a3+) [REMINDER] The organization team and the WCA Delegate must be mindful that time limits influence the strategies of the competitors (e.g. rushing the first two attempts in hopes of meeting a cutoff in a combined round), and that changing time limits after the start of the round can disadvantage some competitors unfairly.
+- A1a3+) [REMINDER] The organization team and the WCA Delegate must be mindful that time limits influence the strategies of the competitors (e.g. rushing the first two attempts in hopes of meeting a cutoff), and that changing time limits after the start of the round can disadvantage some competitors unfairly.
```
lgarron, line 161: I didn't make a copy of this for cutoffs, but I wouldn't be opposed to it either.
(If we added something like this for cutoffs, I think we should make it more concise and point at A1a3+ so that the spirit of the guideline is encapsulated in one place.)

julesDesjardin:
9b is indeed quite weird, it doesn't mention bo1/mo3 (often seen in big cubes or old feet), and bo2/bo3 (often seen in 3BLD).
The explicit mention of bo2/avg5 appeared in the January 1 2018 update, precise indications about format of a combined round didn't exist before.
2 examples on top of my mind : 
[bo1/mo3, done after the new regulations](https://www.worldcubeassociation.org/competitions/FrankfurtCubeDays2018/results/all#e666_c)
[bo2/bo3](https://www.worldcubeassociation.org/competitions/BlindOpen2011/results/all#e333bf)
I took the first bo2/bo3 I could think of, it's really old because I think all competitions I've recently been to have switched to a cumulative time limit for 3BLD instead of a cutoff.

I think we should add those 2 other format (bo1/mo3 to 9b2, and bo2/mo3 to 9b3), if we're sure these 3 formats are the only ones that happened.

wca-guidelines.md
```diff
@@ -128,7 +128,9 @@ To be more informative, each Guideline is classified using one of the following
...
 - 9f4+) [EXPLANATION] The result of an attempt is DNF only if the competitor began the attempt by indicating that they are ready (see [Regulation A3b2](regulations:regulation:A3b2)) but the attempt was disqualified.
-- 9f5+) [CLARIFICATION] The result for an attempt is DNS if the competitor was eligible for the attempt and did not start it (see [Regulation A3b2](regulations:regulation:A3b2)). If the competitor did not qualify for an attempt (e.g. in a combined round), they have no result for the attempt.
+- 9f5+) [CLARIFICATION] The result for an attempt is DNS if the competitor was eligible for the attempt and did not start it (see [Regulation A3b2](regulations:regulation:A3b2)). If the competitor did not qualify for an attempt (e.g. see [Guideline 9g+](guidelines:guideline:9g+)), they have no result for the attempt.
+- 9g+) [CLARIFICATION] A competitor who does not meet the cutoff result during the cutoff phase solves has no results for the remaining solves (rather than a result of DNS for each of the remaining solves).
```
jfly, line 132: I think it's strange to use the phrase "cutoff result" here, where in [regulation 9g](https://www.worldcubeassociation.org/regulations/#9g) we say "designated cutoff".

jfly: nit: I'm not sure the word "solves" has any value in the phrase "cutoff phase solves". I'd consider replacing that with simply "cutoff phase".

jfly (commented):
This all looks good to me. Some of these changes feel like cleanup/clarification to me and not specifically related to removing the phrase "combined round".

SAuroux:
@lgarron I disagree with your interpretation of 9b). What it does in my opinion is to only fix the format of Combined rounds for 9b1) events to Bo2/Ao5 (i.e. preventing Bo1/Ao5 and Bo3/Ao5). There is simply no such limitation for 9b2) and 9b3) events (even though I would personally prefer this).

wca-regulations.md
```diff
@@ -252,8 +252,8 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 9f13) For "Mean of 3" and "Average of 5" rounds, rankings are assessed based on the ordering of the averages/means of the competitors, where "better" is the shorter recorded time.
     - 9f14) For "Mean of 3" and "Average of 5" rounds, if two or more competitors achieve identical average/mean results, rankings are assessed based on the best attempt per competitor, where "better" is defined as the shorter time.
     - 9f15) Competitors who achieve the same result in a round receive an identical ranking for the round.
-- 9g) A Combined Round consists of two phases of attempts, where competitors advance to the second phase if they meet a designated cutoff during the first phase.
-    - 9g2) Whether a competitor proceeds to next phase of a Combined round must be decided by result (all competitors with a better result than x) of the first phase (see [Regulation 9f12](regulations:regulation:9f12)).
+- 9g) If the round format is "Average of 5", a round may have a "Best of 2" cutoff phase where the competitor must meet a given cutoff with either of the first two solves (e.g. a result better than X seconds) in order to be eligible for the remaining three attempts.
```
timreyn, line 255: Hmm, this wording is a little awkward to me.  The term "phase" seems unnecessary.

How about...
9g) A round may have a designated cutoff, where competitors must meet the cutoff time with one of their first solves in order to be eligible for for the remaining attempts.
9g1) If the round format is "Average of 5", the competitor must meet the cutoff time with one of their first two solves.
9g2) For other round formats, the competitor must meet the cutoff time either with their first solve or with one of their first two solves.  This must be announced by the organization team in advance.

I don't like the word "this" in 9g2, but thoughts?

jfly: We recently did a pass over the regulations to make sure there aren't any oversights with regards to multibld and fmc. Using the phrase "cutoff time" would break this for multibld (something that I believe was used at worlds in France recently).

Otherwise, I definitely like getting rid of the concept of "phases". I know many people like to think of cutoffs in this way, but I personally think it's a confusing concept, and is not the way I explain the idea to newcomers.

timreyn: Good point.  How about "cutoff result" instead?

jfly: That sounds better. I'm not sure if we define what it means to "meet" a result anywhere, though. I don't have time to check the regulations, but I think we do define what a "better" result is, so it might be better to use the word "better" here rather than "meet".

xsrvmy:
https://euro2016.cubing.net/schedule

This is an example of bo1/bo2 for multibld. Note that the first attempt has a cutoff time.

wca-regulations.md
AlbertoPdRF: Yes please! "Better" instead of "meet", "result" instead of "time".

wca-guidelines.md
AlbertoPdRF: Hmm, it's weird to me that A1a3+) starts speaking of "time limits" and the example that lists talks about "cutoffs" :P I never thought about that before, but now that I did, I don't like the wording of this Guideline.

lgarron:
It looks like this should be changed to explicitly allow more combined round formats, but I'll hold off on revising until we can come to consensus in #543 whether we should really get rid of the phrase "combined round" and/or if we should replace it with another phrase.

xsrvmy:
I think that at local side event comp (eg. 67 mega), there is a chance that the cutoff from the first round will apply to the second round if unfinished averages manage to advance. Having both a 'combined first' and a 'combined finals' is weird.

lgarron:
Closing this for now.

PR #565, opened 2018-05-01
Change "time" to "result" in 9f13)

AlbertoPdRF:
Thanks to Ron for pointing this out!


wca-regulations.md
```diff
@@ -249,7 +249,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - 9f12a) For timed results, "better" is defined as the shorter time.
         - 9f12b) For Fewest Moves Solving, "better" is defined as the shorter solution length.
         - 9f12c) For Multiple Blindfolded Solving, rankings are assessed based on number of puzzles solved minus the number of puzzles not solved, where a greater difference is better. If the difference is less than 0, or if only 1 puzzle is solved, the attempt is considered unsolved (DNF). If competitors achieve the same result, rankings are assessed based on total time, where the shorter recorded time is better. If competitors achieve the same result and the same time, rankings are assessed based on the number of puzzles the competitors failed to solve, where fewer unsolved puzzles is better.
-    - 9f13) For "Mean of 3" and "Average of 5" rounds, rankings are assessed based on the ordering of the averages/means of the competitors, where "better" is the shorter recorded time.
+    - 9f13) For "Mean of 3" and "Average of 5" rounds, rankings are assessed based on the ordering of the averages/means of the competitors, where "better" is the shorter recorded result.
```
jfly: To make this concrete, the thing we're trying to deal with here is 333fm, right? If so, then the problem with this fix is that "shorter" isn't the right word for a number of moves. "Smaller" might be an improvement, although I think the best fix would be to explicitly state how to handle results in seconds vs results in number of moves.

wca-regulations.md
```diff
@@ -249,7 +249,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
...
     - 9f14) For "Mean of 3" and "Average of 5" rounds, if two or more competitors achieve identical average/mean results, rankings are assessed based on the best attempt per competitor, where "better" is defined as the shorter time.
```
jfly: This also uses the word "time", should it be changed as well?

jfly (commented):
Some nitpicks for you. I'm curious what other people think. Thanks for getting this discussion started!

wca-regulations.md
AlbertoPdRF: We also use "shorter" in 9f12b), and maybe in other places. Should we change them all? I don't mean I'm against, I just don't know.

wca-regulations.md
AlbertoPdRF: I think yes. I'll wait to do it until I have time to review if there are more occurrences of this, and also until we agreed on if we keep "shorter" or change it to any other term.

wca-regulations.md
Laura-O: 9f12b is "a shorter solution length", which is the correct word.
I agree it would be better to split this into seconds and moves, but I think "smaller" is fine as well.

viroulep:
Thanks for adjusting the base branch Alberto!

PR #563, opened 2018-04-08
Improve wording and grammar

micahstairs:
A collection of minor improvements to the wording of the regulation and guidelines.

PR #568, opened 2018-05-29
Improve wording and grammar

viroulep:
@micahstairs changes from #563 submitted against `official` instead of `draft`.

I've just created [this wiki entry](https://github.com/thewca/wca-regulations/wiki/On-submitting-pull-requests) for why we need such changes to land first in `official`, as it's not really obvious.


jfly:
"need such changes to land first in official"? Are you sure you don't mean "draft" here?

viroulep:
I did mean `official`: this is a textual revision that we want both in official and draft.
If it lands first in official, then we can merge official in draft (which is safe, as anything official should be in draft too).
If it lands first in draft, then we would have to merge draft in official (which is unsafe, as draft may contain major changes that need to wait for the next Regulations release).

The point here is that textual revisions should appear asap in the official version of the Regulations, without a need for a release (we just add a "latest textual revision on X"). Whereas "regular" changes do need a release.

jfly:
Ah, I see. This is like a hot fix :) I misinterpreted your first sentence! I thought you were saying that Micah's PR was incorrectly submitted against official when it should have been submitted against draft.

Sorry for the noise.

Issue #546, opened 2018-03-09
Reference in Fewest Moves Solving (Article E)


DanielEgdal:
Currently in [E2c](https://www.worldcubeassociation.org/regulations/#E2c) it refers back to [Regulation 12a](https://www.worldcubeassociation.org/regulations/#12a) when describing what OBTM is. This creates an issue, since the rotations aren't described in 12a. 
As I got in response to my [previous isssue](https://github.com/thewca/wca-regulations/issues/542), I was told to create this, and that renumbering wasn't possible to do.

lgarron:
Done in 8104b4b795b2e9e993cdd220b071354c4ffb0591.

Issue #456, opened 2017-07-14
Clarification of cube rotations used in FMC solutions


Laura-O:
The current regulations allow the following notation for cube rotations (listing only 90 degree clockwise rotations here, same can be applied to 90 degree anti-clockwise and 180 degree):
- small letters in square brackets: [f], [b], [r], [l], [u], [d]
- x, y, z

Competitors like mixing these notations. We have seen:
- small letters in parentheses: (f), (b), (r), (l), (u), (d)
-  x-/y-/z-notation in square brackets: [x], [y], [z]
-  x-/y-/z-notation in parentheses: (x), (y), (z)

As it has lead to countless and endless discussions whether solutions using this wrong notation should be DNF or not, we need a clear decision here. This should also be clarified in the regulations or guidelines.
Maybe reworking the FMC sheet would be a good idea as well.

lgarron:
> As it has lead to countless and endless discussions whether solutions using this wrong notation should be DNF or not, we need a clear decision here. This should also be clarified in the regulations or guidelines.

I don't know why this is a subject of debate. The letter and spirit of E2c++ is clear:

> E2c++) CLARIFICATION The competitor's solution must only use moves that are exactly defined in Regulation 12a. Examples of notation and moves that are not permitted in competition: [F], [R, U], [R: U], R'2, L'w, f, M, U2'.

Laura-O:
It might be clear for you. It is not clear for a large number of delegates and competitors, who approve or use the wrong notation described in the first post.

lgarron:
Yeah, reworking the FMC sheet is probably a good idea to address that. We should probably enumerate all moves and add a line saying "anything moves not explicitly listed here will result in a DNF".

pedrosino:
I personally have no problem with [y] or (y) since it's very clear what the intent is.
(I use to draw ( ) around rotations when grading to help me count the moves...maybe this is not a good idea?)

What would that reworking of the FMC sheet looke like? Would we allow more moves or give more counter-examples?

Claster:
Well, the intention of M, S, E moves is also clear, and most delegates probably understand [M, U2]. Yet we disallow these.
I guess reworking of FMC sheets means adding counter-examples.

pedrosino:
Yeah, my wording was not so great.

I would be in favour of allowing [ ] or ( ) around x/y/z or lowercase letter to mean rotations.

lgarron:
[...]

PR #575, opened 2018-07-10
Notation: move rotations under 12a and attempt to describe valid FMC solutions as clearly as possible.

lgarron:
The new examples for E2c++ should help with concerns from issue #456.
The change to Article 12 also solves the (arguable) issue described in issue #546.

@Laura-O, how do these changes look to you?

### DIFF
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -152,6 +152,10 @@ To be more informative, each Guideline is classified using one of the following 
 - 11e2++) [EXAMPLE] Suppose a competitor has 5 regularly numbered attempts, and #2 leads to extra attempts. The competitor should do extra attempts until there is a valid extra attempt to replace #2. Suppose later another attempt leads to extra attempts, the competitor should proceed in a similar way as for #2.
 
 
+## <article-12><notation><notation> Article 12: Notation
+- 12a2+) [CLARIFICATION] For example, both Rw and 2Rw are valid notation for the same physical move. 1Rw is not valid notation for any NxNxN Cube, and 3Rw is not valid notation for the 2x2x2 Cube or the 3x3x3 Cube (but 3Rw is valid notation for the 4x4x4 Cube and larger NxNxN Cubes).
+
+
 ## <article-A><speedsolving><speedsolving> Article A: Speed Solving
 
 - A1a2+) [ADDITION] In case of a cumulative time limit, the judge records the original recorded time for a DNF on the score sheet in parentheses, e.g. "DNF (1:02.27)".
@@ -202,12 +206,16 @@ To be more informative, each Guideline is classified using one of the following 
 ## <article-E><fewest-moves><fewestmovessolving> Article E: Fewest Moves Solving
 
 - E2b+) [CLARIFICATION] A competitor may choose to stop an attempt early by handing in a solution before the time limit.
-- E2c+) [CLARIFICATION] Although the judge may provide a standard sheet for competitors to submit solutions, a competitor may submit a solution on a different piece of paper, as long as a single move sequence is unambiguously indicated as the submitted solution. (Note that the piece of paper must come from the judge, according to [Regulation E3a](regulations:regulation:E3a).)
-- E2c++) [CLARIFICATION] The competitor's solution must only use moves that are exactly defined in [Regulation 12a](regulations:regulation:12a). Examples of notation and moves that are not permitted in competition: [F], [R, U], [R: U], R'2, L'w, f, M, U2'.
+- E2c+) [CLARIFICATION] Although the judge may provide a standard sheet for competitors to submit solutions, a competitor may submit a solution on a different piece of paper. (Note that the piece of paper must come from the judge, according to [Regulation E3a](regulations:regulation:E3a).)
+- E2c1+) [ADDITION] If the competitor forgot to write their name on their intended solution paper before the end of 60 minutes, they may be allowed to add their name (without modifying their solution or making any other changes) under direct supervision of the judge, at the discretion of the Delegate. The competitor must ask and write their name before they finish submitting their solution (e.g. placing it in the judge's hands and allowing the judge to move on to the next competitor, placing it on the judge's table and walking away), and are not allowed to add their name later.
+- E2c2+) [EXAMPLE] Examples that are not considered a single unambiguous move sequence written out as one or multiple horizontal lines of text: a solution with extra moves, multiple solution sequences without exactly one boxed/circled solution, a solution with ambiguous symbols (e.g. a letter that looks similar to "B" and "R" but is not clearly one of them), a solution written in a spiral, a solution with stars or arrows indicating moves to be performed out of written order (e.g. "insertions" that are not written in-line with the rest of the move sequence, "pre-moves" that are not incorporated into the move sequence using valid notation), any arrangement of symbols that cannot be interpreted as an unambiguous sequence of "top-to-bottom" text lines with each line in "left-to-right" reading order.
+- E2c2++) [CLARIFICATION] The submitted page may contain written moves on both sides, as long as only one side has a boxed/circled solution.
+- E2c3+) [ADDITION] A blacked-out move or symbol must be covered using ink in a rectangle, oval, or other solidly filled shape. A partially filled shape is permitted as long as it clearly blacks out the move, at the discretion of the Delegate.
+- E2c4+) [CLARIFICATION] Examples of symbols or combinations of symbols that are not permitted in solutions for Fewest Moves Solving: M, e, r, Y (uppercase), T, U1, Ui, U3, U-, R++, R2', R'2, L'w, 1R, 2U, 2f, 2-3r, 1Rw, 3Rw, 2-3Rw, rw, Rr, *, →, ⮐, /, ., ?, [r u], [F], <f>, {f}, [f]', [f]2, [x], (x), (L), 2(Rw), (U)2, (U)', (R U)', (R U R'), [R, U], [R: U].
 - E2e+) [CLARIFICATION] Competitors must not derive solutions from any part of the scramble sequence, and solutions should not share significant parts with the inverse scramble sequence.
 - E2e++) [EXAMPLE] Example of solutions that should result in a disqualification of the attempt (DNF): solutions beginning with the same 4 or more moves as the inverse scramble sequence.
-- E3d+) [CLARIFICATION] Any stopwatch or watch used by a competitor must not appear to have any functions that would help the competitor find a solution.
-- E3d++) [CLARIFICATION] Competitors should not consider a personal stopwatch or watch as the official time, and must submit their solution after the judge calls "STOP".
+- E3d+) [CLARIFICATION] Any stopwatch or watch used by a competitor must not have any functions that would help the competitor find a solution.
+- E3d++) [CLARIFICATION] Competitors should not consider a personal stopwatch or watch as the official time, and must submit their solution when the judge calls "STOP".
 
 
 ## <article-H><multiple-blindfolded><multipleblindfoldedsolving> Article H: Multiple Blindfolded Solving
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -166,7 +166,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - 4b3f) Pyraminx: The (random) state must require at least 6 moves to solve.
     - 4b4) Each scramble sequence should be applied during a maximum time frame of 2 hours. This time frame starts when the scramble sequence is applied for the first time.
 - 4d) Scrambling orientation:
-    - 4d1) NxNxN puzzles and Megaminx are scrambled starting with the white face (if not possible, then the lightest face) on the top and the green face (if not possible, then the darkest adjacent face) on the front.
+    - 4d1) NxNxN Cubes and Megaminx are scrambled starting with the white face (if not possible, then the lightest face) on the top and the green face (if not possible, then the darkest adjacent face) on the front.
     - 4d2) Pyraminx is scrambled starting with the yellow face (if not possible, then the lightest face) on the bottom and the green face (if not possible, then the darkest adjacent face) on the front.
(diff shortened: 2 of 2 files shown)


wca-regulations.md
```diff
@@ -493,11 +493,11 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
...
     - E2b) All competitors have a total time limit of 60 minutes to devise a solution.
         - E2b1) The judge should call "5 MINUTES REMAINING" at 55 minutes, and must call "STOP" at 60 minutes.
-    - E2c) At 60 minutes, each competitor must give the judge a legibly written solution with the competitor's name, using the notation defined for Outer Block Turn Metric (described in [Regulation 12a](regulations:regulation:12a)). Penalty: disqualification of the attempt (DNF).
-    - E2d) The length of the solution is calculated in Outer Block Turn Metric (see [Regulation 12a](regulations:regulation:12a)).
-        - E2d1) The competitor is permitted a maximum solution length of 80 (moves and rotations).
+    - E2c) At 60 minutes, each competitor must give the judge a legibly written solution with the competitor's name using the notation for NxNxN puzzles (described in [Regulation 12a](regulations:regulation:12a)), such that starting with a solved puzzle and applying the scramble sequence followed by the solution results in a solved puzzle. Penalty: disqualification of the attempt (DNF).
```
lgarron: An unsubmitted or invalid solution is clearly a DNF. Have we historically allowed leniency for missing names if the competitor can be clearly identified after the fact? (I don't recall recent incidents.)

lgarron: I've added a draft solution in E2c1+) at https://github.com/thewca/wca-regulations/pull/575/files#diff-8356179f2de288bb7a0aa8a610b9a487R208 that balances practical leniency (you can write your name in front of the judge while submitting your solution) with a clearly enforceable line (you can't add it any later).

wca-regulations.md
```diff
@@ -491,13 +492,18 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
...
+    - E2d) The competitor's result is the number of moves in their solution, calculated using Outer Block Turn Metric (see [Regulation 12a5](regulations:regulation:12a5)).
+        - E2d1) The competitor's solution must not exceed 80 moves (including rotations) when calculated using Execution Turn Metric (see [Regulation 12a6](regulations:regulation:12a6)). Penalty: disqualification of the attempt (DNF).
     - E2e) The competitor's solution must not be directly derived from any part of the scramble sequence. Penalty: disqualification of the attempt (DNF), at the discretion of the WCA Delegate.
-        - E2e1) The WCA Delegate may ask the competitor to explain the purpose of each move in their solution, irrespective of scrambling algorithm. If the competitor cannot give a valid explanation, the attempt is disqualified (DNF).
+        - E2e1) The WCA Delegate may ask the competitor to explain the purpose of each move in their solution, irrespective of the scrambled sequence. If the competitor cannot give a valid explanation, the attempt is disqualified (DNF).
```
jfly: "scrambled sequence" -> "scramble sequence"?

jfly (commented):
This all looks like an improvement to me!

lgarron:
Another approach is to require competitors to submit their solutions on FMC sheets, which encourage unambiguous formatting. Competitions usually print FMC sheets, and competitors are used to them, so this is fairly reasonable.

However, requiring FMC sheets means one of two possibilities:

### Every competitor is allowed only one FMC sheet
Tis requires them to juggle backup solutions and will probably lead to competitors transcribing their solutions under pressure in the last few minutes. It might lead to more incidents of competitors scribbling just a few seconds past the time limit.

### Competitors are allowed to ask for extra sheets (possibly up to a limit)
This requires organizers to print out more sheets. 

If a competitor asks for a sheet they're entitled to – but there aren't enough – then we presumably need to fall back to allowing solutions on any paper.  Even though this might lead to fewer incidents, we would still need clear Regulations for how solutions on spare paper are formatted.

However, it might be good enough to keep plenty of blank solution sheets (which TNoodle already generates) on hand. Those can be kept by organizers across competitions, so it's more reasonable to ask organizers to have a lot of them at any given competition.

wca-regulations.md
lgarron: Fixed, thanks!

Baiqiang:
The clarification E2c4+ makes me recall the 19 moves of Vladislav Ushakov. I've argued that the solution contained xyz in () but some thought the () could be ignored.

lgarron:
> The clarification E2c4+ makes me recall the 19 moves of Vladislav Ushakov. I've argued that the solution contained xyz in () but some thought the () could be ignored.

Indeed, one purpose of this PR is to clarify that parentheses are not valid notation.

(The only reason E2c4+ doesn't cover them is because it seemed so obvious that it wasn't worth mentioning. Incident discussions have proven otherwise.)

Laura-O:
There are two main effects requiring solutions to be on official sheets would have:
1. Competitors will have to rewrite their solutions more often (for example because they forgot to write down moves and are not allowed to add them as extra moves).
2. Competitors have less space to write down their solution (only the main, official sheet).

I think this change in combination with a stricter handling of the notation does change the event significantly.

Apart from this, it also adds a potential unfair treatment of competitors: let's say, 3 minutes before the end of the attempt, several competitors ask for an additional sheet and the judge is not fast enough to give everyone a sheet. One competitor manages to write down the solution in time, others don't because they received the sheet later. I think this generates some randomness which we do not want to deal with.

Additionally, I think the descirption of allowed notation is ambiguous. These are two solutions which I found when looking through about 10 of the solution sheets from a competition I delegates and it's already not clear to me, how they would be judged with this PR:

img_20180725_210904

There is no star or arrow, but it's not written in-line and it's an extra move. If that would be allowed, what if there was a small line going from the crossed out move to the replacement move?

What about this one?
img_20180725_212737

Is this a DNF because the competitor drew an arrow and it's an extra move? Or is it OK because it's basically a move to be done in the written order, and one could say it's in-line?

Edit: So to summarize my thoughts on this: I think it will generate a ton of decisions which are handled differently by different delegates, which is making the situation worse, not better.

lgarron:
Well, I designed this PR specifically so that examples like yours are clear DNFs.

First image: T/F is an ambiguous symbol, R is not written in line, and the struck-through move is not clearly blacked out (although the latter is at the discretion of the Delegate).

Second image: Even if it's not an arrow, the bracey thing is not a valid symbol.

Do you think these cases are actually ambiguous under this revision?

pedrosino:
I think both examples Laura gave should remain valid. I also think requiring that the solution is delivered in the "official" sheet is probably overcomplicating, and maybe a kickback.

As for the struck-through move, it's clearly blacked out to me, and I would only consider the one on the top.
Requiring that competitors re-write the whole thing if they find a move missing after it's all done is not fair in my opinion.

lgarron:
> Requiring that competitors re-write the whole thing if they find a move missing after it's all done is not fair in my opinion.

It may not be fun, but it's *fair* if the rules are the same for everyone, and clear ahead of time.

Also note that this revision doesn't prevent writing extra moves between the move slots on the FMC sheet, as long as they are written in line without extra symbols. (I don't think that's an important thing to prevent, so long as the Regulations themselves don't specify how to fill out the official FMC sheets.)

lgarron:
Anyhow, since the two opinions here so far are against requiring an official sheet: what about allowing a separate page as long as it is blank apart from name/WCA ID/round info and the solution?

And if you think that's not a solution either, do you think what I proposed here about boxing/circling solutions is good?

Laura-O:
> First image: T/F is an ambiguous symbol, R is not written in line, and the struck-through move is not clearly blacked out (although the latter is at the discretion of the Delegate).

Where is a T? :confused: 

> Do you think these cases are actually ambiguous under this revision?

Yes, 100% ambiguous.

I also don't see the reason why we should forbid such a notation. I am very certain that I could ask 100 people on the street and 99 of them would interpret the letters correctly and the 100th is a humanoid robot which didn't learn to interpret crossed out letters. :stuck_out_tongue: 

Joking aside, we have always tried to keep regulations as simple and reasonable as possible. For me, DNFing the above examples would be difficult to do consistently (by all delegates) and is unreasonable. I have checked hundreds (if not thousands) of FMC solutions and I can't remember many cases where the addition of extra moves was ambiguous. So why should this be forbidden?

pedrosino:
I'm aligned with Laura. I also personally don't see the problem with ( ) since they have no meaning in our notation (different from [ ] which do have a meaning).

Boxing/circling may not be necessary and/or practical, if you did a lot of crossing out moves for example.
I'm also against requiring a sheet with only name and solution. As I said, it would be a kickback from what we have now, and you're right that requiring a re-write would be fair as everyone would be under the same conditions, but I don't think is necessary or helps the judge/checker.

Most problems I face with FM are wrong notation and bad hand writing.

xsrvmy:
With the issue with parens, we need to make clear that notation such as F (U R U' R')2 F' for repeated moves is forbidden. Commutators are listed as invalid in the guidelines so this probably should be disallowed explicitly as well.

lgarron:
> With the issue with parens, we need to make clear that notation such as F (U R U' R')2 F' for repeated moves is forbidden. Commutators are listed as invalid in the guidelines so this probably should be disallowed explicitly as well.

This current revision of this PR already includes the following examples of invalid notation:

    (U)2, (U)', (R U)'

I hope this makes it clear that (U R U' R')2 is not valid, although I suppose we could as add it to the morass if anyone thinks someone could make the opposite conclusion.

xsrvmy:
@pedrosino How bad is the handwriting issue? There are only 12 valid characters (URFDLBwxyz'2) outside brackets and 6 valid characters (urfdlb) inside brackets and the only ones that look similar are z and 2. Even then z and 2 cannot be interchanged to form valid notation.

Laura-O:
> @pedrosino How bad is the handwriting issue? There are only 12 valid characters (URFDLBwxyz'2) outside brackets and 6 valid characters (urfdlb) inside brackets and the only ones that look similar are z and 2. Even then z and 2 cannot be interchanged to form valid notation.

Sorry, but :woman_facepalming: 

Bad handwriting is by far the most annoying problem in FMC. You don't read WCA reports and might have never judged FMC solutions, but it should be obvious that there are way more letters that are similar when written in a sloppy way.

A few examples:
U written with a small bend on the right side -> D
B written with the middle horizontal line far away from the vertical line -> D
B not fully closed on the bottom -> R
And so on...

pedrosino:
There's also the F that can look like a T, or R too closed that looks like a B, L too curved and looks like a U.

xsrvmy:
I was mainly thinking about the F/T issue. T is not valid notation to begin with, so should the checker always assume that a poorly written F is an F?

Another point I want to make is that sometimes the accepted way of writing a letter can be different in different countries. For example, some Chinese judges might not be able to read a hand-written z, small f or small r.

wca-guidelines.md
```diff
@@ -201,12 +205,16 @@ To be more informative, each Guideline is classified using one of the following
...
+- E2c2+) [EXAMPLE] Examples that are not considered a single unambiguous move sequence written out as one or multiple horizontal lines of text: a solution with extra moves, multiple solution sequences without exactly one boxed/circled solution, a solution with ambiguous symbols (e.g. a letter that looks similar to "B" and "R" but is not clearly one of them), a solution written in a spiral, a solution with stars or arrows indicating moves to be performed out of written order (e.g. "insertions" that are not written in-line with the rest of the move sequence, "pre-moves" that are not incorporated into the move sequence using valid notation), any arrangement of symbols that cannot be interpreted as an unambiguous sequence of "top-to-bottom" text lines with each line in "left-to-right" reading order.
+- E2c2++) [CLARIFICATION] The submitted page may contain written moves on both sides, as long only one side has a boxed/circled solution.
+- E2c3+) [ADDITION] A blacked-out move must be covered using ink in a rectangle, oval, or other solidly filled shape. A partially filled shape is permitted as long as it clearly blacks out the move, at the discretion of the Delegate.
```
xsrvmy: E2c3+) "A blacked-out MOVE must be ..."
E2c3 also includes any other symbol. Should this read "A blacked-out move or symbol ..."

wca-guidelines.md
```diff
@@ -201,12 +205,16 @@ To be more informative, each Guideline is classified using one of the following
...
+- E2c2+) [EXAMPLE] Examples that are not considered a single unambiguous move sequence written out as one or multiple horizontal lines of text: a solution with extra moves, multiple solution sequences without exactly one boxed/circled solution, a solution with ambiguous symbols (e.g. a letter that looks similar to "B" and "R" but is not clearly one of them), a solution written in a spiral, a solution with stars or arrows indicating moves to be performed out of written order (e.g. "insertions" that are not written in-line with the rest of the move sequence, "pre-moves" that are not incorporated into the move sequence using valid notation), any arrangement of symbols that cannot be interpreted as an unambiguous sequence of "top-to-bottom" text lines with each line in "left-to-right" reading order.
+- E2c2++) [CLARIFICATION] The submitted page may contain written moves on both sides, as long only one side has a boxed/circled solution.
```
xsrvmy: E2c2++) Should this read "as long AS only one side..."?

xsrvmy:
Also, a dot is listed as a disallowed symbol. Does that include stray pen marks that cannot be confused with a quote mark (eg. putting a dot on the page by habit or by accident)?

wca-guidelines.md
lgarron: Yep, thanks!

wca-guidelines.md
lgarron: Yeah, good idea. Done.

lgarron:
> Also, a dot is listed as a disallowed symbol. Does that include stray pen marks that cannot be confused with a quote mark (eg. putting a dot on the page by habit or by accident)?

Yep. If you have a better idea that can be judged consistently, I'm all ears, though.

xsrvmy:
@lgarron Is it appropriate to ignore any inline symbols that are clearly invalid? Alternatively, can a dot be considered a solidly filled shape?

lgarron:
Closed in favor of #678.

Issue #573, opened 2018-07-04
Ad4 and A3c3 - resetting the timer during inspection, "timer tests"


dancing-jules:
Hello, I just got into a discussion on Facebook whether a "timer test" would be legal during inspection (starting, stopping and resetting the timer).

The other person argued that it would be legal per A3c3:

> A3c3) The competitor may reset the timer before the start of the solve.

Where as I argue, that it would result in a DNF because of A4d:

> A4d) If a Stackmat timer is in use, the competitor starts the solve by confirming that the timer light is green and then removing their hands from the timer (thus starting the timer). 

The way I see it, this regulation makes clear that any time you start the timer during inspection, you start the solve. 

The way I see A3c3: as long as you don't start the timer, you can push the power and reset buttons as often as you want to. 

Further I argue that if "timer tests" during inspections were legal, it would make it impossible for judges to discern whether a competitor starting the timer is doing a timer test or actually starting the solve.

lgarron:
I'd be disappointed if we have to clarify this.

A3c3) allows resetting the timer, and would explicitly include more details else if starting the timer were allowed.

> Further I argue that if "timer tests" during inspections were legal, it would make it impossible for judges to discern whether a competitor starting the timer is doing a timer test or actually starting the solve.

oya.
Even if the competitor communicates this to the judge, there is a possibility for misunderstanding by the judge as well as people watching/video evidence.

We do have a roughly analogous situation in that you're technically allowed to take off the blindfold in BLD before stopping the timer as long as you don't do any moves. It's arguably safe for the competitor to do what they want with the timer as long as it starts running within 15 seconds of the start of inspection. But the competitor controls the timer in BLD (at least for Stackmat solves), and for normal inspection the judge needs to be able to tell when it's over (so that they can stop using their inspection stopwatch). I don't see enough benefit from starting to allow timer tests after the attempt has started – only more potential for incidents.

I'd be alright with adding A3c3++ for this, if there is a chance competitors would actually want to argue that they're allowed to do a timer test.

xsrvmy:
To be clear, timer tests before inspection is legal, right?

dancing-jules:
@lgarron - Thanks for your reply. I invited the person who argued with me to contribute here. They said they think the regulations should be a bit more specific.

[...]

PR #579, opened 2018-07-17
Clarify that starting the solve ends inspection.

lgarron:
Two changes:
- Reword A4d) to be clear that the solve starts when the competitor starts the timer, regardless of whether they looked for a green light.
- Clarify that the competitor is not permitted to reset (and therefore is not permitted to "test" or "restart") the timer after starting it, even if they attempt to do so before they reach the what would have been the limit for inspection. Addresses #573.


AlbertoPdRF (approved):
LGTM!

wca-regulations.md
```diff
@@ -395,7 +395,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 - A4) Starting the solve:
     - A4b) The competitor uses their fingers to touch the elevated sensor surfaces of the timer. The competitor's palms must be facing down, and located on the side of the timer that is closer to the competitor. Penalty: time penalty (+2 seconds).
         - A4b1) The competitor must have no physical contact with the puzzle between the inspection period and the start of the solve. Penalty: time penalty (+2 seconds).
-    - A4d) If a Stackmat timer is in use, the competitor starts the solve by confirming that the timer light is green and then removing their hands from the timer (thus starting the timer).
+    - A4d) If a Stackmat timer is in use, the competitor should keep their hands on the timer until they see that a green timer light. They start the solve by removing their hands from the timer (thus starting the timer).
```
pedrosino, line 398: This doesn't sound like the usual "should" we have.
Also, maybe because I'm not a native speaker, but the bit "...they see that a green timer light" sounds a bit weird.
Would it be "lights up"? Or "...that the timer shows a green light"?

And finally, "...(thus starting the timer)." maybe can be removed, since it will not happen if they didn't wait for the green light.

xsrvmy: @pedrosino Are you arguing "should" needs to be changed to a "must"? That would imply that the competitor cannot lift his hands off before the green light and continue inspection.

pedrosino: But that's precisely what we're trying to say, no?
You can't put your hands, take them off, and then put them on again and start the timer. Judge should stop the stopwatch once the competitor lifts their hands, no matter if the timer actually started or not.

xsrvmy: I thought this was always allowed in case of malfunctions or if timer is not reset/turned on.

Claster: The judge should stop the stopwatch once the competitor started the timer (that is, green light + lifted hands).
Lifting hands before green light should be allowed, although that just wastes inspection's time.

pedrosino: It's not easy for the judge to see if the light is on or not...
IMO it should not be explicitly permitted in the regulations to do that...it will only cause confusion.

Laura-O: `until they see that a green timer light` seems to be wrong.
If `light` is a noun, there shouldn't be a `that`. If `light` is a verb, it seems to be the wrong tense and we don't have green timers. 😛 
If the sentence can be understood in a different way, it also should be changed as I couldn't find this way after reading it ten times. 😄

xsrvmy: The judge can always look at the display if it's available. Or just wait for a move because of the timer didn't start it's an automatic DNF.

dancing-jules: @xsrvmy Waiting for a move is a very bad idea. What if the competitor starts the timer at 14.5 seconds but makes their first move at 15.5 seconds? They should not get a +2.

lgarron: Thanks for noticing the grammar issue, folks! Please note that I actually already corrected that in the version of the commit that landed: https://github.com/thewca/wca-regulations/commit/fff14fe406fbe42b4f034d6ffe2df35462f3f994

Issue #567, opened 2018-05-28
DNF due to turning off the Alarm during 4BLD


vinitnalhe:
Hi @all,
During 4BLD  round, 3rd solve at Indian Nationals 2018, my cell phone was being used as a timer (since I average way above 10 minutes, time that stackmat timer can record). My phone was kept on Airplane mode as soon as I entered the solving room.

While I was memorising the cube my alarm started ringing withing a second I swiped it off(instinctive reaction so that others arent disturbed during blindfolded round). I completed the solve the times were noted and the Delegate in the room didnt object about it.
Some time later I was given a DNF on that solve stating regulation 2i

> 2i) While competing, competitors must not use electronics or audio equipment (e.g. cell phones, MP3 players, dictaphones, additional lighting).

Firstly I wasn't using any electronic device. My cellphone was being used as a timer (I just tuned the alarm off).
Secondly  what would have happened if the alarm was let to ring and I didnt turn it off, disturbing other participants?

akashrupela:
Thanks Vinit for opening this up. This indeed was a very difficult incident
to handle.
@ All, i was one of the delegates who made the call that this should mostly
be a DNF, similar to Max Park and Alexandre Carlier's solves of interacting
with electronic devices like a camera/phone. What makes this different is
that the phone was also running the timer and not turning off alarm may
disturb other competitors, however this was done by Vinit himself without
asking a judge/delegate to do that with a good intent and instinct. We are
also going to mention this in the competition report.

It is obviously clear that Vinit had no bad intent (just like some other
results that were disqualified by 2i), however due to it being really
similar to how WCA regulations have handled similar cases, we decided that
a DNF was the right course of action. I would definitely like to hear what
others think of this decision and we can revert the decision if necessary.

Regards
Akash

xsrvmy:
We seem to have another problem: If the phone screen dims during memo, can the competitor tap it to prevent it from sleeping? Also, can the competitor stop the timer (rather than the judge)

vinitnalhe:
Also the comment above by @xsrvmy reminds me of one more incident that happened to me during the 1st attempt of MBLD at the same competition.

[...]

PR #580, opened 2018-07-17
2i) Clarify that interaction with the Stackmat timer is allowed, even though it is electronic.

lgarron:
This came up in #567.


sam596:
Should this also include the stopwatch for longer attempts? For instance, I assume I'm allowed to pick up the stopwatch during a MBLD attempt to see my time if I can't see it from my position.

lgarron:
@viroulep Ping. :-)

viroulep:
Shouldn't we add something about the stopwatch?
It should be fine to let competitors touch/move it to better see it.

lgarron:
> Shouldn't we add something about the stopwatch?
> It should be fine to let competitors touch/move it to better see it.

We're still discussing that in #581.

lgarron:
Unless you were referring to the FMC stopwatch, which it appears I do need to take into account:

https://www.worldcubeassociation.org/regulations/#E3d

viroulep:
To clarify: I was referring to the stopwatch used for attempts longer than 10 minutes (such as during mbf or big blinds).

lgarron:
I'm going to merge this now, and bump the priority of the stopwatch discussion in #581.

PR #583, opened 2018-07-17
Replace "Multi-Blind" with "Multiple Blindfolded". Closes #582.

### LINKED ISSUE #582: Replace "Multiple Blindfolded" with "Multi-Blind"
Article H is called "Multiple Blindfolded Solving", and no other event name is abbreviated.

lgarron:
Article H is called "Multiple Blindfolded Solving", and no other event name is abbreviated.


AlbertoPdRF:
This one is tricky :P ["3x3x3 Multi-Blind" is the official event name](https://www.worldcubeassociation.org/regulations/#9b3), so I think the Article name should be changed instead of the changes you are proposing.

jfly (approved):
LGTM!

lgarron:
oic. The site uses "Multi-Blind", too, as did  https://docs.google.com/spreadsheets/d/1ElEEz2I2iKH3vtOV-cFn6BYklmnF11S1_OwEx_m5dNk/edit?usp=drive_web&ouid=114617249108726839773

lgarron:
The tricky part here is changing the article name without breaking old URLs. @viroulep, you know that part better know; is this possible?

viroulep:
I think it's definitely possible, the links/anchors are generated from old and new tags (the two tags before the article title [here](https://github.com/thewca/wca-regulations/blame/draft/wca-regulations.md#L516)).
So updating the article title shouldn't have an impact on anchors and links.

viroulep:
Also see [here](https://github.com/thewca/wca-regulations-compiler/blob/46ea4e8a58a201aaf57d2401fb0a6ba7318062e8/wrc/codegen/cghtml.py#L110-L113) for the part where anchors and links are generated.

lgarron:
> Also see here for the part where anchors and links are generated.

Ah, that's what I suspected – my old approach of nesting `div`s. I'd like to avoid creating another layer just to keep links backwards-compatible. 😔

viroulep:
We don't actually need another layer!
We can just change the article title for "Multi-Blind" and keep the anchors with "multiple-blindfolded".

PR #592, opened 2018-07-25
Update A3c3++) to avoid implying that stopping the timer always results in a DNF.

lgarron:
It's only a DNF if the puzzle is not solved. The puzzle can't be solved if the
competitor hasn't done any moves (due to the scramble requirements in 4b3), but
version of A3c3++) introduced by fff14fe4 didn't specify that it was referring
to that case. This commit avoids directly talking about disqualification
directly in A3c3++), focusing instead on the fact that he solve was stopped.

@AlbertoPdRF, this is a followup to #579, which you reviewed. Could you also review this?


AlbertoPdRF (approved):
Yes, it's better to be explicit about this IMO.

PR #597, opened 2018-07-28
Clarify about darkest green

### LINKED ISSUE #595: Orientation in 4d can be ambiguous
Right now regulation 4d specifies that white, green and yellow are used to determined orientation. It is not possible to have two shades of white, but what if there are two shades of green or yellow (eg. a megaminx)?

I am suggesting:
- Yellow refers to the lightest yellow shade on the puzzle
- Green refers to the darkest green shade on the puzzle or the darkest adjacent green shade
(If a megaminx had black instead of white, and cream is next to light green but not dark green, does the scrambler use cream and top and light green in front, or instead choose the darkest adjacent color in front? I have personally used a scheme like this before unofficially)

Also, for consistency, should 4d3 specify a default front color? I can see why this was a problem before because colors schemes were not consistent, but nowadays the stock color scheme on any good square 1 would require red to be in front.

pedrosino:
Fixes #595

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -166,7 +166,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - 4b3f) Pyraminx: The (random) state must require at least 6 moves to solve.
     - 4b4) Each scramble sequence should be applied during a maximum time frame of 2 hours. This time frame starts when the scramble sequence is applied for the first time.
 - 4d) Scrambling orientation:
-    - 4d1) NxNxN puzzles and Megaminx are scrambled starting with the white face (if not possible, then the lightest face) on the top and the green face (if not possible, then the darkest adjacent face) on the front.
+    - 4d1) NxNxN puzzles and Megaminx are scrambled starting with the white face (if not possible, then the lightest face) on the top and the darkest green face (if not possible, then the darkest adjacent face) on the front.
     - 4d2) Pyraminx is scrambled starting with the yellow face (if not possible, then the lightest face) on the bottom and the green face (if not possible, then the darkest adjacent face) on the front.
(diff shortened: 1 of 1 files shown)


pedrosino:
Another option is to add Guideline:
4d1+) If the puzzle has more than one green face, the darkest one should be used on front for scrambling.

xsrvmy:
So we are going with the darkest green on the puzzle?

SAuroux:
I personally consider this a negligible corner case, if anything this should only be solved via a guideline. 

The current PR is problematic by the way. A puzzle whith multiple green faces, where the darkest green is not adjacent to white will not end up being scrambled with an adjacent lighter green in front.

A correct guideline would be:

4d1+) If the puzzle has more than one green face, the darkest adjacent green should be used on front for scrambling.

wca-regulations.md
```diff
@@ -166,7 +166,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - 4b3f) Pyraminx: The (random) state must require at least 6 moves to solve.
     - 4b4) Each scramble sequence should be applied during a maximum time frame of 2 hours. This time frame starts when the scramble sequence is applied for the first time.
 - 4d) Scrambling orientation:
-    - 4d1) NxNxN puzzles and Megaminx are scrambled starting with the white face (if not possible, then the lightest face) on the top and the green face (if not possible, then the darkest adjacent face) on the front.
+    - 4d1) NxNxN puzzles and Megaminx are scrambled starting with the white face (if not possible, then the lightest face) on the top and the darkest green face (if not possible, then the darkest adjacent face) on the front.
```
lgarron: Nit: darkest adjacent green face.

lgarron:
I think a Guideline would be okay, but it only costs one more word to be clear in the Regs themselves so I'm going to go ahead and merge this with that fix.

PR #601, opened 2018-08-03
Rename "Competitors Area" to "Competitor Waiting Area".

lgarron:
The draft competition tutorial uses "waiting area". Per Leon Schmidtchen:

> Only for this case I would also not go with the term used in the Regulations. "Competitors Area" can be easily mistaken for the general practicing/socializing tables. "Waiting Area" indicates that you have to wait on those chairs inbetween attempts.

Personally, I also find "Competitors Area" to be an awkward term. This commit changes it to "Competitor Waiting Area", which can also be shortened to "waiting area" informally.

@Laura-O, does this change sound reasonable to you?

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -87,7 +87,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 2e3) Competitors who have no citizenship may compete as "Stateless".
 - 2f) Competitors must obey venue rules and conduct themselves in a considerate manner.
 - 2g) Competitors must remain quiet when inside the designated competition area. Talking is permitted, but must be kept at a reasonable level, and away from competitors who are actively competing.
-    - 2g3) Competitors in the Competitors Area must not communicate with each other about the scrambled states of the puzzles of the round in progress. Penalty: disqualification of the competitor from the event, at the discretion of the WCA Delegate.
+    - 2g3) Competitors in the Competitor Waiting Area must not communicate with each other about the scrambled states of the puzzles of the round in progress. Penalty: disqualification of the competitor from the event, at the discretion of the WCA Delegate.
 - 2h) Competitors must be fully dressed while in the competition venue. At the discretion of the WCA Delegate, competitors may be disqualified from the competition for inappropriate clothing.
 - 2i) While competing, competitors must not use electronics or audio equipment (e.g. cell phones, MP3 players, dictaphones, additional lighting).
     - 2i1) Competitors may use non-electronic aids that do not give an unfair advantage, at the discretion of the WCA Delegate. This includes:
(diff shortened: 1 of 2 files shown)


lgarron:
We can also use this opportunity to tweak the Regulations to acknowledge that there may be multiple waiting areas, but I'll make a separate change for that.

wca-regulations.md
```diff
@@ -87,7 +87,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 2e3) Competitors who have no citizenship may compete as "Stateless".
 - 2f) Competitors must obey venue rules and conduct themselves in a considerate manner.
 - 2g) Competitors must remain quiet when inside the designated competition area. Talking is permitted, but must be kept at a reasonable level, and away from competitors who are actively competing.
-    - 2g3) Competitors in the Competitors Area must not communicate with each other about the scrambled states of the puzzles of the round in progress. Penalty: disqualification of the competitor from the event, at the discretion of the WCA Delegate.
+    - 2g3) Competitors in the Competitor Waiting Area must not communicate with each other about the scrambled states of the puzzles of the round in progress. Penalty: disqualification of the competitor from the event, at the discretion of the WCA Delegate.
```
pedrosino, line 90: Maybe add a (see Article 7h) here?

wca-regulations.md
```diff
@@ -208,8 +208,8 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - 7f1d) Surface: The flat surface on which the Stackmat has been placed. The mat is considered a part of the surface. The timer is not considered a part of the surface.
     - 7f2) The Stackmat timer must be attached to the mat and placed on the surface, with the timer on the side of the mat nearest to the competitor.
         - 7f2a) Exception: For Solving With Feet, the Stackmat must be placed directly on the floor. The timer device may be placed on the side of the mat farthest from the competitor.
-- 7h) The competition area must have a Competitors Area.
-    - 7h1) The organization team may require that a competitor who has been called to compete must remain within the Competitors Area until the competitor has finished all attempts for the round.
+- 7h) The competition area must have a Competitor Waiting Area.
+    - 7h1) The organization team may require that a competitor who has been called to compete must remain within the Competitor Waiting Area until the competitor has finished all attempts for the round.
```
pedrosino, line 212: Is there any case when the organization team would *not* require this?
Perhaps the text could be merged with 7h:

> The competition area must have a Competitor Waiting Area, in which competitors who are called to compete must remain until they have finished all attempts for the round.

This would also be consistent with A2a below.

lgarron: > Is there any case when the organization team would not require this?

Yes, for rounds with individually schedule attempts (e.g. FMC/multi BLD). And competitors may also want to go to the bathroom, which is usually not a problem for the flow of the competition.

Do you have a suggestion for how to improve this, taking such details into account?

pedrosino: Ah, indeed. Here's an attempt:

7h) The competition area must have a [designated] Competitors Waiting Area, in which those who were called to compete must remain until they have finished all attempts for the round.
7h1) The competitor may be allowed to leave the [Competitors] Waiting Area for specific reasons (i.e. going to the bathroom).
7h2) In the case of individually scheduled attempts (see 2u1) it is not necessary that the competitors stay in the [Competitors] Waiting Area.

lgarron: Not a bad start! Could you create a new PR for that?
(Might want to wait until #601 and #602 have settled.)

Laura-O:
I think we had a discussion about the term a while back and concluded that the "Competitors Area" is the waiting area + the area with timer stations.

Changing this definition means that competitors can ignore the regulations when competition (i.e. they can talk about scrambles at the timer stations) and if required, they are also breaking 7h1).

lgarron:
> I think we had a discussion about the term a while back and concluded that the "Competitors Area" is the waiting area + the area with timer stations.

I think I would still prefer to specify a waiting area, but I can't think of a way to do it without introducing additional terms. I don't think it's worth the extra terminology, so I'll abandon this PR for now and update #602 accordingly.

lgarron:
> > I think we had a discussion about the term a while back and concluded that the "Competitors Area" is the waiting area + the area with timer stations.

Also, just leaving a note: interpretation doesn't square super well with:

> [7h2+](https://www.worldcubeassociation.org/regulations/guidelines.html#7h2+) ADDITION The competitors in the Competitors Area should not be able to see the puzzles of the competitors on stage.

I imagine that the folks who wrote some of the relevant Regulations/Guidelines sometimes only had the waiting area in mind.

PR #606, opened 2018-08-03
Set a time limit on restoring missing signatures, and require Delegate discretion.

lgarron:
Based on a point raised in a discussion about a recent incident: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/wca-wrc/WmaFE7da3-U/upa1meiwAQAJ


lgarron:
I think "by the end of the same day" is also reasonable, but some competitions have multi-day rounds. (In particular, most events at US Nationals have staff groups for first rounds on a separate day from competitors.)

I guess we can say "by the same calendar date as the end of the round", though? I'd prefer to keep it simple, and trust the Delegate.

Jambrose777:
The issue with on the same day as when competed, especially in the case of us nats. Is that the results aren’t on CubeComps until a future day, meaning other comeptitors have an advantage over staff members since they can sign the day of the solves and the staff members won’t find out until days later.

Till the end of the competition is also a fair option IMO

pedrosino:
I had a case recently, and wanted to bring something to discussion.

The Results Team has allowed people to ask for result corrections on times they signed on during competition, i.e. your scorecard says 12.34 but you have a video showing 12.31 or 11.34 or whatever. Or even if it has an ambiguous 1 that was entered as 7 or a flaky 5 which was entered as 9.

On the other hand, if you don't sign the scorecard, but have a video showing the attempt, no correction/reinstatement of the time is allowed.

I'm having trouble seeing why one case is fine and the other is not. Quoting freely from @SAuroux "we like to have results on the database match reality". Why then is there a different treatment/"punishment" for not signing versus signing something wrong?

The current procedure favours competitors who record all solves (or have someone record them) over those who can't do it (I for example pretty much never record anything, since I'm busy with the competition). I'm not entirely decided on this matter, but I think both scenarios should either merit a chance for fixing or no chance at all.

One could argue that making sure the signatures are there is the competitor's responsibility. The Regulations were recently amended to reflect this: _"A7c4) The competitor should not sign an attempt before the judge has recorded and signed the attempt. The competitor is ultimately responsible for ensuring that they and the judge have signed an attempt."_

But it is also the competitor's responsibility to _"...check the recorded result, and sign (or initial) the score sheet to acknowledge that the result is correct, complete, properly formatted, and clearly readable. (A7c)"_

So I guess my point is: either we make the competitor responsible and accountable for the correctness, readability, format and signatures on the sheets - meaning no corrections are allowed later - or we should also allow corrections for missing signatures if there's video evidence.

Another thing one may say is that the DNF for missing signatures is clear in the Regulations. But there's nothing in the regulations about fixing times after the competition based on video evidence. Perhaps that should be clarified as well.

lgarron:
> So I guess my point is: either we make the competitor responsible and accountable for the correctness, readability, format and signatures on the sheets - meaning no corrections are allowed later - or we should also allow corrections for missing signatures if there's video evidence.

One simple option is to change our policy not to correct times in the competitor's favor (i.e. never to a better result). This arguably discourages posting videos, but I don't think it's more of a disincentive than the current situation.

Jambrose777:
Just the result side of me, we should have our results reflect what actually happened. If video supports something then we should use it (regardless of if it helps or hurts the competitor).

The problem with a missing signature is that the competitor must sign the attempt it is a DNF. If there’s another regulation broken, that has a clear DNF penalty, he result is DNF.

pedrosino:
I could probably live with that, but I still find hard to understand why not signing is so much worse than signing on a wrong result. What is the purpose of the signatures? Isn't it to have both parties verify that the time written matches what happened? If I sign on a wrong time (i.e. I didn't check what is written), I should have no grounds for reversing that, just like I don't have if I fail to sign or to make sure the judge signed.

Or if we want to/can be lenient with that, in order to have results as accurate as possible (considering that the whole writing and signing part is not really the essence of our sport, but a formality we hope to get rid of in the future with automatic scoring of some sort), we should also be lenient with missing signatures that can be verified by other means.

Jambrose777:
There’s plenty of arguments proposed by WRC on this GitHub somewhere. Is anybody able to find and link that?

pedrosino:
> The problem with a missing signature is that the competitor must sign the attempt it is a DNF. If there’s another regulation broken, that has a clear DNF penalty, he result is DNF.

Well, it is a DNF because we put it specifically in the Regulations (as recently as Jan 1st, 2018). Keep in mind that all Regulations are created by WCA people, and not "nature rules". They are all arbitrary - some are considered more, some less arbitrary.

What bothers me is that if someone has a parent recording everything, they don't have to check very carefully, since they'd be able to fix a wrongly recorded result later with the video. If we believe the competitor should be responsible that their sheet is correct (in all aspects), we must let them deal with the in-corrections.

Laura-O:
> Till the end of the competition is also a fair option IMO

Agreed. Sometimes, results are not even entered on the same competition day, so the end of the competition is the only fair approach IMHO.

AlbertoPdRF:
To address @pedrosino concerns: the thing is that when a signature is missing, this is crystal clear to the score taker and that's why we can DNF it, while the score taker doesn't have any clue whether the written time is correct or not, thus we can't have a similar Regulation. Actually, it would somewhat be consistent to DNF a signed wrong result every time someone asks for a result correction based on video evidence, although that would only result in people not showing the videos just to not be DNF'ed.

That being said, I think I would be fine with allowing changing a DNF due to not signing to the actual result based on video evidence, although it's really not that difficult to just sign :)

And also:

> The issue with on the same day as when competed, especially in the case of us nats. Is that the results aren’t on CubeComps until a future day, meaning other comeptitors have an advantage over staff members since they can sign the day of the solves and the staff members won’t find out until days later.

I think that having staff members not signing would be a way bigger problem than the disadvantage that you mention :/

AlbertoPdRF (approved):

Jambrose777:
> I think that having staff members not signing would be a way bigger problem than the disadvantage that you mention :/

Mistakes happen.  Not every staff member is perfect.

wca-guidelines.md
```diff
@@ -172,7 +172,7 @@ To be more informative, each Guideline is classified using one of the following
 - A6e+) [CLARIFICATION] If the puzzle falls into the lap of the competitor after stopping the timer, this is considered touching the puzzle.
 - A6g+) [ADDITION] While the judge is determining whether to assign a penalty for misalignment, they should not touch the puzzle.
 - A7c+) [ADDITION] If a result signed by a competitor is found to be incomplete or unreadable, it should be interpreted the worst reasonable way (e.g. "1:05" will be considered 1:05.99, "25.X3" will be considered 25.73 if X could be either a 1 or a 7). Given that the Stackmat timer displays a time of X:0Y.ss as X:Y.ss, a missing tens digit should be considered more likely than a missing ones digit (e.g. "1:2.27" will be considered 1:02.27). The judge who signed an ambiguous result may be requested to clarify it, at the discretion of the score taker.
-- A7c4+) [ADDITION] Missing signatures may be rectified if the responsible judge and/or competitor can be tracked, at the discretion of the score taker.
+- A7c4+) [ADDITION] Missing signatures found by the score taker may be added by the judge and/or competitor before the end of the competition, at the discretion of the Delegate.
```
Claster: For consistency, it is "WCA Delegate" everywhere else in regs.

Claster:
lgtm otherwise

wca-guidelines.md
lgarron: Thanks, I've fixed that!

lgarron:
I think there are some fair points made, but I think this change is strictly better than what we currently have. Since @Claster approved it, I'm going to merge and we can discuss other approaches in other issues.

PR #613, opened 2018-08-20
9f2) Change "nearest second" rounding examples to x.49 and x.50

lgarron:
This makes the range from 0.40 to 0.49 extra-clear, and matches the precision that mose competition stopwatches have.

@viroulep, could you review?


Claster:
lgtm

PR #615, opened 2018-08-20
Update wca-states.md to show one list without descriptions

Mollerz:
Current ongoing discussion - seems like the route we are headed is to keep the list of countries the same, however remove the descriptions and reasons for allowing certain countries that are not UN Member/Observer States to avoid any political bias or misinformation.

### DIFF
--- a/wca-states.md
+++ b/wca-states.md
@@ -1,10 +1,13 @@
 # <wca-title>WCA List of Recognized Countries
 <wca-states>
 <version>Version: June 19, 2017
+  
+  
+The WCA recognizes a total of 199 states/regions:
 
 
-<states-list> The WCA recognizes a total of 199 countries.
-
+## <states-list> 
+  
 - (AF:_Asia) Afghanistan
 - (AL:_Europe) Albania
 - (DZ:_Africa) Algeria
--- a/wca-states.md
+++ b/wca-states.md
@@ -1,13 +1,13 @@
 # <wca-title>WCA List of Recognized Countries
 <wca-states>
 <version>Version: June 19, 2017
-  
-  
-The WCA recognizes a total of 199 states/regions:
+
+
+The WCA recognizes a total of 199 countries/regions:
 
 
 ## <states-list> 
-  
+
 - (AF:_Asia) Afghanistan
 - (AL:_Europe) Albania
 - (DZ:_Africa) Algeria
--- a/wca-states.md
+++ b/wca-states.md
@@ -3,10 +3,7 @@
 <version>Version: June 19, 2017
 
 
-The WCA recognizes a total of 199 independent countries.
-
-
-## <states-list>UN Member States:
+The WCA recognizes a total of 199 countries.
 
 - (AF:_Asia) Afghanistan
 - (AL:_Europe) Albania
@@ -81,7 +78,9 @@ The WCA recognizes a total of 199 independent countries.
 - (GW:_Africa) Guinea Bissau
 - (GY:_South America) Guyana
 - (HT:_North America) Haiti
+- (VA:_Europe) Holy See
 - (HN:_North America) Honduras
+- (HK:_Asia) Hong Kong
 - (HU:_Europe) Hungary
 - (IS:_Europe) Iceland
 - (IN:_Asia) India
@@ -97,6 +96,7 @@ The WCA recognizes a total of 199 independent countries.
 - (KZ:_Asia) Kazakhstan
 - (KE:_Africa) Kenya
 - (KI:_Oceania) Kiribati
+- (XK:_Europe) Kosovo
 - (KW:_Asia) Kuwait
 - (KG:_Asia) Kyrgyzstan
 - (LA:_Asia) Laos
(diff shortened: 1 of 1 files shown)


wca-states.md
```diff
@@ -3,10 +3,7 @@
 <version>Version: June 19, 2017
 
 
-The WCA recognizes a total of 199 independent countries.
-
-
-## <states-list>UN Member States:
```
jfly: The build is failing (https://travis-ci.org/thewca/wca-regulations/builds/418300471#L474), presumably because you removed the `## <states-list>` token. Maybe try changing line 6 to:

```
<states-list> The WCA recognizes a total of 199 countries.
```

?

jfly (changes requested):

jfly:
Thanks for taking care of this, @Mollerz!

wca-states.md
Mollerz: Yeah I don't know how coding works :P I will add that back in, it makes sense!

jfly: Ok, sorry, nevermind! It looks like the compiler expects you to include the `##` as well. (relevant code here: https://github.com/thewca/wca-regulations-compiler/blob/6b6aee02de716bb2c7a1b1626200b61bcc364bee/wrc/parse/lexer.py#L77). I think to get the build passing, you'll have to change this to:

```
## <states-list> The WCA recognizes a total of 199 countries.
```

(and I'd consider changing that period at the end to a colon)

It's also worth noting that adding the `##` will change how this looks over on https://www.worldcubeassociation.org/regulations/countries/ (the sentence will look like a heading), which may be undesirable.

@viroulep because he's the expert on this.

Mollerz:
We should also have a private page that probably board/WRC can access so that internally we understand the reasoning for having this list the way it is. In the future when new countries are born, or if there are any changes we need to make, people can easily see the reasoning behind our current list instead of having to reason it from scratch themselves.

jfly:
James, it looks like the build is still failing, perhaps due to some trailing whitespace?

I think it would be great for us to have an internal document for this, I think a Google doc would be the right place for this for now.

Mollerz:
I will create the Google Doc - I think waiting for @viroulep might be best to see what we can do about it failing :)

wca-states.md
```diff
@@ -1,13 +1,13 @@
 # <wca-title>WCA List of Recognized Countries
 <wca-states>
 <version>Version: June 19, 2017
+  
+  
+The WCA recognizes a total of 199 states/regions:
```
lgarron: I would prefer to avoid "region", since we already use "region" to refer to explicitly refer to nations, continents, and the world: https://www.worldcubeassociation.org/regulations/#9i1

Could we just say "states?

wca-states.md
```diff
@@ -1,13 +1,13 @@
 # <wca-title>WCA List of Recognized Countries
```
lgarron: Should the title be updated?

wca-states.md
Baiqiang: @lgarron If we used only one word, Chinese people would treat China and Taiwan as the same kind/level of a state (or some word else). This is like the current word "country".

viroulep: @lgarron please check the current email thread about this matter, were this specific wording is being discussed (we'd rather have the discussion at one single place).

viroulep:
Indeed the build is simply failing because of trailing whitespaces on lines 4, 5, and 10.

jfly (approved):

wca-states.md
```diff
@@ -3,10 +3,10 @@
 <version>Version: June 19, 2017
 
 
-The WCA recognizes a total of 199 independent countries.
+The WCA recognizes a total of 199 countries/regions:
 
 
-## <states-list>UN Member States:
+## <states-list>
```
jfly, line 9: Looks like the build is still failing, unfortunately. I think that's because the regex here: https://github.com/thewca/wca-regulations-compiler/blob/6b6aee02de716bb2c7a1b1626200b61bcc364bee/wrc/parse/lexer.py#L77 is expecting there to be at least one character after `<states-list>`. I think there are two ways we can fix this:

1. Change the wca-regulations-compiler to not require anything after `<states-list>`. If we're going to that effort, then it probably doesn't make sense for it to require the `##` anymore, either.
2. Combine lines 6 and 9 so this looks like:

```
## <states-list> The WCA recognizes a total of 199 countries/regions:
```

Mollerz: That makes sense - we don't have to format it with ## either right? We don't want to make this a huge title text after all.

jfly: @viroulep, do you see any issues with changing the regs compiler as I've described in 1) above?

cubewhiz:
This change has been approved by the Board, so when the code is ready, you can proceed with this change.

wca-states.md
viroulep: Fixed in thewca/wca-regulations-compiler@3ebbd8fe8fec7c9167296f59b2677696fe61a954.

viroulep:
I just noticed this PR has been created against the `draft` branch, which is not what we want in this case (see [here](https://github.com/thewca/wca-regulations/wiki/On-submitting-pull-requests) why) as we want it to land in `official` without going through the Regulations release process.

I'll see what I can do.

viroulep:
Superseded by #621.

The commits were on top of all the draft changes that we don't want in official yet, so what I did is the following:
  - squashed James commits into one clean commit
  - cherry-picked this commit on top of official, and submitted #621

The diff is the same, so I'll go ahead and publish this!

PR #612, opened 2018-08-20
Consolidate time limit definitions in the Regulations. Closes #611.

lgarron:
This moves many details from the Guidelines to the Regulations. It also includes
two functional changes:

- Time limits must be announced before the competition, rather than just before the
  round.
- Only one cumulative time limit can apply to to a given round.

Both of these are almost always (or always?) followed in practice.

PR #628, opened 2018-09-30
Update event names throughout text

micahstairs:
Update event names throughout text to match official event names.

Some of the events were being referenced using inconsistent wording (in particular, the 3x3x3 Multi-Blind event). This change resolves these inconsistencies.

### DIFF
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -191,7 +191,7 @@ To be more informative, each Guideline is classified using one of the following 
 - C1b+++) [REMINDER] Use of the surface is permitted while solving or repairing the puzzle.
 
 
-## <article-D><feet><solvingwithfeet> Article D: Solving With Feet
+## <article-D><feet><solvingwithfeet> Article D: Feet Solving
 
 - D1b+) [CLARIFICATION] The competitor may wear socks while solving.
 - D1b++) [CLARIFICATION] In the past, competitors were required to inspect the puzzle using their feet for 3x3x3 With Feet. They are now permitted to touch the puzzle with their hands or any other part of their body during inspection (see [Guideline A5b+](guidelines:guideline:A5b+)).
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -179,9 +179,9 @@ To be more informative, each Guideline is classified using one of the following 
 
 
 
-## <article-B><blindfolded><blindfoldedsolving> Article B: Blindfolded Solving
+## <article-B><blindfolded><blindfoldedsolving> Article B: 3x3x3 Blindfolded
 
-- B1+) [REMINDER] The competitor must use a puzzle without textures, markings, or other features that distinguish similar pieces (see [Regulation 3k](regulations:regulation:3k)). This should be given special attention for Blindfolded Solving.
+- B1+) [REMINDER] The competitor must use a puzzle without textures, markings, or other features that distinguish similar pieces (see [Regulation 3k](regulations:regulation:3k)). This should be given special attention for 3x3x3 Blindfolded.
 - B1b+) [RECOMMENDATION] Blindfolds should be checked by the WCA Delegate before use in the competition.
 
 
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -192,7 +192,7 @@ To be more informative, each Guideline is classified using one of the following 
 - C1b+++) [REMINDER] Use of the surface is permitted while solving or repairing the puzzle.
 
 
-## <article-D><feet><solvingwithfeet> Article D: Solving with Feet
+## <article-D><feet><solvingwithfeet> Article D: Solving With Feet
 
 - D1b+) [CLARIFICATION] The competitor may wear socks while solving.
 - D1b++) [CLARIFICATION] In the past, competitors were required to inspect the puzzle using their feet for 3x3x3 With Feet. They are now permitted to touch the puzzle with their hands or any other part of their body during inspection (see [Guideline A5b+](guidelines:guideline:A5b+)).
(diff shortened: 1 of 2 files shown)


lgarron:
Hi @micahstairs, I haven't merged this because it's not quite complete. :-(
For example, "Solving with Feet" still appears a few times, and the Guidelines aren't updated at all. Would you possibly have a chance to do another full pass in the next few days?

micahstairs:
Thanks for pointing out that I missed something!

I've fixed up both of the files. Can you take another look?

lgarron:
@AlbertoPdRF, could you review this?

If you have more time, it would also be good to figure out how to adapt https://github.com/thewca/wca-regulations/pull/630 based on this.

AlbertoPdRF:
Sure, I'll take care of this, hopefully during this week!

wca-guidelines.md
```diff
@@ -185,21 +185,21 @@ To be more informative, each Guideline is classified using one of the following
...
 - C1b+) [CLARIFICATION] The competitor may use both hands during inspection.
 - C1b++) [CLARIFICATION] The competitor is not required to use the same solving hand for different attempts of the same round.
 - C1b+++) [REMINDER] Use of the surface is permitted while solving or repairing the puzzle.
 
 
-## <article-D><feet><solvingwithfeet> Article D: Solving With Feet
+## <article-D><feet><solvingwithfeet> Article D: 3x3x3 With Feet
 
 - D1b+) [CLARIFICATION] The competitor may wear socks while solving.
 - D1b++) [CLARIFICATION] In the past, competitors were required to inspect the puzzle using their feet for Solving with Feet. They are now permitted to touch the puzzle with their hands or any other part of their body during inspection (see [Guideline A5b+](guidelines:guideline:A5b+)).
```
```suggestion
- D1b++) [CLARIFICATION] In the past, competitors were required to inspect the puzzle using their feet for 3x3x3 With Feet. They are now permitted to touch the puzzle with their hands or any other part of their body during inspection (see [Guideline A5b+](guidelines:guideline:A5b+)).
```

AlbertoPdRF (commented):
I have an issue with this as it is. For instance, we change `One-Handed Solving` to `3x3x3 One-Handed`, but we leave `Blindfolded Solving` as is. I get why this is this way, but I would prefer to not have such inconsistency.

Opinions?

micahstairs:
Thanks for pointing out that inconsistency Pedro. I've replaced "Blindfolded Solving" with "3x3x3 Blindfolded". This is also now consistent with the event list  in the dropdowns on https://www.worldcubeassociation.org.

AlbertoPdRF:
The problem is that Article B not only applies to 3x3x3 Blindfolded, but to all Blindfolded events. I think keeping the old names (with small tweaks) for articles may make more sense.

micahstairs:
> The problem is that Article B not only applies to 3x3x3 Blindfolded, but to all Blindfolded events. I think keeping the old names (with small tweaks) for articles may make more sense.

Ah good point. Okay, I've fixed that and reverted the article names back to the original names. It wasn't obvious to me which "small tweaks" you wanted to make to the old names.

AlbertoPdRF:
I think the only necessary tweak is Article H's name, which I would rename to "Multi-Blind Solving". Also, can you fix the conflict in `wca-regulations.md`?

Thanks for the hard work @micahstairs!

lgarron:
Note: now that https://github.com/thewca/wca-regulations/pull/612 has landed, A1a2) will also need an update. (And you'll actually have to merge conflicts for this branch. gitHub has a UI for this, but let me know if you want help with that.)

micahstairs:
Okay I did a sync and fixed the merge conflicts. I also updated A1a4 (I think you meant that one instead of A1a2).

micahstairs:
I've also addressed @AlbertoPdRF's suggestion to rename Article H.

Let me know if there's anything else that needs done, but I think we're good.

wca-regulations.md
```diff
@@ -489,7 +489,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 
 ## <article-E><fewest-moves><fewestmovessolving> Article E: Fewest Moves Solving
 
-- E2) Procedure for Fewest Moves Solving:
+- E2) Procedure for 3x3x3 Fewest Moves:
```
lgarron, line 493: This is now inconsistent with the article title. I think we should keep it as-is.

lgarron (changes requested):
Thanks for bearing with all the revisions!

We're still not quite done – I noticed that Article E still needs to be renamed in the Guidelines. Would you mind doing another very thorough pass and double-checking that you've caught every instance?

AlbertoPdRF:
Also, when you did this last review, could you squash all the commits into one?

lgarron:
Also just added: #630 affects 9f1 and 9f2.

micahstairs:
@lgarron, I took care of 9f1 and 9f2 and I did a full pass over the regulations and guidelines. I changed the article from "Solving With Feet" to "Feet Solving" for consistency with the other article naming.

I'm about to squash the commits (once I figure out how to do so).

micahstairs:
@AlbertoPdRF I'm not sure how to squash them. It looks like the person merging this in is the one who is supposed to press "Squash and Merge" (at least that's from I'm gathering from a Google search). Mind pointing me in the right direction?

AlbertoPdRF:
I see that we indeed have the option to "Squash and Merge" so maybe it's just easier to do it that way

AlbertoPdRF:
As this PR has become difficult to follow, I opened #672 resolving the conflicts and squashing everything. If that's ok, I think the best is to close this PR and continue there.

AlbertoPdRF:
Closing this one in favor of #672.

PR #648, opened 2018-11-27
Update H1d+/H1d++ to avoid implying that the multi BLD time is the full result.

lgarron:


AlbertoPdRF (approved):
LGTM!

PR #651, opened 2018-11-29
Require the Delegate to restrict scramble access to groups before they start.

lgarron:
@AlbertoPdRF and @Laura-O, could you review?

### DIFF
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -39,6 +39,7 @@ To be more informative, each Guideline is classified using one of the following 
 - 1c3++) [ADDITION] Physical records (e.g. score sheets, written solutions from Fewest Moves Solving) must be retained for a minimum of one month. Digital records (e.g. scramble sequences, results) must be retained permanently.
 - 1c3b+) [CLARIFICATION] If there are multiple groups, it is not necessary to identify which competitor was in which groups.
 - 1c4+) [RECOMMENDATION] Corrections to the results should be available within one week of the competition date.
+- 1c8+) [ADDITION] The Delegate must only give access to scramble sequences for future groups to other members of the organization team if they will be temporarily unavailable (e.g. competing themselves), and should give others access to the minimum possible amount of scrambles in this case.
 - 1c10+) [CLARIFICATION] It is sufficient to ensure access to a digital copy of the Regulations.
 - 1h+) [RECOMMENDATION] Competitors in the same group should use the same scramble sequences. Different groups should use different scramble sequences.
 - 1h++) [RECOMMENDATION] All attempts in the final round of an event, as well as all Fewest Moves Solving attempts, should have the same scramble sequences for all competitors (i.e. only 1 group).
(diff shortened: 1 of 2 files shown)


wca-guidelines.md
```diff
@@ -39,6 +39,7 @@ To be more informative, each Guideline is classified using one of the following
 - 1c3++) [ADDITION] Physical records (e.g. score sheets, written solutions from Fewest Moves Solving) must be retained for a minimum of one month. Digital records (e.g. scramble sequences, results) must be retained permanently.
 - 1c3b+) [CLARIFICATION] If there are multiple groups, it is not necessary to identify which competitor was in which groups.
 - 1c4+) [RECOMMENDATION] Corrections to the results should be available within one week of the competition date.
+- 1c8+) [EXAMPLE] The Delegate should only give access to scrambles for future groups to other members of the organization team if they will be temporarily unavailable (e.g. competing themselves), and should give others access to the minimum possible amount of scrambles in this case.
```
Laura-O: I think `scrambles` should be `scramble sequences` to have a consistent terminology.

Additionally, why `should` and not `must`?

Laura-O (commented):
Two comments/questions. The rest LGTM!

xsrvmy:
Does this imply that the delegate should email each scramble group at a time to each scrambling station separately? I have never ever seen this done at a local competition.

Laura-O:
> Does this imply that the delegate should email each scramble group at a time to each scrambling station separately? I have never ever seen this done at a local competition.

But you have seen delegates emailing all scrambles? That's interesting, tell us more about it.

xsrvmy:
> > Does this imply that the delegate should email each scramble group at a time to each scrambling station separately? I have never ever seen this done at a local competition.
> 
> But you have seen delegates emailing all scrambles? That's interesting, tell us more about it.

When I scramble at competitions it has always been to scrambler's responsibility to open the correct pdf. Of course the delegate/organizer will tell us to switch groups. However scramblers do need to be approved by the delegate/organizer first.

AlbertoPdRF (approved):
Apart from @Laura-O's comments, LGTM!

wca-guidelines.md
lgarron: Done. Since this is in the Guidelines, I've also changed this from `[EXAMPLE]` to `[ADDITION]`.

xsrvmy:
Shouldn't a MUST entry go in the regulations instead of the guidelines?
Also, if a delegate accidentally fails to respect this regulation before the new encrytion scheme is used in tnoodle, does the entire comp become disqualified?

lgarron:
> Shouldn't a MUST entry go in the regulations instead of the guidelines?

It should, but sometimes we make exceptions. This one's an addition to a SHOULD in the Regulations, and I don't feel comfortable codifying the exception until we've seen what happens in practice.

> Also, if a delegate accidentally fails to respect this regulation before the new encrytion scheme is used in tnoodle, does the entire comp become disqualified?

No (unless scrambles were leaked).

PR #663, opened 2018-12-06
Allow competitors to write their name while submitting an FMC solution.

lgarron:


wca-guidelines.md
```diff
@@ -204,6 +204,7 @@ To be more informative, each Guideline is classified using one of the following
...
 - E2c++) [CLARIFICATION] The competitor's solution must only use moves that are exactly defined in [Regulation 12a](regulations:regulation:12a). Examples of notation and moves that are not permitted in competition: [F], [R, U], [R: U], R'2, L'w, f, M, U2'.
+- E2c+++) [ADDITION] If the competitor forgot to write their name on their intended solution paper, they may be allowed to add their name, at the discretion of the judge. In this case, the competitor must ask the judge while submitting their solution and then immediately write their name (and make no other changes) under direct supervision of the judge. Once a competitor has finished submitting their solution (e.g. placing it in the judge's hands and allowing the judge to move on to the next competitor, placing it on the judge's table and walking away), they are not permitted to add their name. If the competitor writes their name after the time limit but before asking the judge, the attempt remains disqualified (DNF).
```
AlbertoPdRF: Is the discretion of the judge necessary here? I would always allow it, thus preventing any kind of abuse of power. Also, having two examples makes it a bit difficult to read in my opinion, so I would just have one.

lgarron: Done; could you review again?

AlbertoPdRF (approved):
LGTM!

PR #671, opened 2018-12-15
Get rid of BoX formats in 9b1a).

### LINKED ISSUE #658: Get rid of the exception in 9b1a)
These formats have been requested just a couple times during 2018 and I think an acceptable reason to hold them was never provided, so all the requests were declined. If we can't find any good reason to allow holding these formats, the best would be to just get rid of the exception that makes it possible to request them.

AlbertoPdRF:
Fixes #658.

### DIFF
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -126,7 +126,6 @@ To be more informative, each Guideline is classified using one of the following 
 
 ## <article-9><events><events> Article 9: Events
 
-- 9b+) [ADDITION] The preferred format for the final of an event is "Average of 5" or "Mean of 3", if possible.
 - 9b++) [ADDITION] Events other than those specified in [Regulation 9b](regulations:regulation:9b) may be held during a competition, but will be considered unofficial and therefore will not be included in the official results of the competition.
 - 9b3b+) [CLARIFICATION] The "Mean of 3" statistic does not affect the competitor's ranking in a "Best of 3" round (which is based on the best result). The WCA recognizes it outside of the competition format.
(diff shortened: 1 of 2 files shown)


wca-guidelines.md
```diff
@@ -126,7 +126,6 @@ To be more informative, each Guideline is classified using one of the following
 
 ## <article-9><events><events> Article 9: Events
 
-- 9b+) [ADDITION] The preferred format for the final of an event is "Average of 5" or "Mean of 3", if possible.
```
lgarron, line 129: Could we keep "Average of 5" to clarify that combined rounds are not preferred for finals?

AlbertoPdRF: But are they? I think we use them quite commonly, and I don't see a problem with it.

viroulep: It would be good to check the db to have a more precise overview, but I've seen a whole lot of combined final rounds for events which has only one round (typically 5/6/7/clock around here).
If events have multiple rounds, I definitely agree that combined rounds should be avoided for finals.

AlbertoPdRF: Maybe we can reword it a bit to make that clear?

lgarron: How's this?

```suggestion
- 9b+) [ADDITION] If an event has multiple rounds, every round after the first round (especially the final round) should use the "Average of 5" format.
```

AlbertoPdRF: I think I would keep the "Mean of 3" too, otherwise LGTM!

lgarron: Ah, sorry, I was somehow thinking 9b+ was "9b1+".

```suggestion
- 9b+) [ADDITION] For an event with multiple rounds, every round after the first round (especially the final round) should be "Average of 5" or "Mean of 3" if one of these is a valid format.
```

lgarron (approved):

lgarron:
@AlbertoPdRF, could you see if you're okay with my "Mean of 3" revision and merge the change if so?

AlbertoPdRF:
Yes, sure!

PR #676, opened 2018-12-17
Add a Guideline about visual isolation of the scramblers.

lgarron:
This has come up in discussion with the WQAC, as one minor issue among several issues with scramble secrecy. I have seen some pretty sloppy competition setups where competitors could easily watch puzzles being scrambled without looking suspicious.

We had similar issues with sloppiness back when it was acceptable to use bare scorecards as "covers", but adding "makes it impossible for any competitors or spectators" to [A2c1](https://www.worldcubeassociation.org/regulations/guidelines.html#A2c1) fixed that. (See [A2c1+](https://www.worldcubeassociation.org/regulations/guidelines.html#A2c1+)).

Since A2c already mostly requires this and I'm introducing it as a "should", I hope this is uncontroversial enough for 2019.


AlbertoPdRF (approved):
LGTM!

Issue #584, opened 2018-07-17
What stickers can be used in FMC and how?


lgarron:
- Can the stickers be different shapes?
- Can they be colored? Patterned? Textured?
- Can you write on them?
- Can they be numbered ahead of time?
  - If so, can they contain more information? Is there a clear boundary for what counts as [assistance](https://www.worldcubeassociation.org/regulations/#A5b)?
- Can you do anything to the puzzles except apply and peel the additional stickers?

Branched out from https://github.com/thewca/wca-regulations/pull/538#issuecomment-405521337

xsrvmy:
"What counts as assistance?"
The wording in the requirement for logos is probably good enough.

xsrvmy:
"Can the stickers be different shapes?" Modern cube sticker sets are different shapes to begin with. If not, they can always be distorted by the competitor.

Also, an additional question: Can you write on the puzzle?

I think disallowing distinguishing the stickers at all partially defeats the purpose of the stickers for insertion finding, especially for anything more than a 3 cycle.

theoleinad:
Checking the reg: 
**E3c) Stickers (self-supplied).**

So they can be any shape. 
Any Color, with pattern or texture.
Yes, we can write on the stickers and we can numbered them before the attempt.
_(Assistance comes from speedsolving not fewest moves solving, so N/A)_
You basically can turn apart the cube, and reassemble if you want, there's no limitations on this.

If we want to clarify, maybe we should just clarify that basic labeled stickers are allowed, I personally use: 
1,2,3...8,9 (corners)
A,B,C,D,E,F (edges)
X, X, X... (free)

So we avoid algorithms written in the stickers... 
(Which I have thought)

Shall we limit number of stickers as well? 
20 sounds enough for one attempt for me... 

Also, we could state that all of this items are upon delegate's approval.

Laura-O:
> Shall we limit number of stickers as well?
> 20 sounds enough for one attempt for me...

Looking forward to adding an hour to each FMC attempt because I have to count the stickers of each competitor first! 😃 

But seriously, any additional restrictions regarding stickers might sound nice in theory but it's completely unrealistic to check that before, and also during an FMC attempt. Competitors often bring several sheets of stickers which makes it impossible to check the texture, color or writings on all of them. I think we all agree that writing down algorithms is a 100% clear case of assistance, so we don't need to discuss this.

[...]

PR #678, opened 2018-12-18
 Overhaul Fewest Moves

lgarron:
- Clean up the structure.
- Define "information to identify the attempt" (E2c1): name or WCA ID, and optionally comp/round/attempt #.
- Specify when outside competitor may write information to identify the attempt:
  - Before the attempt. (E2a1)
  - During the attempt (implicit)
  - While turning in a solution, in front of the judge. (E2c1+)
  - NOT after turning in a solution. (E2c1+)
- Define an unambiguous, valid solution. (E2c2 to E2c5)
- Specify that a competitor should black out/scribble out extra moves. (E2c3) (This is a
  "should", but placing it in the Regulations will hopefully encourage
  competitors to do this move consistently, leading to fewer frustrating calls
  while grading).
- Allow judges to interpret `{r}`, `<r>`, or `(r)` as `[r]` for inexperienced competitors. (E2c2++)


lgarron:
@Laura-O, would you mind reviewing this change?
(Or asking someone else to, if you won't have time during the next day?)

@thewca/wrc-team: this is a pretty substantial overhaul, so I'd appreciate a review from anyone who has input!

wca-guidelines.md
```diff
@@ -205,10 +205,15 @@ To be more informative, each Guideline is classified using one of the following
...
+- E2c2++) [ADDITION] If a competitor with little 3x3x3 Fewest Moves experience uses the incorrect type of brackets/parentheses to write a rotation, the judge may interpret the competitor's intention as the correct brackets (and the competitor should be informed), at the discretion of the Delegate. Example: {r}, <r>, or (r) may be interpreted as [r]. Note that only the bracket type may reintepreted this way, and notation such as or [R], [r]2 is still considered invalid (see [Guideline E2c4+](guidelines:guideline:E2c4+)).
```
lgarron:
```suggestion
 - E2c2++) [ADDITION] If a competitor with little 3x3x3 Fewest Moves experience uses the incorrect type of brackets/parentheses to write a rotation, the judge may interpret the competitor's intention as the correct brackets (and the competitor should be informed), at the discretion of the Delegate. Example: {r}, <r>, or (r) may be interpreted as [r]. Note that only the bracket type may reinterpreted this way, and notation such as or [R], [r u], or [r]2 is still considered invalid (see [Guideline E2c4+](guidelines:guideline:E2c4+)).
```

wca-guidelines.md
```diff
@@ -205,10 +205,15 @@ To be more informative, each Guideline is classified using one of the following
...
+- E2c2+) [EXAMPLE] Examples that are not considered a single unambiguous move sequence: multiple solution sequences without exactly one clearly indicated solution, a solution with ambiguous symbols (e.g. a letter that looks similar to "B" and "R" but is not clearly one or the other), a solution with stars or long arrows indicating moves to be performed out of written order (e.g. "insertions" that are not written in-line with the rest of the move sequence, "pre-moves" that are not incorporated into the move sequence using valid notation), a move sequence that is not written out roughly as a series of lines in reading order, any arrangement of symbols that cannot be interpreted as an unambiguous move sequence. Exceptions to the examples above: small typographic corrections (e.g. blacking out a move and writing a move above that is clearly meant to replace it, moves written in subscript or superscript to fit between other moves).
```
AlbertoPdRF: Do we want the `moves written in subscript or superscript to fit between other moves`? This feels a bit weird to me, as it could turn out into `a move sequence that is not written out roughly as a series of lines in reading order` or `any arrangement of symbols that cannot be interpreted as an unambiguous move sequence`.

wca-guidelines.md
```diff
@@ -205,10 +205,15 @@ To be more informative, each Guideline is classified using one of the following
...
+ - E2c2++) [ADDITION] If a competitor with little 3x3x3 Fewest Moves experience uses the incorrect type of brackets/parentheses to write a rotation, the judge may interpret the competitor's intention as the correct brackets (and the competitor should be informed), at the discretion of the Delegate. Example: {r}, <r>, or (r) may be interpreted as [r]. Note that only the bracket type may reinterpreted this way, and notation such as or [R], [r u], or [r]2 is still considered invalid (see [Guideline E2c4+](guidelines:guideline:E2c4+)).
```
```suggestion
  - E2c2++) [ADDITION] If a competitor with little 3x3x3 Fewest Moves experience uses the incorrect type of brackets/parentheses to write a rotation, the judge may interpret the competitor's intention as the correct brackets (and the competitor should be informed), at the discretion of the WCA Delegate. Example: {r}, <r>, or (r) may be interpreted as [r]. Note that only the bracket type may reinterpreted this way, and notation such as or [R], [r u], or [r]2 is still considered invalid (see [Guideline E2c4+](guidelines:guideline:E2c4+)).
```

AlbertoPdRF (approved):
A comment and a nitpick, but overall LGTM!

wca-guidelines.md
Laura-O: > long arrows indicating moves to be performed out of written order (e.g. "insertions" that are not written in-line with the rest of the move sequence, "pre-moves" that are not incorporated into the move sequence using valid notation)

I think this is not 100% clear and difficult to judge:
1. What is a *long* arrow? Are insertions written with a *short* arrow fine?
2. I think not all Delegates know what exactly an insertion is or how it looks like, so this might not be applied consistently.

wca-guidelines.md
Laura-O: Also, that's quite unclear: what is *little experience*?
I would suggest to either remove it completely (i.e. allow it for all competitors) or using the same term as in A7g (`caused by a new competitor's inexperience`).

Laura-O (changes requested):

wca-guidelines.md
lgarron: > Do we want the `moves written in subscript or superscript to fit between other moves`?

These are intended to cover the cases where a competitor doesn't *quite* write things in lines, but does something that is clearly intended to be read that way: https://github.com/thewca/wca-regulations/pull/575#issuecomment-407871264

After finishing the 2019 draft I want to work on an FMC [guide](https://github.com/thewca/wca-regulations/issues/236), where I want to include examples like the ones more clear.

> 1. What is a _long_ arrow? Are insertions written with a _short_ arrow fine?

Yeah, the intention is to allow simple things like in your image (where moves aren't exactly "written out of order, but definitely use small arrows). However, I wrote this before I added the "small typographic corrections" part, which I believe covers it. I've taken out "long".

> 2\. I think not all Delegates know what exactly an insertion is or how it looks like, so this might not be applied consistently.

I would like to leave this in, since it's a well-known term in most of the community (similar to "stickerless") and makes the intent pretty clear.

wca-guidelines.md
lgarron: I'd really like to avoid implying that it's okay to use invalid notation.
I've changed this to "due to inexperience".

lgarron:
@Laura-O, could you review again? :-D

wca-guidelines.md
```diff
@@ -205,10 +205,15 @@ To be more informative, each Guideline is classified using one of the following
...
+- E2c2+) [EXAMPLE] Examples that are not considered a single unambiguous move sequence: multiple solution sequences without exactly one clearly indicated solution, a solution with ambiguous symbols (e.g. a letter that looks similar to "B" and "R" but is not clearly one or the other), a solution with stars or arrows indicating moves to be performed out of written order (e.g. "insertions" that are not written in-line with the rest of the move sequence, "pre-moves" that are not incorporated into the move sequence using valid notation), a move sequence that is not written out roughly as a series of lines in reading order, any arrangement of symbols that cannot be interpreted as an unambiguous move sequence. Exceptions to the examples above: small typographic corrections (e.g. blacking out a move and writing a move above that is clearly meant to replace it, moves written in subscript or superscript to fit between other moves).
```
SAuroux: How about the more general case of blacking out _m_ moves and writing _n_ moves above them?

Obviously, the current version allows for _m = n_ and _m > n_. What about _m < n_?

How about using plural instead to cover this (and allow for last minute adjustments): 

`blacking out moves and writing moves above that are clearly meant to replace them`

wca-regulations.md
```diff
@@ -496,10 +496,16 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
...
     - E2b) All competitors have a total time limit of 60 minutes to find and write a solution.
         - E2b1) The judge should call "5 MINUTES REMAINING" at 55 minutes, and must call "STOP" at 60 minutes.
-    - E2c) At 60 minutes, each competitor must submit to the judge a single piece of paper with the competitor's written name (optionally accompanied by WCA ID) and solution, using the notation for the 3x3x3 Cube described in [Regulation 12a](regulations:regulation:12a). Penalty: disqualification of the attempt (DNF).
+    - E2c) At 60 minutes, each competitor must submit to the judge a single piece of paper with a written solution and information to identify the attempt.
+        - E2c1) Information to identify the attempt is: the competitor's name or WCA ID (at least one, optionally both), and optionally the competition name, round, and/or attempt number. Penalty for a submitted solution without a name or WCA ID: disqualification of the attempt (DNF).
```
SAuroux, line 504: Technically, "name" is quite ambiguous. Surely, we cannot require the exact WCA names (local names, Brazilians, etc.). Out of curiosity, which of the following "names" would you consider valid for myself: "S. Auroux", "Sébastien A.", "Sébastien", "Séb".

wca-guidelines.md
lgarron:
```suggestion
- E2c2+) [EXAMPLE] Examples that are not considered a single unambiguous move sequence: multiple solution sequences without exactly one clearly indicated solution, a solution with ambiguous symbols (e.g. a letter that looks similar to "B" and "R" but is not clearly one or the other), a solution with stars or arrows indicating moves to be performed out of written order (e.g. "insertions" that are not written in-line with the rest of the move sequence, "pre-moves" that are not incorporated into the move sequence using valid notation), a move sequence that is not written out roughly as a series of lines in reading order, any arrangement of symbols that cannot be interpreted as an unambiguous move sequence. Exceptions to the examples above: small typographic corrections (e.g. blacking out moves and writing moves above that are clearly meant to replace them, moves written in subscript or superscript to fit between other moves).
```

SAuroux:
I'm a little lost regarding writing identity information while turning in a solution, in front of the judge. Unless I am missing something, this is currently no longer included/planned?

lgarron:
> I'm a little lost regarding writing identity information while turning in a solution, in front of the judge. Unless I am missing something, this is currently no longer included/planned?

Looks like I lost E2c1+ somewhere in one of the draft commits; thanks for noticing. 😳

I've added it back.

campos20:
[perhaps this isn't the place?] I'm not particularly crazy about how we allow movements that are not listed like Rw. Explaining notations seems way too unnecessary. I think we should give some examples: F, F2, F', Fw, but imo there's no reason to list all the 18 face moves and all 18 cube rotations.

Just to remember, if the change on regs lead to change on layout, we perhaps need to call for translations, which currently has [23 different](https://github.com/thewca/tnoodle/tree/master/webscrambles/src/i18n) languages.

wca-regulations.md
lgarron: I *think* specifying this not particularly useful. A competitor who wants to get a valid result will write a sufficiently clear name, and I don't see this as a useful avenue for cheating. Open to suggestions, though!

wca-guidelines.md
Laura-O: I still don't like this and I think this will be interpreted inconsistently.
What is a "small typographic correction"? What if I made a mistake when copying my solution and accidentally skipped 4 moves? Is this still small?
And how can the judge make sure if it's an insertion or not (especially relevant when the insertion is a sexy move or something like `R U2 R'`)?

lgarron (commented):
> [perhaps this isn't the place?] I'm not particularly crazy about how we allow movements that are not listed like Rw. Explaining notations seems way too unnecessary. I think we should give some examples: F, F2, F', Fw, but imo there's no reason to list all the 18 face moves and all 18 cube rotations.

I think I agree with you, but I'm not sure what you are saying. Are you asking for anything to change in/based on this PR?

wca-guidelines.md
lgarron: An insertion is fine if it's in-line, and I think most typographic corrections are clear in practice.

Do you have alternative suggestions for how to express that the competitor is responsible for writing a single move sequence in order, while still being lenient to the examples you gave in the other draft PR?

kingmathyall (commented):
1. I don't think E2a1 is necessary - that kind of is just what you'd do regardless of this reg in place or not?
2. E2c1+++ - I think if they are able to find out who did the attempt, it shouldn't be DNFed so I'm not for this change
3. E3c4+ I would like to see this added somewhere on an FMC sheet if possible - letting everyone know what moves aren't allowed - this would take up a significant amount of space so I understand this might not be the best idea :/

wca-guidelines.md
```diff
@@ -205,10 +205,16 @@ To be more informative, each Guideline is classified using one of the following
 
 ## <article-E><fewest-moves><fewestmovessolving> Article E: Fewest Moves Solving
 
+- E2a1+) [RECOMMENDATION] If a judge finds a submitted solution that does not appear to have information to identify the attempt, they should check both sides of the paper.
```
kingmathyall, line 208: I don't think E2a1 is necessary - that kind of is just what you'd do regardless of this reg in place or not?

wca-guidelines.md
```diff
@@ -205,10 +205,16 @@ To be more informative, each Guideline is classified using one of the following
...
+- E2c1++) [EXPLANATION] The competitor is not required to submit a paper with the competition name, round, and/or attempt number, but may wish to do so in order to help the organization team keep track of submitted solutions.
+- E2c1+++) [ADDITION] If a judge finds a submitted solution without information to identify the attempt (e.g. while grading the attempt), they should find out the competitor who wrote the solution in order to assign them a DNF (rather than DNS). Note that the competitor must not be allowed to write information to identify the attempt at this point, so they must receive a DNF result.
```
kingmathyall, line 213: E2c1+++ - I think if they are able to find out who did the attempt, it shouldn't be DNFed so I'm not for this change

kingmathyall (dismissed):

wca-guidelines.md
lgarron: I think we need to be explicit about what is allowed to avoid inconsistent rulings from Delegates. The incidents from this year suggest that not every Delegate is on the same page about this right now. :-(

wca-guidelines.md
lgarron: I think we need to be explicit about what is allowed to avoid inconsistent rulings from Delegates. The incidents from this year suggest that not every Delegate is on the same page about this right now.

@AlbertoPdRF, would you be in favor of allowing the result to count at the discretion of the Delegate?

lgarron:
> 3\. E3c4+ I would like to see this added somewhere on an FMC sheet if possible - letting everyone know what moves aren't allowed - this would take up a significant amount of space so I understand this might not be the best idea :/

My goal after the draft Regs are out are to work on guides, one of which would be a page with FMC grading instructions. I don't think we can fit it on the FMC sheet, but my hope is that competitors can be aware of the grading sheet and look at it if they want to be careful.

wca-guidelines.md
kingmathyall: If you can find out who did the attempt, you know who did the attempt, you have their attempt - it shouldn't be a DNF. That's just being way too strict and I think it's unnecessary to be that strict in FMC

lgarron: Hmm, I just double-check and we had a unanimous WRC consensus for "Can be added in front of judge while submitting" (as opposed to the explicit alternative "Can be added until the end of the competition"): https://github.com/thewca/wca-regulations/issues/655#issuecomment-443497338

Would you be alright sticking with that, or is it possible some of us didn't take into account this case while voting?

Laura-O (approved):

Laura-O:
> @Laura-O, could you review again? :-D

Sorry, completely missed that this is still open.

lgarron:
> > @Laura-O, could you review again? :-D
> 
> Sorry, completely missed that this is still open.

Thanks!

wca-guidelines.md
kingmathyall: I'd want it clarified that if the delegate can determine whose attempt it was to put the DNF in, why not put in the time? I guess when voting I completely overlooked that option. It just doesn't make any sense to me that if fit can be determined who did the attempt, that it gets entered as a DNF. If someone leaves and they can't confirm with the competitor that it is indeed their attempt, then sure, DNF the attempt, but if you can confirm it was there, I think it should be added.

AlbertoPdRF: Allowing this at the discretion of the WCA Delagate shouldn't be an option in my opinion, as we would still have the same inconsistent rulings.

lgarron:
This has been dangling for a while, and there are at least 3 additional changes I want to consider. In order to avoid complicating this, I'm going to dismiss @kingmathyall's review on this PR and make separate PRs for easier review for those parts.

jamesquinn1:
Could clarification be given on if people are allowed to write insertion stickers ahead of time? It gives a potential way to cheat by writing out algorithms or (in the worst case scenario) someone could have gotten the scramble ahead of time and written the solution out on stickers. I could see a solution to this problem being delegates have to check beforehand, but in a large scale competition that doesn't seem viable.

lgarron:
> Could clarification be given on if people are allowed to write insertion stickers ahead of time?

#584 covers this. Unless we want to enforce draconian measures around the world, we need to trust competitors quite a bit for FMC. If you have any suggestions for how to restrict this, feel free to comment there.

PR #679, opened 2018-12-21
Allow competitors to touch stopwatches. Closes #581

### LINKED ISSUE #581: Clarify whether competitors are allowed to touch or stop stopwatch timers.
From https://github.com/thewca/wca-regulations/issues/567#issuecomment-392616988

> I think there are two perspectives here: A) we have to trust the competitor a little anyhow, so let them interact with the stopwatch, and B) we can draw an easily enforceable line by saying that the competitor is not allowed to touch the stopwatch, but they should be able to find out the current stopwatch time if they want. (A strawman idea: they have the right to request that the judge place the stopwatch in such a way that the running time is visible to them, else they may ask the judge for the time whenever during the solve they want).

I still think the strawman works:
- The competitor is never allowed to touch the stopwatch.
- If the stopwatch face is not visible to the competitor (when they don't have a blindfold on), they may ask the judge to move it so that they can see the time.

lgarron:
This also has a provision for the organization team to provide the competitor an unofficial stopwatch instead.

I thought we landed this change, but apparently it was lost. 😳


kingmathyall (commented):
LGTM!

PR #684, opened 2018-12-24
Update A4d

pedrosino:
See discussion over at #573

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -342,7 +342,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 ## <article-A><speedsolving><speedsolving> Article A: Speed Solving
 
 - A1) Speed Solving attempts must abide by the following procedures.
-    - A1a) The organization team applies exactly one kind of time limit (time limit per attempt or cumulative time limit) to each round.
+    - A1a) The organization must apply exactly one kind of time limit (time limit per attempt or cumulative time limit) to each round.
         - A1a1) The default time limit is 10 minutes for each attempt in a round, though the organization team may announce a higher or lower time limit per attempt. In this case, the limit must be the same for all attempts in a round.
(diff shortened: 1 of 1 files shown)


lgarron (commented):
Thanks! One important consistency question, though.

xsrvmy:
So if I understand correctly, the competitor only get one chance to attempt to start the timer? I've actually had occasional issues with the red light not even lighting up and having to retry.

pedrosino:
> Thanks! One important consistency question, though.

I didn't understand. What question?

wca-guidelines.md
```diff
@@ -165,6 +165,7 @@ To be more informative, each Guideline is classified using one of the following
 - A3c3++) [CLARIFICATION] The competitor cannot "test" or "restart" the timer during inspection, since inspection ends once the competitor starts the solve (see [Regulation A4d](regulations:regulation:A4d)). If the competitor stops the timer after they have started it for the first time, this is considered stopping the solve (see [Regulation A6](regulations:regulation:A6))), even if they do so within 15 seconds of starting inspection.
 - A3c4+) [CLARIFICATION] If the use of a thin object inside the puzzle has been enforced, the judge must not remove the object on behalf of the competitor.
 - A3c4++) [CLARIFICATION] The competitor may remove the object during the solve, if they did not do it during the inspection (e.g. they forgot).
+- A4d+) [CLARIFICATION] If the competitor started the solve without waiting long enough for the green light to appear, the solve is disqualified (DNF).
```
lgarron: The competitor can only start the solve by starting the timer, so this doesn't really make sense. Could you rephrase this to describe what the competitor actually can't do (e.g. perform moves on the puzzles)?

lgarron:
> I didn't understand. What question?

Sorry, that got lost. Added it back.

> So if I understand correctly, the competitor only get one chance to attempt to start the timer?

Hmm, how are you drawing that conclusion?
(The wording seems clear to me, but perhaps I'm glossing over something.)

xsrvmy:
> Hmm, how are you drawing that conclusion?
(The wording seems clear to me, but perhaps I'm glossing over something.)

It's kind of the same question you commented. I think we need to clarify that the solve is started only if the timer is started and not just when the hand is released.

wca-guidelines.md
pedrosino: That's why I propose to change to 
> They start the solve by removing their hands from the timer.

Once you remove your hands - having started the timer or not - you can't go back and "re-start", because the judge is supposed to have stopped the stopwatch.

xsrvmy: What happens if the red light fails to turn on?

Alternative proposal: The competitor can retry starting the timer as long as the judge did not call 12 yet.

pedrosino:
Closing this because I messed up the branches.

PR #688, opened 2019-01-13
Allow the competitor to write their registrant ID for FMC.

lgarron:
This is important to specify because otherwise writing the registrant ID before
the attempt (E2a1) is a DNF.

See https://github.com/thewca/tnoodle/issues/379 for a TNoodle issue about
adding a registrant ID field to TNoodle FMC sheets.

@Baiqiang, @Laura-O, @jfly, would you mind providing input on this?

The WCIF specifies "registrant ID": https://docs.google.com/document/d/1hnzAZizTH0XyGkSYe-PxFL5xpKVWl_cvSdTzlT_kAs8/edit#bookmark=id.io26im5pym1q
https://github.com/thewca/tnoodle/issues/379 uses "competitor number".

I think we should try to standardize on one name, else we end up in a situation where some people call it one and some call it the other (see ["combined round" vs. "cutoff round"/"round with a cutoff"](https://github.com/thewca/wca-regulations/issues/543)).


jfly:
As we discussed offline, the WCIF definitely needs some more explanation. I don't feel strongly about "registrant id" vs "competitor id", so I'm happy to go with whatever people prefer.

I think "registrant id" would be slightly less work because that's what the WCIF already specifies.

lgarron:
> I think "registrant id" would be slightly less work because that's what the WCIF already specifies.

After thinking about it for a bit, I think I prefer "competitor number", since:
- "Registrant" is a slightly weird word, and the phrase should be as easy as possible to understand if we put it somewhere like in the Regs and on the FMC sheet.
- Using "ID" in the name is more likely to cause accidental confusion with "WCA ID".
- Specifying that it must be a number will make it easier to handle consistently, and sets a simple, accurate expectation for anyone interacting with the concept. (Then again, it cuts down on innovation if e.g. competitor code phrases or alphanumeric identifiers turn out to be a better idea.)

How hard would it be to update the WCIF?

jfly:
One reason I prefer the word "registrant" to "competitor" is because "competitor" is only a small number of pixels away from the word "competition".

@jonatanklosko and @ronaldmansveld to answer the question about how hard it would be to change `registrantId` to something else.

Baiqiang:
This looks good to me. I prefer competitor number too.

jonatanklosko:
@jfly answering your question, it shouldn't be too difficult, as far as I can see it's just updating the WCIF spec and WCIF rendering on the WCA website. Would also mean changing current wca-live and groupifier-next code, but that's a single replace anyway. It will be harder to do when more apps rely on WCIF, so we should decide on the good name at this point.

I don't feel strongly about the exact name either, all mentioned above sound good.
One additional subtlety: `registrantId` is a number for every person with registration (even if its status is pending/deleted), whereas `competitorNumber`/`competitorId` sounds a bit like an id of someone who actually competes. Every name has its pros and cons ^^

jfly:
It would be nice to avoid using the word "number" for the reason @lgarron mentioned about "cutting down on innovation if e.g. competitor code phrases or alphanumeric identifiers turn out to be a better idea". Maybe "competitor code"?

lgarron:
I still can't think of something strictly better than "registrant ID", so I think we might as well stick with prior art.

kingmathyall (approved):
I still don't completely agree with DNFing the result if they forget their information, but we are still able to identify them. If we know who did the attempt to only give them a DNF, I don't see what's wrong with giving them their result at that point.

lgarron:
> I still don't completely agree with DNFing the result if they forget their information, but we are still able to identify them. If we know who did the attempt to only give them a DNF, I don't see what's wrong with giving them their result at that point.

Yeah, that's a reasonable stance, but we do have a majority of WRC members against that detail. 🤷‍♀️

campos20:
Ping. So, did we came to a conclusion on this?

lgarron:
> Ping. So, did we came to a conclusion on this?

Could you specify a bit more? 229bf8e73859d1e49edcf028556b984de81f1ba9 landed in the draft Regs.

PR #689, opened 2019-01-13
Specify a penalty in E2a1.

lgarron:
The default penalty is already DNF, but spelling it out is better.

@kingmathyall, could you review?


AlbertoPdRF (approved):
LGTM!

PR #691, opened 2019-01-13
Specify that competitors may join FMC late, if no one has left.

lgarron:
@Claster, could you review?


wca-guidelines.md
```diff
@@ -201,6 +201,7 @@ To be more informative, each Guideline is classified using one of the following
 
 - E2a1+) [RECOMMENDATION] If a judge finds a submitted solution that does not appear to have information to identify the attempt, they should check both sides of the paper.
 - E2b+) [CLARIFICATION] A competitor may choose to stop an attempt early by handing in a solution before the time limit.
+- E2b++) [ADDITION] A competitor may join a Fewest Moves attempt that is aleready in progress, at the discretion of the Delegate. In this case, their time limit is the time remaining (i.e. they must finish when other competitors reach their 60 minute time limit). Exception: a competitor must not be permitted to join a Fewest Moves attempt if any competitor has finished their attempt already (see [Guideline E2b+](guidelines:guideline:E2b+)), or if there is any other risk that information about the scramble may have become known to the competitor who wishes to join.
```
```suggestion
- E2b++) [ADDITION] A competitor may join a 3x3x3 Fewest Moves attempt that is aleready in progress, at the discretion of the WCA Delegate. In this case, their time limit is the time remaining (i.e. they must finish when other competitors reach their 60 minute time limit). Exception: a competitor must not be permitted to join a 3x3x3 Fewest Moves attempt if any competitor has finished their attempt already (see [Guideline E2b+](guidelines:guideline:E2b+)), or if there is any other risk that information about the scramble may have become known to the competitor who wishes to join.
```

lgarron: Ah, good catch. I'm still thrown off by that, but your correction is the right choice for consistency!

AlbertoPdRF (approved):
LGTM now!

wca-guidelines.md
```diff
@@ -201,6 +201,7 @@ To be more informative, each Guideline is classified using one of the following
 
 - E2a1+) [RECOMMENDATION] If a judge finds a submitted solution that does not appear to have information to identify the attempt, they should check both sides of the paper.
 - E2b+) [CLARIFICATION] A competitor may choose to stop an attempt early by handing in a solution before the time limit.
+- E2b++) [ADDITION] A competitor may join a 3x3x3 Fewest Moves attempt that is aleready in progress, at the discretion of the WCA Delegate. In this case, their time limit is the time remaining (i.e. they must finish when other competitors reach their 60 minute time limit). Exception: a competitor must not be permitted to join a 3x3x3 Fewest Moves attempt if any competitor has finished their attempt already (see [Guideline E2b+](guidelines:guideline:E2b+)), or if there is any other risk that information about the scramble may have become known to the competitor who wishes to join.
```
andreaspung, line 204: Should be "already".

Claster:
LGTM besides "aleready".

wca-guidelines.md
lgarron: Hmm, I thought I landed the "aleready" fix, but looks like I missed it. Will fix directly on `draft`.

HippolyteM:
Hi everyone,
Sorry, I may have missed a previous discussion. I didn't find it.
Why do we want to specify that? Is it so unfair to prevent competitors that cannot read a schedule to join fmc? Someone joining late might be disturbing for other competitors, and it takes time and attention from the delegate.
Has it been discussed?
Also, the fact that it's at the discretion of the delegate might create some unfairness between competitors from different areas. If we think it's ok to come whenever you want as long as nobody left (which I don't), why would delegates be able to forbid it just because they don't like it?

lgarron:
Some Delegates *do* currently allow competitors to join late, so it's a codification of existing practice. I believe that forbidding competitors from joining late would be a larger change, which we could discuss in a new issue.

> Also, the fact that it's at the discretion of the delegate might create some unfairness between competitors from different areas. If we think it's ok to come whenever you want as long as nobody left (which I don't), why would delegates be able to forbid it just because they don't like it?

Because someone has to make the call on whether there was a risk of the competitor having information before joining.

HippolyteM:
I don't really get your point.

> Some Delegates do currently allow competitors to join late

So you recognize that some don't. If we follow your own reasoning, forbidding joining late would just be a codification of existing practice.
Delegates currently do or don't allow joining late, I don't see why one of the two solutions would be treated as the default one (and thus codified in the regulation without further thinking about which is best).

> Because someone has to make the call on whether there was a risk of the competitor having information before joining.

The current wording of your proposal doesn't say that this decision (the discretion of the delegate) have to be based upon that risk. It just says that if there is a risk, joining is forbidden. But it totally allows a delegate to refuse someone joining even if there's no risk.

lgarron:
> So you recognize that some don't. If we follow your own reasoning, forbidding joining late would just be a codification of existing practice.

The existing practice is that Delegates do this at their discretion.

lgarron:
> The current wording of your proposal doesn't say that this decision (the discretion of the delegate) have to be based upon that risk. It just says that if there is a risk, joining is forbidden. But it totally allows a delegate to refuse someone joining even if there's no risk.

Ah, I see.

Well, someone still needs to judge that there was no risk, since that's hard to write down objectively.

There are also some other reasons it's reasonable to refuse a competitor to join late. I can think of:
- Applying [2u](https://www.worldcubeassociation.org/regulations/#2u) strictly across an entire competition.
- Deciding that a competitor joining with 5 minutes remaining in the attempt is not reasonable.
- All judges are competing in the event, so it would be disruptive to let someone join late.

*Requiring* Delegates to take late-comers would also not reflect the status quo.

PR #690, opened 2019-01-13
Restore Multi-Blind and FMC combined rounds.

lgarron:
These formats were used at e.g. Euro 2014 (combined multi) and Hungarian Open 2018 (combined FMC). Based on that precedent, the removal of this
format in earlier 2019 Regs drafts was an accident.

@AlbertoPdRF, could you review?

PR #692, opened 2019-01-13
Clarify that 9b specifies the valid combined rounds.

lgarron:
After #690 (Restore Multi-Blind combined rounds), this should make it unambiguous which combined round formats are allowed for which event.


AlbertoPdRF (approved):
LGTM!

Issue #700, opened 2019-01-31
Explicitly allow competitors to use hand warmers / drinks bottles


Mollerz:
As discussed in a few recent reports, the WRC has allowed the use of hand warmers during an attempt. Does this also include electronic hand warmers?

Also discussed is that it is quite common for competitors to want to have a drinks bottle with them during longer events, e.g. 5BLD, MBLD, FMC. I think we should be allowing any non-alcoholic beverage to be consumed during an attempt as opposed to just water. Consider a diabetic that might need to drink a sugary beverage to increase blood-sugar level.

Both of these should append to A5b most likely.

lgarron:
I think this is pretty reasonable.

> As discussed in a few recent reports, the WRC has allowed the use of hand warmers during an attempt. Does this also include electronic hand warmers?

I don't think we have full consensus on that yet, but now's a good time to discuss.
I don't see a problem with electric hand warmers before the attempt begins, but I think we should keep heavy restrictions on electronic interaction during an attempt. So, I'd vote for only non-electric hand warmers during an event.

Drink labels shouldn't be used to cheat (e.g. have algs encoded on them), but I can't think of anything clever. I think it's fine to just allow a moderate amount of food and drink, at the discretion of the Delegate.  (There are situations where the Delegate would need to be restrictive, e.g. if the venue strictly disallows foods or beverages.)

Would you be interested in drafting a change for this?

lgarron:
Alright, I think this is fairly solved for now.

Discretion of the Delegate can handle misc. edge cases.

PR #701, opened 2019-01-31
Explicitly allow non-electronic hand warmers and food&drink

Mollerz:
Instead of adding this to A5b it makes more sense to add them to 2i1. This parent regulation states already that it is at the discretion of the WCA delegate so it does not need to be included within 2i1c and 2i1d which are being added.

See: https://github.com/thewca/wca-regulations/issues/700

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -83,6 +83,8 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 2i1) Competitors may use non-electronic aids that do not give an unfair advantage, at the discretion of the WCA Delegate. This includes:
         - 2i1a) Medical/physical aids worn by the competitor (e.g. glasses, wrist brace). As an exception to [Regulation 2i](regulations:regulation:2i), medical aids may be electronic if the competitor does not have comfortable non-electronic alternatives (e.g. if the competitor has a personal hearing aid or pacemaker).
         - 2i1b) Earplugs and earmuffs (but not electronic headphones and earbuds).
+        - 2i1c) Hand warmers.
+        - 2i1d) Food and drink.
     - 2i2) Competitors may use cameras at the solving station at the discretion of the WCA Delegate, but the following restrictions apply from the start of the attempt until the competitor stops the solve. Penalty for breaking a restriction: disqualification of the attempt (DNF).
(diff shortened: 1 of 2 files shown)


Jambrose777:
What about a heated jacket or light up shirt? It’s electronic but part of the clothing they are wearing.

Jambrose777:
Sorry, completely wrong place to leave this comment. I meant to comment on the issue :(

wca-regulations.md
```diff
@@ -83,6 +83,8 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 2i1) Competitors may use non-electronic aids that do not give an unfair advantage, at the discretion of the WCA Delegate. This includes:
         - 2i1a) Medical/physical aids worn by the competitor (e.g. glasses, wrist brace). As an exception to [Regulation 2i](regulations:regulation:2i), medical aids may be electronic if the competitor does not have comfortable non-electronic alternatives (e.g. if the competitor has a personal hearing aid or pacemaker).
         - 2i1b) Earplugs and earmuffs (but not electronic headphones and earbuds).
+        - 2i1c) Hand warmers.
```
lgarron, line 86: Would you mind specifying "non-electronic" here?

Mollerz: I don't see why? The parent regulation clearly states:

>    - 2i1) Competitors may use non-electronic aids that do not give an unfair advantage, at the discretion of the WCA Delegate. This includes:

lgarron: Yeah, I kind of wanted to be more clear.
The "While competing" of 2i is becoming steadily less sufficient.

How about adding a Guideline instead, to the effect of "2i1c+ [CLARIFICATION] Electronic hand warmers may be used before or after an attempt. Non-electronic hand warmers may be used during an attempt."

Ideally, I'd like to align this with the camera Regs (start of attempt to end of solve, so that interaction is allowed before signing the score card). I don't have time to think through that change right now, though.

Mollerz: Added the clarification and made slightly more specific https://github.com/thewca/wca-regulations/pull/701/commits/ce3173a1699c409daba1fbab4d93d37ba6cdce39

lgarron (approved):

PR #706, opened 2019-02-20
FMC signatures

Ivan-Brigidano:
Become mandatary to have three signatures: the competitor's one, the judge's (usually a WCA Delegate) and the solution checker's one.


xsrvmy:
I think this might cause scheduling issues and also forces the competitor to stay longer at the venue after the attempt

lgarron:
Hi @Ivan-Brigidano, I don't see any commits in this range, so I'm guessing you just meant to suggest this as a normal issue. I'll break this into a separate issue to avoid confusion.

Ivan-Brigidano:
Thanks you @lgarron  I'm new in Git so I made this little mistake, sorry.
The idea is to have signatures to make it more professional and have the certainty of who has corrected, judged... . I personally don't think it would suppose many time delays.
However, I don't compete in FMC and maybe I am suggesting something unnecessary.

PR #672, opened 2018-12-15
Update event names throughout text to match official event names.

AlbertoPdRF:
This PR is #628 with conflicts fixed, all commits squashed, and adding to the commit message that it fixes #582.

I would appreciate a quick review so we can solve all remaining inconsistencies and merge it up! Can you take a look @micahstairs?

PR #557, opened 2018-03-23
Reword A6g and remove A6g+

pedrosino:
Fixes #556

PR #709, opened 2019-02-22
Fixes mistakes noticed by Judith Laister

viroulep:
Two main errors we overlooked:
  - #557 actually removed the exception for Clock in A6g, which makes it impossible for the judge to check if a Clock is solved or not.
  - the standardization of event names in #672 and #628 lead to "Blindfolded Solving" being replaced by "3x3x3 Blindfolded" B1+, when it's actually true for any blindfolded event.

### DIFF
--- a/style-guide.md
+++ b/style-guide.md
@@ -96,7 +96,7 @@ This document is meant to clarify some of the implicit conventions adopted in WC
 - "Each" (individual).
 - "Any" (potential, i.e. in case of applicability).
 - "Submit" instead of "hand in" or "turn in".
-- "Scramble program" of "scramble sequences" instead of "scrambling program" or "scrambling sequence".
+- "Scramble program" or "scramble sequences" instead of "scrambling program" or "scrambling sequence".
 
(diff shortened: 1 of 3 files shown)


xsrvmy:
Why is the exception in A6g still needed for clock? That commit says that the judge must not touch the puzzle when resolving a misalignment and there is no such thing on clock.

viroulep:
Because usually competitors do not bother putting back the clock standing up, and you still have to check both sides for any misaligned pin (that results in a DNF).

sam596:
@viroulep I think what Max is trying to say is that A6g applies whenever there is ambiguity regarding a near 45 degree (for instance, on 3x3) ending state, whereas Clock doesn't have +2 misalignment penalties.

But I still agree putting that exception in there to be clear. It could be argued that a single dial off by 1 o'clock could be described as a misalignment.

lgarron:
> It could be argued that a single dial off by 1 o'clock could be described as a misalignment.

After thinking about this a bunch, I'l totally fine with this interpretation.

PR #717, opened 2019-03-26
Require permission for the competitor to test the timer at the start of a BLD attempt.

lgarron:
I've thought about this a whole bunch, and I couldn't think of a really good way to specify that the competitor can ask to test the timer (see https://github.com/thewca/wca-regulations/commit/c98b5f7e36451824ab6f07a5f74e83c7672851f3 for a simple attempt). The best compromise I could think of is to allow the *judge* to do this, since this:
- is hard to abuse,
- is unlikely to lead to ambiguity, and
- solves the situation for the only valid use case we can think of (making sure that the timer will actually work for the attempt).

@cubewhiz, what do you think of this revision?
@thewca/wrc-team, could you review?

### DIFF
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -183,7 +183,8 @@ To be more informative, each Guideline is classified using one of the following 
 
 - B1+) [REMINDER] The competitor must use a puzzle without textures, markings, or other features that distinguish similar pieces (see [Regulation 3k](regulations:regulation:3k)). This should be given special attention for Blindfolded Solving.
 - B1b+) [RECOMMENDATION] Blindfolds should be checked by the WCA Delegate before use in the competition.
-- B2d+) [ADDITION] Since the competitor normally starts the solve by starting the timer, the competitor must not "test" or "restart" the timer once the judge has indicated that they are ready (see [Regulation B2a](regulations:regulation:B2a)). Exception: the competitor may test (start and stop/reset) the timer once if they first ask the judge.
+- B2d+) [ADDITION] Since the competitor starts the solve by starting the timer, the competitor must not "test" or "restart" the timer once the judge has indicated that they are ready (see [Regulation B2a](regulations:regulation:B2a)). However, they may ask the judge to start and reset a timer in order to verify that it is working.
+
 
 ## <article-C><one-handed><onehandedsolving> Article C: One-Handed Solving
 
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -183,7 +183,7 @@ To be more informative, each Guideline is classified using one of the following 
 
 - B1+) [REMINDER] The competitor must use a puzzle without textures, markings, or other features that distinguish similar pieces (see [Regulation 3k](regulations:regulation:3k)). This should be given special attention for Blindfolded Solving.
 - B1b+) [RECOMMENDATION] Blindfolds should be checked by the WCA Delegate before use in the competition.
-- B2d+) [ADDITION] By default, the competitor starts the solve the first time that they start the timer after the judge has indicated that they are ready (see [Regulation B2a](regulations:regulation:B2a)). If they want to check that the timer is in working order, they must ask (and receive an acknowledgement from) the judge before starting/resetting the timer during this phase. Penalty for starting and resetting a timer without asking the judge: disqualification of the attempt (DNF).
+- B2d+) [ADDITION] By default, the competitor starts the solve the first time that they start the timer after the judge has indicated that they are ready (see [Regulation B2a](regulations:regulation:B2a)). If they want to check that the timer is in working order, they must (ask and) receive confirmation from the judge each time before starting/resetting the timer during this phase. Penalty for starting and resetting a timer without confirmation from the judge: disqualification of the attempt (DNF).
 
 
 ## <article-C><one-handed><onehandedsolving> Article C: One-Handed Solving
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -183,7 +183,7 @@ To be more informative, each Guideline is classified using one of the following 
 
 - B1+) [REMINDER] The competitor must use a puzzle without textures, markings, or other features that distinguish similar pieces (see [Regulation 3k](regulations:regulation:3k)). This should be given special attention for Blindfolded Solving.
 - B1b+) [RECOMMENDATION] Blindfolds should be checked by the WCA Delegate before use in the competition.
-
+- B2d+) [ADDITION] Since the competitor normally starts the solve by starting the timer, the competitor must not "test" or "restart" the timer once the judge has indicated that they are ready (see [Regulation B2a](regulations:regulation:B2a)). Exception: the competitor may test (start and stop/reset) the timer once if they first ask the judge.
 
 ## <article-C><one-handed><onehandedsolving> Article C: One-Handed Solving
 
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -183,7 +183,7 @@ To be more informative, each Guideline is classified using one of the following 
 
 - B1+) [REMINDER] The competitor must use a puzzle without textures, markings, or other features that distinguish similar pieces (see [Regulation 3k](regulations:regulation:3k)). This should be given special attention for Blindfolded Solving.
 - B1b+) [RECOMMENDATION] Blindfolds should be checked by the WCA Delegate before use in the competition.
-- B2d+) [ADDITION] Since the competitor starts the solve by starting the timer, the competitor must not "test" or "restart" the timer once the judge has indicated that they are ready (see [Regulation B2a](regulations:regulation:B2a)). However, they may ask the judge to start and reset a timer in order to verify that it is working.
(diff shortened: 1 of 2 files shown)


Jambrose777:
I’d prefer to list specifically “the judge says READY” instead of saying “indicating they are ready”. Otherwise it kinda leaves the door open for ambiguity and is unfair for the competitor.

lgarron:
> I’d prefer to list specifically “the judge says READY” instead of saying “indicating they are ready”. Otherwise it kinda leaves the door open for ambiguity and is unfair for the competitor.

Do you think they should say "READY" even when the competitor is using strong earmuffs?

Jambrose777:
I say this directly following regulation 2ba’s wording. They should at least be the same.

I personally feel the judge shouldn’t have to indicate they are ready in any way as it goes against being quiet and can disrupt the competitors concentration pre attempt. Plus the repurcussuons for not waiting for the judge to be ready are not defined anywhere, which is a bit worrying.

Jambrose777:
And then to use your own argument: I can see a lot of cases that may happen where: competitor tests the timer, judge dnfs the attempt, the competitor says they never heard the judge say ready. What would we do then?

lgarron:
> They should at least be the same.

Hmm, could you explain why it's more important to keep them the same, compared to the consideration that audio cues are less likely to work for BLD?

> I personally feel the judge shouldn’t have to indicate they are ready in any way as it goes against being quiet and can disrupt the competitors concentration pre attempt.

Well, that's what placing the puzzles or thumbs-up is for, no?

> Plus the repurcussuons for not waiting for the judge to be ready are not defined anywhere, which is a bit worrying.

If you don't follow the procedure, you'll end up with either a DNS or a DNF, depending on what you do. Could you describe a particular situation you think is not covered?

> And then to use your own argument: I can see a lot of cases that may happen where: competitor tests the timer, judge dnfs the attempt, the competitor says they never heard the judge say ready. What would we do then?

Hmm, I sort of see your point. But:
- I don't really see that as likely to happen, given that the main point of this Reg is "don't test the timer yourself!". I'd expect every moderately experienced BLD solver to be careful.
- How would you modify this so that it's clear to the competitor when they are allowed to start? For certain situations (e.g. multi BLD) both they and the judge may be situated at the table significantly before the judge is ready.

An easy approach would be to specify that the competitor must never test the timer while sitting at a BLD station, but that fix is orthogonal to communicating readiness.

Jambrose777:
> > They should at least be the same.
> 
> Hmm, could you explain why it's more important to keep them the same, compared to the consideration that audio cues are less likely to work for BLD?

Because regulation consistency is important. This regulation refers to 2ba where “the judge indicates they are ready” is defined with “the judge says they’d Re already”. If we want the judge to instead say they need to indicate that they dare ready, then regulation b2a needs to be change to this wording and further define indicating they are ready in an additional guideline.

> 
> > I personally feel the judge shouldn’t have to indicate they are ready in any way as it goes against being quiet and can disrupt the competitors concentration pre attempt.
> 
> Well, that's what placing the puzzles or thumbs-up is for, no?
> 
> > Plus the repurcussuons for not waiting for the judge to be ready are not defined anywhere, which is a bit worrying.
> 
> If you don't follow the procedure, you'll end up with either a DNS or a DNF, depending on what you do. Could you describe a particular situation you think is not covered?

I’m not okay with a regulation being ambiguously define with such a harsh repercussion. This only opens the door to WRC rulings that have potential of having inconsistent outcomes. Explicitly defining everything about this in he regulations to make it as clear as possible for the competitors to know when that they must wait for the judge to be ready and how they know the judge is ready is the only way to maintain competitor fairness with this.

> 
> > And then to use your own argument: I can see a lot of cases that may happen where: competitor tests the timer, judge dnfs the attempt, the competitor says they never heard the judge say ready. What would we do then?
> 
> Hmm, I sort of see your point. But:
> 
> * I don't really see that as likely to happen, given that the main point of this Reg is "don't test the timer yourself!". I'd expect every moderately experienced BLD solver to be careful.
> * How would you modify this so that it's clear to the competitor when they are allowed to start? For certain situations (e.g. multi BLD) both they and the judge may be situated at the table significantly before the judge is ready.
> 

I’m not concerned about moderately experienced bld competitors. The regulations apply to everyone not just that specific subgroup of competitors.

See above: I’d define the regulations to make it absolutely clear (across all regulations needed, so 2ba and a possible additional guideline) what a judge can do to indicate they are ready

> An easy approach would be to specify that the competitor must never test the timer while sitting at a BLD station, but that fix is orthogonal to communicating readiness.

That’s just harsh for no apparent reason.

xsrvmy:
I can see a problem with banningtimer tests: what is a competitor is usingthe timer to time a practice attempt when the judge/runner brings in a puzzle?

cubewhiz:
> I can see a problem with banningtimer tests: what is a competitor is usingthe timer to time a practice attempt when the judge/runner brings in a puzzle?

Why would they be doing this? Does this happen at competitions you attend?

cubewhiz:
I think it is tricky to get the wording of this one just right, but I vastly prefer the new wording to the old wording.

xsrvmy:
> > I can see a problem with banningtimer tests: what is a competitor is usingthe timer to time a practice attempt when the judge/runner brings in a puzzle?
> 
> Why would they be doing this? Does this happen at competitions you attend?

I haven't seen this during BLD, but this happens in stationary finals a lot. It's not a problem with speedsolving though because the timer does not start the attempt in speedsolving.

In terms of requiring the competitor to wait for the judge to be ready, I think we need to be a bit careful not to end up with a similar situation to missing names in FMC where a lot of competitors will be DNFed for something that happens frequently by accident but does not affect the integrity of the solve. We do still need to consider the possibility that the judge might miss the start of the solve and thus not start the stopwatch or  not place the sightblocker. If the judge is unable to catch a stopwatch start that should definitely be a DNF. The sightblocker problem should probably be discussed in the other issue.

lgarron:
@Jambrose777: if I understand correctly, your objection is that we don't unambiguously specify a way for the competitor to know that the judge indicated readiness.

Would it address your concern if we turned the three example signals into the only three allowed signals?

I don't really like the idea of overspecifying this, given that A3b1, A3d, and A7a (all similar points of communication) have a wide variety of interpretations. But I'm not particularly opposed to it.

lgarron:
I don't want to be too dismissive of @Jambrose777's concerns, but I think the current wording is about as good a compromise I can make between something strict and prescriptive and something that resembles what happens in practice.

@thewca/wrc-team, could one of you review this?

viroulep:
I understand this change is about consistency with the change in #579 (which bans testing timer during inspection), but it seems to address drastically an issue that occurred only once (AFAIK) during CubingUSA's Nationals 2018 (incident 27), where a competitor tested the timer before the attempt.

This change will make possible for competitors to test the timers before the attempt in events with inspection but not for blindfolded events.
I don't think testing the timer before blindfolded attempts is an issue. I did see it a couple of time and it's usually clear what the intent of the competitor is.

Clarifying that the competitor must communicate clearly to the judge that he's just testing the timer sounds fine though, and specifying that an ambiguous start will be considered as starting the attempt sounds fine too (meaning that ambiguously testing the timer would result in a DNF).

lgarron:
> Clarifying that the competitor must communicate clearly to the judge that he's just testing the timer sounds fine though, and specifying that an ambiguous start will be considered as starting the attempt sounds fine too (meaning that ambiguously testing the timer would result in a DNF).

Okay,  I've updated B2d+ to allow competitors to do the timer testing themselves. Could you review again?

wca-guidelines.md
```diff
@@ -183,6 +183,7 @@ To be more informative, each Guideline is classified using one of the following
 
 - B1+) [REMINDER] The competitor must use a puzzle without textures, markings, or other features that distinguish similar pieces (see [Regulation 3k](regulations:regulation:3k)). This should be given special attention for Blindfolded Solving.
 - B1b+) [RECOMMENDATION] Blindfolds should be checked by the WCA Delegate before use in the competition.
+- B2d+) [ADDITION] By default, the competitor starts the solve the first time that they start the timer after the judge has indicated that they are ready (see [Regulation B2a](regulations:regulation:B2a)). If they want to check that the timer is in working order, they must ask (and receive an acknowledgement from) the judge before starting/resetting the timer during this phase. Penalty for starting and resetting a timer without asking the judge: disqualification of the attempt (DNF).
```
lgarron: Per a suggestion by @Claster via Slack:

```suggestion
- B2d+) [ADDITION] By default, the competitor starts the solve the first time that they start the timer after the judge has indicated that they are ready (see [Regulation B2a](regulations:regulation:B2a)). If they want to check that the timer is in working order, they must (ask and) receive confirmation from the judge each time before starting/resetting the timer during this phase. Penalty for starting and resetting a timer without confirmation from the judge: disqualification of the attempt (DNF).
```

Claster (approved):

Iurig:
Is there an issue I don't see with allowing timer testing if they don't touch the cube cover while the timer is running? I don't see how that could be explored since any fumbling start I can think of involves the cube cover lift, not the timer starting itself. Perhaps even both at the same time (i.e.: They can test the timer if they either communicate it's a timer test or they don't touch the cube cover while the timer is running)

PR #731, opened 2019-05-01
Improve wording for E2c1++ and E2c1++++. Closes #727.

### LINKED ISSUE #727: New FMC regulation concerns
`E2c) At 60 minutes, each competitor must submit to the judge a single piece of paper with a written solution and information to identify the attempt.
E2c1+++) [EXPLANATION] The competitor is not required to submit a paper with the competition name, round, and/or attempt number, but may wish to do so in order to help the organization team keep track of submitted solutions.
`

These two regulation/guidelines are a contradiction to one another. Not only that, but this is encouraging cheating as a competitor who DNFs may now not turn in a solution in hopes to be awarded a DNS instead. This only increases the inaccuracy of our results and can become more of a burden on organizers. 

A solution that I use at comeptitions is to take attendance before he attempt to guarantee that we always award the correct DNF/DNS result for the competitors. This simply isn’t possible at competitions like Euros or US Nationals, where a competitor can easily sneak out and nobody ever realize that they complete the attempt.

`E2c1++) [ADDITION] If the competitor forgot to write information to identify the attempt on their intended solution paper before the time limit, they are permitted to add it in front of a judge. In this case, the competitor must ask the judge while submitting their solution and then immediately write information to identify the attempt (and make no other changes) under direct supervision of the judge. Once a competitor has finished submitting their solution (e.g. placing it in the judge's hands and allowing the judge to move on to the next competitor, placing it on the judge's table and walking away), they are not permitted to add any information. If the competitor writes anything after the time limit but before asking the judge, the attempt remains disqualified (DNF).`

The wording of this is very dependent on the comma and can be easily mistaken as “If the competitor forgot to write information to identify the attempt on their intended solution paper, before the time limit they are permitted to add it in front of a judge.” Which drastically changes the meaning. If we remove “before the time limit” the regulation maintains its original meaning and doesn’t cause any possible misinterpretations.

lgarron:
cc @Jambrose777


Jambrose777:
Ahhh this clears this up immensely, thank you!

PR #734, opened 2019-05-02
Update grammar in 2k3 and 2k4 so that it's easier to read.

lgarron:


wca-regulations.md
```diff
@@ -94,8 +94,8 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 2k1) Fails to check in or register in time for the competition.
     - 2k2) Is suspected of cheating or defrauding the officials during the competition.
         - 2k2a) The WCA Delegate may disqualify any suspected results.
-    - 2k3) Behaves in a way that is unlawful, violent or indecent; or intentionally damages venue facilities or personal property within the venue.
```
lgarron, line 97: We don't really use semicolons like this elsewhere, but we do use slashes to list alternatives sometimes. So I've used the latter to turn the semicolon into a comma.

hanwu85: I think this one need a English native user to confirm?

Ivan-Brigidano (approved):

wca-regulations.md
lgarron: @KatieHull, @Sydney-Weaver: does the new version (in green) look reasonable to you?

Ivan-Brigidano: @KatieHull @Sydney-Weaver ?

PR #735, opened 2019-05-03
All pyraminx colors are adjacent

tussosedan:
Removing redundant "adjacent" wording.


xsrvmy:
Not sure if this actually works. Yellow could technically be the darkest color, so a literal reading would mean yellow on bottom and yellow in front...

tussosedan:
Oh no, I think you're right. Dang.

PR #750, opened 2019-05-26
Solve instead of attempt

Goosly:


wca-guidelines.md
```diff
@@ -156,7 +156,7 @@ To be more informative, each Guideline is classified using one of the following
...
-- A1a2++) [EXAMPLE] If a round has a cumulative time limit of 30 minutes and the competitor's results first two attempts are 6:00 and DNF (10:00), then the time limit for the third attempt is 30:00 - 6:00 - 10:00 = 14:00. For this particular example, note that only the final result for an attempt counts. Therefore, a third attempt with an original recorded time of 13:59 without penalties meets the time limit because the final result is 13:59 (see [Regulation 9f2](regulations:regulation:9f2)), while a third attempt with a result of 13:59 + 2 = 14:01 does not meet the time limit.
+- A1a2++) [EXAMPLE] If a round has a cumulative time limit of 30 minutes and the competitor's first two results are 6:00 and DNF (10:00), then the time limit for the third attempt is 30:00 - 6:00 - 10:00 = 14:00. For this particular example, note that only the final result for an attempt counts. Therefore, a third attempt with an original recorded time of 13:59 without penalties meets the time limit because the final result is 13:59 (see [Regulation 9f2](regulations:regulation:9f2)), while a third attempt with a result of 13:59 + 2 = 14:01 does not meet the time limit.
```
Goosly: This should've been a separate PR but Github doesn't allow me to create branches (I think?). I changed "competitor's results first two attempts" to "competitor's first two results"

hanwu85: Hi, sorry about this late reply.  We want to keep the same terminology in the Regulations and the Guidelines.  So in this case, the "attempt" should be kept.  But this sentence do needs modification.  How about "....and the competitor's results **of the** first two attempts..."
Could you make the change?

wca-regulations.md
```diff
@@ -266,7 +266,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 ## <article-10><solved-state><solvedstate> Article 10: Solved State
 
 - 10b) Only the resting state of the puzzle, after the timer has been stopped, is considered.
-- 10c) The puzzle may be in any orientation at the end of the attempt.
+- 10c) The puzzle may be in any orientation at the end of the solve.
```
hanwu85, line 269: End of a solve is when you stop turning the puzzles.
End of a attempt is when you/a judge stop the timer.
They are slightly different.

wca-regulations.md
```diff
@@ -452,7 +452,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 - B5) Stopping the solve:
     - B5a) When using the Stackmat timer, the competitor stops the solve by releasing the puzzle and then stopping the timer.
     - B5b) When using a stopwatch, the competitor stops the solve by placing the puzzle back onto the surface and notifying the judge that they are stopping the solve. At that moment, the judge stops the timer.
-    - B5c) If the competitor is not touching the puzzle, they may remove the blindfold before they stop the timer. They must not touch the puzzle until the end of the attempt. Penalty for touching the puzzle: disqualification of the attempt (DNF).
+    - B5c) If the competitor is not touching the puzzle, they may remove the blindfold before they stop the timer. They must not touch the puzzle until the end of the solve. Penalty for touching the puzzle: disqualification of the attempt (DNF).
```
hanwu85: Same reason mentioned above.

hanwu85 (changes requested):
Please make a change according to the comment and PR again.  Thank you!

wca-regulations.md
Goosly: The end of the solve is when the timer stops, as defined in:
https://www.worldcubeassociation.org/regulations/#A6a
The end of the attempt is when the competitor signs, as defined in:
https://www.worldcubeassociation.org/regulations/#A7c

hanwu85: I am sorry the definition from my memory is not correct.  Please let me confirm with the WRC internally.  I will get back to you asap.

Jambrose777:
Why is his being pursued? A solve is whenever the competitor solves the puzzle. So it is inherently a subset of what attempt encompasses.

Goosly:
> Why is his being pursued? A solve is whenever the competitor solves the puzzle. So it is inherently a subset of what attempt encompasses.

Nothing is being pursued, I did not randomly change "attempt" into "solve". I've updated these two regulations to correctly say "end of the solve" instead of incorrectly saying "end of the attempt".

wca-regulations.md
hanwu85, line 455: I agree with this one, if it is "attempt", it will conflict with [A6e](https://www.worldcubeassociation.org/regulations/#A6e)

hanwu85:
@Goosly  thank you for spotting these.

PR #753, opened 2019-06-29
Improve the wording of 4b3.

lgarron:
Suggested by Gregor Grigorjan.

4b3b to 4b3d are still a tad ambiguous, but I think it's better to keep them concise instead of 100% clear.


hanwu85 (approved):
looks clear to me.

Rouxles (approved):
Looks good to me as well.

PR #762, opened 2019-07-21
Add credit section on `wca-states.md`.

AlbertoPdRF:
@thewca/board agreed that we should credit @Mollerz for his work crafting this list. I'll work on the PR for https://github.com/thewca/worldcubeassociation.org now!

### DIFF
--- a/wca-states.md
+++ b/wca-states.md
@@ -207,3 +207,8 @@ The WCA recognizes a total of 199 countries/regions:
 - (YE:_Asia) Yemen
 - (ZM:_Africa) Zambia
 - (ZW:_Africa) Zimbabwe
+
+
+## <credit>
+
+The list above was compiled by [James Molloy](mailto:jmolloy@worldcubeassociation.org).


jfly:
LGTM!

AlbertoPdRF:
Should I merge this one up @lgarron?

lgarron (approved):
LGTM assuming that

1) the Regs compiler can handle the `<credit>` title, and
2) Mollerz is happy to be credited.

Feel free to merge if both are the case!

AlbertoPdRF:
2. is the case, but for 1. I'll defer to @viroulep (https://github.com/thewca/worldcubeassociation.org/pull/4331#issuecomment-513775461).

viroulep (changes requested):
You can definitely install it and try to build it, it's a `pip` package and the build command is included in the generated json (eg [here](https://github.com/thewca/worldcubeassociation.org/blob/master/WcaOnRails/config/wca-states.json#L2)).

As I said in thewca/worldcubeassociation.org#4331 it most likely doesn't work because the compiler doesn't understand the `<credit>` section (it won't magically turn it into a nice title, and I think it will just crash :p).

I haven't had time to look into it yet!

AlbertoPdRF:
A quick update: I tried this locally and nothing crashed, it just didn't include anything regarding the credit section on the output JSON :slightly_smiling_face: 

Anyway, don't feel pushed for this! It will be nice to have, but not a high priority.

PR #792, opened 2019-08-26
Remove ambiguousness from A3b1

kotarot:
@hanwu85 has pointed out this ambiguousness. Would it be clearer if we correct words like this?

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -371,7 +371,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - A3a) The competitor may inspect the puzzle at the start of each attempt.
         - A3a1) The competitor is allotted a maximum of 15 seconds to inspect the puzzle and start the solve.
     - A3b) The judge prepares the timer by turning it on and resetting if necessary. Separately, the judge also prepares a stopwatch for timing inspection.
-        - A3b1) When the judge believes the competitor is ready, the judge asks "READY?". The competitor must be ready to start the attempt within one minute of being called, or else the competitor forfeits the attempt (DNS), at the discretion of the judge.
+        - A3b1) When the judge believes the competitor is ready, the judge asks "READY?". The competitor must be ready to start the attempt within one minute from the moment they are asked, or else the competitor forfeits the attempt (DNS), at the discretion of the judge.
         - A3b2) The competitor starts the attempt by confirming that they are ready, and the judge uncovers the puzzle and starts timing the inspection.
     - A3c) The competitor may pick up the puzzle during inspection.
         - A3c1) The competitor must not apply moves during inspection. Penalty: disqualification of the attempt (DNF).


xsrvmy:
I would word it more concisely as "within one minute of being asked".

mirandapedro:
Hello.

There are a few aspects of this article that I'd like to discuss.

`A3b1) When the judge believes the competitor is ready, the judge asks "READY?". The competitor must be ready to start the attempt within one minute of being called, or else the competitor forfeits the attempt (DNS), at the discretion of the judge.`

Firstly, in your experience in competitions, how often do judges actually ask a competitor if they are "READY"? I find competitors and judges comunicate mostly through body language, through a simple nod of the head, or by quickly saying "I'm ready". I only recall being asked "READY?" once, and it was a delegate doing it. 

I think we can either: 

1. Flexibilize this article so that the time the competitor can take to get ready is officially more fluid than one minute, relying on context to determine wheter or not a competitor is taking too long (as it's already done, considering almost no one actually asks a competitor if they are "READY" times the 1 minute they have);  or

2. Implement mechanisms so that the 1 minute is actually taken precisely and have a harsher approach to enforcing this article.

 To me, option 1 seems more elegant, although it brings up another issue, which is judge discretion.
How much should we keep it around? As far as I'm aware, judge discretion is mention six times on the Regulations, and I believe most of the should be reconsidered. A6e and A6f are specially troublesome for this.

TLDR: we should apply A3b1 more sistematically or we should admit it's heavily dependent on context and deal with the issue of judge discretion that surrounds it.

xsrvmy:
I don't think a stricter enforcement is reasonable. There are basically two options for stricter enforcement I can think of:
- Actually make it a time limit from being called. This is unfair as competitors often fail to notice getting called, and sometimes name confusions come up (I've even gotten an extra because of this, four Max's in my region). Also, often a competitor has to wait for a station after being called.
- Make it a time limit from some defined moment after the competitor reacts to the call (eg. from sitting down at the station). The problem I see here is deliberate time wasting in the competitor's area. Also a stationary rounds can make things hard to define.

Another reason I think discretion is necessary is that if something happens in the last 10 seconds that prevents the competitor from starting the attempt, there still needs to be an exception and that's when judge discretion is required.

IMO this reg is really similar in A1c for preventing time wasting. Therefore I don't think strict enforcement is necessary. 

And yes, A6e needs to be worded the other way around, and the current wording technically puts the discretion on whether to give DNF or +2 given that no move is made, which is definitely not what is supposed to be communicated.

hanwu85 (approved):

lgarron (approved):
Apologies, I just saw this now!

I think there are good arguments for being flexible, but also for allowing organizers to be strict if they have a tight schedule.

I think this is a good balance between the two, in that it doesn't make it harder for the former, but clarifies the latter.

Issue #718, opened 2019-03-29
Re-add surface to C1b


xsrvmy:
There has been some debate as to whether table abuse is allowed due to the wording of C1b. I believe part of the reason is that when we changed 'touch'to 'operate', the argument that 'the table must touch the cube a millisecond after the timer starts' no longer holds. 

I would suggest readding the surface to C1b and remove C1b+++

lgarron:
C1b+++) covers this, but a small fix update to C1b itself might be good!

PR #830, opened 2019-09-23
Clarify Puzzle on Surface Solving

hanwu85:
Clarifying the use of surface during one handed solve.
Issue[#718](<https://github.com/thewca/wca-regulations/issues/718>)

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -391,7 +391,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - A4e) Time penalties for starting the solve are cumulative.
 - A5) During the solve:
     - A5a) While inspecting or solving the puzzle, the competitor must not communicate with anyone other than the judge or the WCA Delegate. Penalty: disqualification of the attempt (DNF).
-    - A5b) While inspecting or solving the puzzle, the competitor must not receive assistance from anyone or any object other than the surface (also see [Regulation 2i](regulations:regulation:2i)). Penalty: disqualification of the attempt (DNF).
+    - A5b) While inspecting or solving the puzzle, the competitor must not receive assistance from anyone or any object other than the surface (also see [Regulation 2i](regulations:regulation:2i))(e.g. holding the puzzle against the surface to help operate the puzzle). Penalty: disqualification of the attempt (DNF).
 - A6) Stopping the solve:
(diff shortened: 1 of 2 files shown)


Goosly:
I'm not sure why this addition is needed. Isn't this covered in [C1b+++](https://www.worldcubeassociation.org/regulations/guidelines.html#C1b+++) already?

hanwu85:
Yes, it is indeed covered by C1b+++.
But C1b+++ states you can do A and B, but the reason why it is okay is not specified in the Regulations.  Some competitors think the wording is not clear enough in the Regulations.  Maybe it needs to be explicit stated.

wca-regulations.md
```diff
@@ -461,6 +461,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - C1b) During the solve, the competitor must use only one hand to operate the puzzle. Penalty: disqualification of the attempt (DNF).
         - C1b2) If a puzzle defect occurs, and the competitor chooses to repair it, they must repair it using only the solving hand. Penalty: disqualification of the attempt (DNF).
         - C1b3) If other body parts of the competitor come in contact with the puzzle or parts of the puzzle without the competitor's intention and without applying moves, this is not considered operating the puzzle, at the discretion of the judge.
+        - C1b4) During the solve, the competitors may choose to put the puzzle on the surface.
```
lgarron: How about this?

```suggestion
        - C1b4) During the solve, the competitor may hold the puzzle against the surface to help operate the puzzle.
```

Alternatively, I think it would be fine to reuse the wording from C1b+++ .

lgarron (commented):
Consideration: If we want to be thorough about this, we might want to make sure the clarification also works for speedsolves. I believe it's fairly common for Megaminx now.

lgarron (changes requested):
Also, I think we should make sure to remove C1b+++ at the same time!

wca-regulations.md
hanwu85: This is much better!  I will remove C1b+++ and move this one to Chapter A and make sure it applies to all speedsolving.

Ivan-Brigidano (changes requested):
Typo: The dot must be at the end of the sentence

hanwu85:
Fixed! Thanks!

Ivan-Brigidano (approved):

Issue #196, opened 2014-06-27
Intentional Misalignments in Inspection


KitClement:
Mike Hughey brought up an interesting point on misalignments in inspection:

_What if, for instance, the 3x3x3 was 5 degrees off, and you made it 10 degrees off? I actually see A3c2 as implying (but not very clearly stating) that this would not be acceptable - it's not aligning the faces; it's misaligning the faces ("the competitor may align the faces"). And one could see where making such an adjustment might actually help - it could make it easier to corner-cut, for instance. So perhaps such a silly example is actually important? (But I hope not, because I think watching for a 5 degree misalignment is essentially impossible to ask of judges.) Ability to corner-cut really can make a mess of these regulations if we want to protect against people taking advantage of misalignments._

_To give an example (fortunately, I've never seen this in competition yet), if you as a delegate saw some competitor who always made a slight turn (just a few degrees) on their 3x3x3 during inspection to match the first turn they're going to make in the solve, would that constitute sufficient grounds for DNF? And if so, on what basis? Because of what I mentioned above with A3c2? I'm afraid this could get quite ugly for judges._

A slight point, but regardless, it could use more clarification.

PR #833, opened 2019-09-26
Prevent intentional misalignment moves

hanwu85:
For addressing Issue #196.


lgarron:
Hmm, I'm a little concerned about this, given that small misalignments can often happen on many puzzles when you pick them up.

I think this would need to set a higher threshold or a sign that the competitor is trying to take advantage.

hanwu85:
You are right.  I have not thought about that.  I think this regulation is not completed yet.  I will modify it when I or someone else comes up with a better wording.

Issue #746, opened 2019-05-16
[Suggestion] Remove Article 6 from the Guidelines


EdHollingdale:
It seems odd that Prizes and Awards are covered in the regulations/guidelines. Longer term, should this become a seperate policy managed by WQAC? 

I think it fits more into a "making competitions good" rather than a "rules and regulations" setting. This would also give us opportunities to flesh out details and define expectations for awards ceremonies.

This may be better placed in wca-documents, but I'll leave it here for now.

cubewhiz:
As they are all just recommendations (and nothing in there is mandatory), I think the content belongs more fittingly in a guide for organizers or something similar. I agree they don't belong in the Regulations (but I don't think they need to be specified in a policy, either).

PR #835, opened 2019-09-26
6a+ fix

Ivan-Brigidano:
There is no sense of calling it an ADDITION as it doesn't refer to a regulation.

### DIFF
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -105,7 +105,7 @@ To be more informative, each Guideline is classified using one of the following 
 
 ## <article-6><awards><awards> Article 6: Awards/prizes/honors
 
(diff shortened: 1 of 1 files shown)


hanwu85:
I just realize the whole chapter 6 in the regulations is removed, should we still keep the chapter 6 in guidelines?  Or should we merge these guidelines to other chapters?

Ivan-Brigidano:
@hanwu85 the Issue #746 is about that. As it seems not to be a priority for 2020 Regulations, what's sure is that this little change should be done.

lgarron (approved):
We should definitely think about moving this out some day, but I agree that this change is appropriate as long as this is still in the Guidelines.

Issue #766, opened 2019-07-29
Allow scrambling extra backup cubes for Multi-Blind


lgarron:
Duplicate scrambles for two cubes in the same Multi-Blind are bad, **more easily preventable than most duplicate scrambles**, and it's been proven that you can pretty much eliminate the problem with attentive staff.

That said, duplicate scrambles are still happening. This hurts the competitor:
1) The competitor will need to spend extra time looking at cubes out of order to verify whether it's a duplicate scramble.
2) The cube has to be taken out of the total, lowering their maximum result by 1. Granting an extra is not practical at most competitions, and can be a drain on the competitor.
3) This creates an incentive *not* to report the duplicate scramble. Most competitors are honest, but it's better to avoid such incentives.

I've heard a suggestion that we should allow preparing one or two extra cubes for large attempts. They would be kept outside view and reach of the competitor, but the judge could swap out a duplicate scramble with one of these. This reduces the issue to only the first problem (wasted time/distraction for the competitor).

I think we should recommend this for attempts over, say, 10 cubes. This should be better for competitors, and should help us gather better data.

pedrosino:
I think in theory that looks good, but not everybody has another 2 cubes to serve as backups.
Last time I had to use two 7-cm 3x3x3 to fill up the number I wanted.

mintydev789:
I would also propose changing the scrambling procedure a bit, so that all cubes are scrambled in the same orientation to make checking for duplicates easier, and then randomizing the orientation for all of the cubes. This random order may also be printed out for the MBLD event.

To be clear, I am suggesting this in addition to what you proposed, not instead.

Jambrose777:
The scrambles randomize the rotation already and IMO his already makes it easier for checking for duplicates (as cubes will have he same orientation if they have the same scramble)

mintydev789:
> The scrambles randomize the rotation already and IMO his already makes it easier for checking for duplicates (as cubes will have he same orientation if they have the same scramble)

oh yeah, nvm I'm dumb

xsrvmy:
tbh problems like scrambler loosing track of cube number would be solved if there is a dedicated scramble viewer in the future (wasn't that discussed a while ago when the passwords were instroduced?)

Samuel-Baird:
As for the backup cube(s) idea

[...]

PR #851, opened 2019-10-10
Allowing backup puzzles as extra during Multi-Blind

hanwu85:
Addressing issue #766 .

Let's discuss if there are loophole or grey area in these wordings and over how many cubes should we recommend?

### DIFF
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -222,6 +222,8 @@ To be more informative, each Guideline is classified using one of the following 
 
 ## <article-H><multiple-blindfolded><multipleblindfoldedsolving> Article H: Multiple Blindfolded Solving
 
+- H1a+) [ADDITION] The competitors may choose to bring 1 or 2 extra puzzles as backup in case of incidents found during the attempt (e.g. duplicate scrambled puzzles). Attempts higher than 6 puzzles are highly recommendated to bring extra puzzles.
+- H1a++) [ADDITION] The extra puzzles should be scrambled together with the intended puzzles but kept away from the competitors untill incidents are found.
 - H1b1+) [REMINDER] The attempt is not disqualified for reaching the time limit, due to exceptions for 3x3x3 Multi-Blind (see [Regulation A1a4](regulations:regulation:A1a4) and [Regulation A1a5](regulations:regulation:A1a5)).
 - H1b1++) [ADDITION] The judge may permit the competitor to continue the attempt unofficially, but the attempt must be stopped and judged first, in full accordance with the Regulations.
 - H1d+) [EXAMPLE] Example: Suppose a competitor attempts 10 cubes, stops the solve with a time of 59:57, and has two time penalties. The time of the final result is 59:57 + 2*2 = 60:01 (also see [Regulation A1a5](regulations:regulation:A1a5)).


Rouxles (changes requested):
> - H1a+) [ADDITION] The competitors may choose to bring 1 or 2 extra puzzles as backup in case of incidents found during the attempt (e.g. duplicate scrambled puzzles). Attempts higher than 6 puzzles are highly recommendated to bring extra puzzles.

For this, would something like `The competitor may choose to bring 1 or 2 extra puzzles as backup in case of duplicate scrambles found during the attempt` be better?

For this, `Attempts higher than 6 puzzles are highly recommendated to bring extra puzzles.`, would it be better to put it as a sub-guideline (as in it's under H1a+)?

hanwu85:
My idea is the following,  if mis-scramble cubes, cubes with logos (or other illegal cubes) are found, the judges or delegates can swap the cubes.  If we say "duplicate cubes", those are not allowed to be swapped.   So that's why I think we should brainstorm to find loopholes.

What do you mean by sub-guidelines?

Ivan-Brigidano:
Why only 1 or 2 cubes and not was many as the competitor consider?

hanwu85:
> Why only 1 or 2 cubes and not as many as the competitor consider?

That's also my question, I think as many as they want will cause other issues like deliberately wasting competition resource.  
How many is considered enough is open for discussion.
 
I would also like to know how often MBLD incidents happen in competitions.

Ivan-Brigidano:
Maybe something like one cube for backup every 6 could work

Laura-O:
> Why only 1 or 2 cubes and not was many as the competitor consider?

And in case one of them is illegal, maybe also a replacement for them?

Seriously, has one of you ever organized a competition holding MultiBLD?
MultiBLD already is an organizational nightmare for several reasons. The more cubes competitors attempt and the more cubes you have to deal with, the worse it gets.

Apart from finding time to scramble puzzles, the main problem is that you have to find a place where you can collect the puzzles, scramble them and then find a way to somehow carry them to the place where competitors will do their attempts. All of this needs to be done while making sure that competitors do not see a scrambled cube.
We have scrambled MultiBLD in hallways, cellars or stuffy backstage rooms, sometimes standing there for 45 minutes because there was no space for a chair. I guess you can imagine that there would also be not much room for additional cubes.

The suggested change makes this worse and even creates more potential incidents:
1. You would have to keep 2 piles of cubes for each competitor (one for regular, one for backup cubes) and make sure they do not get mixed up, which basically means that you would have to print two scoresheets for each competitor.
2. You have to keep the scrambled backup cubes at a place where competitors can't see them until the round ends. At most competitions, it's already difficult to find a place to scramble them and keep them until the round starts. The backup cubes would have to be placed somewhere where they are accessible during the round but still hidden from the competitors.
3. What happens when regular cubes and backup cubes get mixed? With a large number of regular and backup cubes that's not unlikely.

Ivan-Brigidano:
> Seriously, has one of you ever organized a competition holding MultiBLD?

Yes and I'm aware of how difficult is (sometimes) to hold it.

I think that a few cubes are reasonable but no unlimited. I know this could make holding the event even more difficult and cause some new problems. I think this is reasonable if backup cube are not in a different room but in the same one covered with and opaque cloth.

Laura-O:
> I think that a few cubes are reasonable but no unlimited. I know this could make holding the event even more difficult and cause some new problems. I think this is reasonable if backup cube are not in a different room but in the same one covered with and opaque cloth.

This still means that you need to place them somewhere and you need more covers.

I can imagine that this works well at smaller competitions with a lot of space/rooms. However, in areas where competitions usually reach the competitor limit and many competitors are allrounders, who frequently do MultiBLD, this can be a serious problem.
I can remember several venues we used where putting 20 additional piles of cubes somewhere (and assuring scramble secrecy) would have been nearly possible. Not to mention the additional time it would have needed to scramble them.

Ivan-Brigidano:
> I can imagine that this works well at smaller competitions with a lot of space/rooms. However, in areas where competitions usually reach the competitor limit and many competitors are allrounders, who frequently do MultiBLD, this can be a serious problem.

Yes, there could be several issues in some competitions. Maybe another option could be allowing to re-scramble the cube that is duplicated.

EdHollingdale:
I agree fully with Laura. What problem are we trying to fix here? It feels like you are creating the possibility of a logistical nightmare to make a rare incident marginally better. The instances this would fix are when there is a duplicate scramble, or an illegal cube. Both of these can be fixed far better by encouraging a scramble checker - which also solves additional problems.

As an example, the other weekend we had 21 multi competitors, including a few with large attempts. It is already hard enough managing the cubes (making sure they stay together, are scrambled correctly, are hidden from view, are brought out to the solving stations). Let alone managing an extra ~40 back-up cubes which would need to be handled completely independently so as not to be mixed up with the main cubes, still need to be scrambled and moved into the main room. And now I need to get an extra 21 opaque cloths to cover them? All for the possibility that our most trusted scrambler and a delegate who was checking their work made a mistake? 

And even if there is a duplicate, how does having a backup cube help? It's not like the competitor is unaffected - they still have to notice the duplicate, call a delegate over, wait while the delegate confirms this, and then wait while the delegate works out under which of the 21 opaque cloths their additional cube is placed, then start memorising the new cube once the delegate has brought it over.

> Maybe another option could be allowing to re-scramble the cube that is duplicated.

Good idea! No regulations change needed!

wca-guidelines.md
```diff
@@ -222,6 +222,8 @@ To be more informative, each Guideline is classified using one of the following
 
 ## <article-H><multiple-blindfolded><multipleblindfoldedsolving> Article H: Multiple Blindfolded Solving
 
+- H1a+) [ADDITION] The competitors may choose to bring 1 or 2 extra puzzles as backup in case of incidents found during the attempt (e.g. duplicate scrambled puzzles). Attempts higher than 6 puzzles are highly recommendated to bring extra puzzles.
```
pedrosino, line 225: To be more in line with the rest of the text, I'd suggest to change "The competitors" to simply "Competitors" or "The competitor" (same as H1a)
Also not sure if the "1 or 2" is necessary. Competitors trying 40 cubes may want to have more backups, as the risk of incidents is higher.

Also, recommendated -> recommended

wca-guidelines.md
```diff
@@ -222,6 +222,8 @@ To be more informative, each Guideline is classified using one of the following
...
+- H1a++) [ADDITION] The extra puzzles should be scrambled together with the intended puzzles but kept away from the competitors untill incidents are found.
```
pedrosino, line 226: Suggestion:
"The extra puzzles must be scrambled together with the intended ones but kept away from the competitor. Extra puzzles must only be used if any incident is found during the attempt."

wca-guidelines.md
xsrvmy, line 225: Another language issues I see here: in the last sentence, "attempts" cannot bring extra puzzles.
Maybe "For attempts with more than 6 puzzles, extra puzzles are highly recommended." or "Competitors attempting more than 6 puzzles should bring extra puzzles."

lgarron:
> > Maybe another option could be allowing to re-scramble the cube that is duplicated.
> 
> Good idea! No regulations change needed!

Based on the feedback, the WRC has agreed that this is an appropriate way to handle it.

Issue #125, opened 2013-12-26
Publishing FMC Solutions


lgarron:
Competition organizers usually don't look at FMC solutions too much because it takes a lot of time -- they're just glad to have scored them once.

I still have physical FMC solutions from past US Nationals that I intended to type up and publish -- but I've been partly concerned that some competitors have an expectation of privacy in their solution. Considering that cubing is traditionally an open sport (e.g. the community is very collaborative, and filming official solves of fast cubers to learn from/reconstruct is popular), I think it would be awesome if FMC results were public data, possibly even _part of the result_ (i.e. available on the results site). However, for now it would help to make explicit that the organizer has the _right_ to publish an FMC solution.

(Do we have a precedent where a competition organizer published a solution after a record was set, instead of the competitor reporting it? After SF09, I put up results [here](http://ca.cubingusa.com/sanfrancisco2009/files/fmc_results.htm) but I don't know how many people have seen that.)

If we had a good mechanism, digital scoring might actually be faster than turning physical cubes (and give us data as a byproduct). The current FMC sheets came out of my attempts to make this happen, going back to 2009:
- FMCTyper: http://www.speedsolving.com/forum/showthread.php?8090-FCF-and-FMCTyper
- FMC Scramble Sheets: http://www.speedsolving.com/forum/showthread.php?37704-FMC-Scramble-Sheets

I still think digital FMC scoring is worth pursuing, but I've had other things to worry about.

iVinshe:
Easier to err on the side of privacy. Would it be simple to add a simple
opt-in check box for this?

lgarron:
You mean, on the FMC sheet? I think case-by-case is rather messy; one simple rule would be preferable. (Like the clear choices of "gender public" and "age private".)

I don't think anyone _should_ have an expectation of privacy for an FMC solution, I was mainly concerned that some _might_ have.

lgarron:
From #660 

> I'd like to support better FMC grading software, including software that allows uploading solutions for computer grading or assistance from people who are not the competition (thus freeing competition staff). This allows for easier competition logistics, provides a better record for accountability, and supports the open, constructive nature of the cubing community. (If I understand correctly, practically all top competitors freely share their official solutions and approaches.)
>
> To this end, I'd like us to clarify that FMC solutions are not secret.

Berkmann18:
It's definitely a good idea whether the workings behind a solution or not are published alongside said solution.
[...]

PR #852, opened 2019-10-10
Allow FMC solutions to be published

hanwu85:
Addressing issue #125

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -501,6 +501,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - E3b) 3x3x3 Cubes (at most 3, self-supplied), as described in [Article 3](regulations:article:3).
     - E3c) Stickers (self-supplied).
     - E3d) Stopwatch or watch (self-supplied) for keeping track of the elapsed time, if it is approved by the WCA Delegate.
+- E4) The organization team may choose to publish the solutions (e.g. photo of the score sheets or transcriptions of solutions).
 
 
 ## <article-F><clock><clocksolving> Article F: Clock Solving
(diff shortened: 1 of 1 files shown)


Laura-O:
Why the organization team?
I think this should always be approved by Delegates, especially as they are responsible for keeping the FMC sheets.

hanwu85:
Changed!  Thanks Laura!

wca-regulations.md
```diff
@@ -501,6 +501,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - E3b) 3x3x3 Cubes (at most 3, self-supplied), as described in [Article 3](regulations:article:3).
     - E3c) Stickers (self-supplied).
     - E3d) Stopwatch or watch (self-supplied) for keeping track of the elapsed time, if it is approved by the WCA Delegate.
+- E4) The Wca Delegate may choose to publish the solutions (e.g. photo of the score sheets or transcriptions of solutions).
```
lgarron, line 504: Here's a way to phrase it that is more consistent with other Regulations:

```suggestion
- E4) Competitor solutions (e.g. photo of the score sheets, transcriptions of solutions) may be published at the discretion of the WCA Delegate.
```

(We usually use commas instead of "or", and "WCA in "WCA Delegate" should be uppercase.")

lgarron (commented):
I have some tweaks, but sounds good to me!

I'd also recommend a Guideline clarifying that this may happen during the competition, or even during the attempt if enough precautions are taken.
(Note that someone can walk out of an attempt after 1 minute and post the scramble on the internet, so this is theoretically already a concern. But we still need to be careful.)

wca-regulations.md
lgarron: @hanwu85, this has a thumbs-up from @Ivan-Brigidano; are you alright with it?

hanwu85: @lgarron, Yes, this looks okay to me.

lgarron: Thanks!

lgarron (approved):

Issue #724, opened 2019-04-11
Rename "opaque object" for BLD to "sight blocker"


lgarron:
"Sight blocker" seems to be the most common modern term for it.

Samuel-Baird:
The term “sight blocker” may not be as clear to someone reading the regs as an opaque object. If the term sight blocker was to be defined in the regulations as an opaque object to be used for obstructing the competitors line of view between them and the cube during the execution phase then I think it would be good :)

dancing-jules:
I agree with renaming it. A fork is an "opaque object", but it's terrible as a sight blocker.

Samuel-Baird:
How’s this? Define the sight blocker as an opaque object which prevents the competitor from seeing the puzzle (so no one uses forks :P)

B4c) The judge must ensure that there is an sight blocker object between the competitor's face and the puzzle while the competitor is solving. The sight blocker is opaque and must fully block the competitors view of the puzzle.

lgarron:
I will gladly approve and merge a pull request to change B4c and B4c1!

(If not, we'll get around to it with the 2020 Regs.)

PR #864, opened 2019-10-13
Rename "opaque object" for BLD to "sight blocker"

Ivan-Brigidano:
Addressing #724


lgarron (approved):
Looks like this is the only time the term appears in the Regulations or Guidelines, so looks good to me. Thanks for drafting this! :-D

Issue #867, opened 2019-10-20
Clarify if A1c applies to FMC


Samuel-Baird:
I recently saw someone on Facebook say that they have rejected competitors solutions in the past for wasting competition resources when competitors use an excessive amount of rotations or wide moves. That of course is completely legal according to the regs but then a member of the WRC stated that the WRC confirmed that A1c can be applied to FMC for things like competitors using excessive rotations, wide moves or intentionally making their solutions longer. A1c doesn’t apply to FMC according to the regs and I haven’t been able to find anything public stating that such practices are illegal so I’m wondering if this is the case.

Ivan-Brigidano:
Solved in PR #868

PR #868, opened 2019-10-20
Turn A1c into 2j3

Ivan-Brigidano:
Addressing Issue #867. This way the regulation affects all events and not only speed solving events.


lgarron:
This would probably be the best solution if we were introducing this today.

However, I believe A1c is one of the Regulations most often referenced by number. Do you think there might be a good way to do this that preserves the A1c numbering?

xsrvmy:
> Do you think there might be a good way to do this that preserves the A1c numbering?

How about leaving A1c as is, and adding
E2d2) The solution must not excessively use wide moves or rotations in a way that makes it hard to judge. Penalty: DNF, at the discretion of the WCA Delegate

Ivan-Brigidano:
@lgarron what do you mean with "referenced"? There in no reference inside the regulations so I guess you mean in the reports.

Maybe we can create 2j3 and maintain it simpler `A1c) A competitor participating in an event must be able to fulfill the event's requirements (see [Regulation 2j3](regulations:regulation:2j3)).`.

xsrvmy:
@Ivan-Brigidano I think he meant regular cubers quote this reg as A1c instead of the contents a lot.

lgarron (approved):
After reflection, I don't really have a major concern with changing this to 2j3. The history is available, and the intention does not change, so I don't expect significant confusion.

We try to preserve Regulations numbering when possible, but it's more of a nice-to-have andthe location of 2j3 makes more sense overall.

xsrvmy:
Hmm I see that this still does not address the concern that the competitor can write a solution that is an absolute pain to grade (eg. translating everything to wide moves) since it is not a poor result.

wca-regulations.md
```diff
@@ -90,6 +90,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 2j1) If a competitor is disqualified from an event for any reason, they are not eligible for any more attempts in the event.
         - 2j1a) The results of all remaining attempts in the event are recorded as DNF.
     - 2j2) If a competitor is disqualified during the course of an event, their earlier results remain valid. Exception: cheating or defrauding (see [Regulation 2k2a](regulations:regulation:2k2a)).
+    - 2j3) If a competitor does not fulfill the event's requirements (e.g. know how to solve the puzzle). A competitor must not compete with the expectation of a DNF result or an intentionally poor result.
```
hanwu85, line 93: I think this might be better.

If a competitor does not fulfill the event's requirements (e.g. know how to solve the puzzle)**, the** competitor must not compete with the expectation of a DNF result or an intentionally poor result.

xsrvmy: The original wording contains a fragmented sentence. The suggested wording does not carry the correct meaning (the reg would not ban knowing how to solve the puzzle, but simply wasting time)

Ivan-Brigidano: @hanwu85 all 2j uses `a` instead of `the`. So we can change all 2j or keep as how it is.

> The original wording contains a fragmented sentence. The suggested wording does not carry the correct meaning (the reg would not ban knowing how to solve the puzzle, but simply wasting time)

Yes, I will correct it!

PR #854, opened 2019-10-10
Require scramblers signatures

Ivan-Brigidano:
Addressing #686

PR #883, opened 2019-11-30
Clarify how to handle missing scramble signatures.

lgarron:
This is a complement to #854.

While scramble signatures are a useful step for accountability, avoids holding competitors responsible or penalizing them for something new. With scramble signatures nevertheless being required, we should be able to get information on whether whether being more strict in the future 1) is practical, and 2) would have addressed incidents that come up.


wca-guidelines.md
```diff
@@ -165,6 +165,8 @@ To be more informative, each Guideline is classified using one of the following
...
 - A1c+) [CLARIFICATION] WCA Delegates should only use their discretion to prevent competitors from being a severe detriment to the competition (e.g. wasting time and/or competition resources). Competitors should not be disqualified for a "poor" result when they are competing to the best of their abilities.
 - A2c1+) [REMINDER] In the past, score sheets have been used to cover puzzles while leaving some sides exposed. This is no longer permitted.
+- A2d1+) [ADDITION] If a missing scramble signature is discovered before an attempt, the puzzle should immediately be returned to a scrambler for verification and a corresponding signature.
+- A2d1++) [ADDITION] If a missing scramble signature is discovered after the beginning of the attempt, the result may be kept, at the discretion of the Delegate. However, the WCA Delegate should not allow such incidents to become common.
```
lgarron: We could suggest looking at video footage. But I don't think it's a good idea to require that (even if available) and the incentives are the wrong way around for it. So I think it's better to leave out.

Ivan-Brigidano:
```suggestion
- A2d1++) [ADDITION] If a missing scramble signature is discovered after the beginning of the attempt, the result may be kept, at the discretion of the WCA Delegate. However, the WCA Delegate should not allow such incidents to become common.
```

Ivan-Brigidano (changes requested):

wca-guidelines.md
lgarron: Thanks!

Ivan-Brigidano (approved):

Issue #798, opened 2019-08-30
Systematic Leniency


lgarron:
A lot of times, we need to make the Regulations more strict to be more consistent (for "fair and equal conditions"). However, I think we can be more lenient in some situations. I'd like to make a list of them here as they come to mind. Please suggest more!

- Allow video evidence to be used if the time was not recorded properly, as long as we have confidence in the accuracy of the result.
- Tolerating interaction with electronic equipment, as long as we have confidence that no advantage was gained. (via @SAuroux)
- Tolerating camera screen visibility due to realistically no tangible advantage can be gained (similar to the reason why stickerless cubes were eventually allowed) (via @SAuroux)
- Tolerating touching the puzzle (in the context of A6e) if it is clear that no move has been applied. (via @SAuroux)
- Don't remove competitors from rounds they were originally allowed to participate in unless it was *their fault*. In particular, don't enforce https://www.worldcubeassociation.org/regulations/#9p1 if a competitor from a pervious round is retroactively disqualified.
- Allow competitors to use a scramble/get an extra if they lifted a cover and before the attempt and it was clearly by accident.
- Allow an extra attempt if a competitor was accidentally set up camera before the attempt a screen facing them (since they won't be able to make this "mistake" based on seeing the scramble).

Baiqiang:
For video evidence, the screen of timer must be in the evidence because display may be incorrect.

xsrvmy:
I think it depends. If the time written down is ambiguous like 11 vs 17, it is possible to time the video. But if the difference is smaller (eg. 10 was written down instead of 10.00) then the actual timer time would be better.
btw isn't there a way to check if a display is working properly by displaying all 8s?

Jambrose777:
@xsrvmy 

1. Would you rather have a completely bogus time that is known to be incorrect as a time that is in our results, or one that accurately represents what happened for the attempt and can be proven?
2. Display will sometimes lag behind if they don’t work well causing them to display a time that is not on the timer. The all 8s thing only applies to the timer.

xsrvmy:
By the all 8 thing I meant testing to see if the display's LEDs are correctly working. As far as I am aware that's the only reason where I have seen an incorrect time on the display

SAuroux:
> However, I think we can be more lenient in some situations. I'd like to make a list of them here as they come to mind. Please suggest more!

There is a ton of potential systematic leniency could both simplify and improve the current regulations and make decisions more efficient in practice. Some examples:

[...]

PR #884, opened 2019-11-30
"Systematic leniency": allow waiving penalties under certain conditions. Addresses #798.

lgarron:
Ideally, we would like to learn when Delegates believe that it is more fair to waive certain penalties instead of being more strict. This gives them a mechanism to do so. The change requires noting this in the Delegate report so that we can understand which Regulations may need clarifications or revisions in order to ensure fair rulings across the world.

I believe that Delegates are generally in favor of this, although we didn't really get feedback about how to implement it. I believe that this is a sufficiently flexible start that we can refine in the future.


xsrvmy:
I've not sure if intentionally is the best word here. I would argue that touching camera during inspection is intentional.

wca-regulations.md
```diff
@@ -291,6 +291,10 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 11a3) Equipment malfunction.
 - 11b) If an incident occurs, the WCA Delegate determines an impartial and appropriate course of action.
 - 11d) If the WCA Regulations are not fully clear or if the incident is not covered by the WCA Regulations, then the WCA Delegate must make a decision based on fair sportsmanship (also see [Regulation 11e3](regulations:regulation:11e3)).
+  - 11d1) The WCA Delegate may waive a penalty or award an extra attempt even when Regulation does not directly specify the use of their discretion. This must be noted in the Delegate report, and satisfy the following conditions:
```
```suggestion
  - 11d1) The WCA Delegate may waive a penalty or award an extra attempt even when Regulations does not directly specify the use of their discretion. This must be noted in the Delegate report, and satisfy the following conditions:
```

wca-regulations.md
```diff
@@ -291,6 +291,10 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
...
+    - 11d1a) The competitor did not cause the incident, or did not cause it intentionally.
+    - 11d1b) The incident did not give the competitor a significant advantage for their result.
```
```suggestion
    - 11d1b) The incident did not give the competitor an unfair advantage for their result.
```

wca-guidelines.md
```diff
@@ -144,6 +144,7 @@ To be more informative, each Guideline is classified using one of the following
 
 ## <article-11><incidents><incidents> Article 11: Incidents
 
+- 11d1+) [EXAMPLE] Examples of incidents where it is appropriate to waive a penalty: touching a camera during inspection (see [Regulation 2i2b](regulations:regulation:2i2b)), lightly touching a puzzle at the end of the attempt (see [Regulation A6e](regulations:regulation:A6e)).
```
Ivan-Brigidano, line 148: Why `lightly` and not `slightly`? Is that because the one you use is American English and the other is British?

Ivan-Brigidano (changes requested):
I would also add:
`11d1a+) [REMINDER] The competitor must not have caused the incident internationally.`
Or something similar.

Note: I didn't find why it fails the CI

wca-guidelines.md
lgarron: Perhaps. I think "lightly" is definitely more idiomatic.

wca-regulations.md
lgarron: Good idea, that is more consistent with other with other Regulations.

wca-regulations.md
lgarron:
```suggestion
  - 11d1) The WCA Delegate may waive a penalty or award an extra attempt even when the relevant Regulation does not directly specify the use of their discretion. This must be noted in the Delegate report, and satisfy the following conditions:
```

lgarron:
> I've not sure if intentionally is the best word here. I would argue that touching camera during inspection is intentional.

You're right, this could be clearer. I've changed this so that it says they must not have tried to gain an advantage.

lgarron:
@Claster, @julesDesjardin: Are you comfortable with this wording?

wca-guidelines.md
lgarron: "lightly touch" seems to have about 1x as many Google results as "slightly touch", so I think I'd like to stay with that.

Ivan-Brigidano:
I would add the third point

> 11d1c) The competitor must not have caused the incident intentionally.

Touching the camera is not always intentionally, in fact, most cases of competitors touching the camera are because of a knee-jerk reaction. Maybe a guideline could clarify it:

> 11d1c+) [CLARIFICATION] A knee-jerk reaction is not consider an intentionally incident.

Maybe not the best wording thought.

lgarron:
Given that we 

> I would add the third point
> 
> > 11d1c) The competitor must not have caused the incident intentionally.
> 
> Touching the camera is not always intentionally, in fact, most cases of competitors touching the camera are because of a knee-jerk reaction. Maybe a guideline could clarify it:
> 
> > 11d1c+) [CLARIFICATION] A knee-jerk reaction is not consider an intentionally incident.
> 
> Maybe not the best wording thought.

The best I know is "forced" vs. "unforced" error. But that's not a super common phrase.

Given the discussion here and other places, I think I'm okay with leaving this draft out of the 2020 Regulations and refining it in collaboration with Delegates.

PR #885, opened 2019-11-30
Update A6c from judge discretion to Delegate

Ivan-Brigidano:


lgarron (approved):
We concluded that it's better to involve the Delegate because of incidents like:
- https://www.worldcubeassociation.org/incidents/23 
- https://www.worldcubeassociation.org/incidents/45

Such incidents already reach the Delegate (and the Delegate report) often.

PR #886, opened 2019-11-30
Update A6e to Delegate discretion

Ivan-Brigidano:
Addressing [this incident](https://www.worldcubeassociation.org/incidents/23)


lgarron (approved):
Thanks for the catch, #885 was indeed meant to include A6e as well (they're both about touching).

PR #889, opened 2019-12-01
Change "combined rounds" to "cutoff rounds". Closes #543.

### LINKED ISSUE #543: Get rid of the term "combined round"

lgarron:


Ivan-Brigidano (changes requested):
You forgot to change it in regulation 9o).

Rouxles:
Can't seem to find the PR from here, is the link dead?

lgarron:
> Can't seem to find the PR from here, is the link dead?

Hmm, https://github.com/thewca/wca-regulations/pull/889/files should show it.

Rouxles (approved):

Ivan-Brigidano (approved):

Issue #382, opened 2016-08-11
Changes to Clock Scrambles/Regulations


CoreySakowski:
A couple of proposals to make clock run more easily (more specifically, to avoid problems with the puzzle falling over while the cover is being lifted, and thus having to rescramble it):

1) For all clock scrambles, the pins will all be either up or down.
This would make the event easier to run, as the clock can be placed flat on the surface prior to inspection, and there would be no fear of it falling over. We would only have to make sure the clock is actually facing up the correct way.

2) Pins can be altered during inspection, and the position of the pins prior to inspection does not matter.
This would make things even easier. It is worth noting that pin positions do not affect whether the puzzle is solved, so by that logic, changing a pin does not affect the scrambled state of a puzzle either. But of course, this would shave off a significant amount of time at the beginning of the solve.

timreyn:
I agree with (1).  I think that pin positions are more work to enforce than they're worth.  Actually I would go a step further and say that the pins will all be up for consistency.  That way the scrambler can just immediately lie the puzzle on the table when they're done scrambling.

I disagree with (2).  I think that shaves a reasonably large amount of time.  I also think it makes first-clock skips easier, since you don't need to remember to go to a different pin position than you normally do, and the main reason that skips are hard to deal with is the change in routine.  Finally, I don't like the idea of allowing the competitor to manipulate the puzzle in only certain ways during inspection; I think this will make our lives more difficult in the long run.  The more event-specific rules we have, the more complicated things become for inexperienced judges, and I don't want judges to get confused about which manipulations are allowed during inspection.

AaronJamesBlair:
1) This would definitely save a massive amount of time for scramblers, runners, judges, and competitors. At a large event, something like US Nationals, this would save several minutes of time, allowing for the entire competition to stay on track a little easier. At a small or more localized event, this makes it easier for those judging who aren't competitors or familiar with clock to judge without confusion. I think all of the pins should be up no matter what, for consistency and ease of runner delivery. It speeds up scrambling and the start of the attempt for the competitor. The judge, at the most, would need to push the clock onto the table to push all of the pins up before starting the attempt.

2) I disagree with this due to the complications of ensuring that only pins are moving. While I can't speak for everyone, I know that even if I'm only playing with the pins on a clock, I push them up with my hands in position to start rotating the wheels.  This could make it hard for a judge to tell if only pins are moving.

[...]

PR #890, opened 2019-12-01
Specify that clock stands are allowed.

lgarron:
This is an alternative to #382 that addresses clocks falling over when the cover is removed.

We don't have a lot of experience with competitions trying clock stands, but we would like to learn if they are sufficiently beneficial.

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -509,7 +509,8 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 
 - F1) Standard speed solving procedures are followed, as described in [Article A](regulations:article:A) (Speed Solving). Additional regulations that supersede the corresponding procedures in [Article A](regulations:article:A) are described below.
 - F2) The judge places the scrambled puzzle onto the mat in a standing position.
-    - F2a) A stand may be used to hold the puzzle in standing position.
+    - F2a) The organization team may enforce using a stand for the puzzle to prevent it from falling before the start of the attempt. If these objects are used, this must be announced before the round starts.
+        - F2a1) The judge must remove the stand from the mat immediately after the competitor removes the puzzle from it.
 - F3) At the end of the inspection period, the competitor places the puzzle in a standing position. Penalty: disqualification of the attempt (DNF).
     - F3a) The competitor must not change the positions of any pins from their scrambled positions before the start of the solve. Penalty: disqualification of the attempt (DNF).
 
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -509,6 +509,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 
 - F1) Standard speed solving procedures are followed, as described in [Article A](regulations:article:A) (Speed Solving). Additional regulations that supersede the corresponding procedures in [Article A](regulations:article:A) are described below.
 - F2) The judge places the scrambled puzzle onto the mat in a standing position.
+    - F2a) A stand may be used to hold the puzzle in standing position.
 - F3) At the end of the inspection period, the competitor places the puzzle in a standing position. Penalty: disqualification of the attempt (DNF).
     - F3a) The competitor must not change the positions of any pins from their scrambled positions before the start of the solve. Penalty: disqualification of the attempt (DNF).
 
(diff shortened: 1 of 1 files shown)


wca-regulations.md
```diff
@@ -509,6 +509,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 
 - F1) Standard speed solving procedures are followed, as described in [Article A](regulations:article:A) (Speed Solving). Additional regulations that supersede the corresponding procedures in [Article A](regulations:article:A) are described below.
 - F2) The judge places the scrambled puzzle onto the mat in a standing position.
+    - F2a) A stand may be used to hold the puzzle in standing position.
```
Ivan-Brigidano: I would prefer a more similar wording to `A2b1) For Square-1, the organization team may enforce placing a thin object in the puzzle to prevent accidental moves from being applied before the start of the attempt. If these objects are used, this must be announced before the round starts.`

```suggestion
    - F2a) The organization team may enforce placing a stand in the puzzle to prevent accidental falling before the start of the attempt. If these objects are used, this must be announced before the round starts.
```

hanwu85:
I have a question.  For clock stands, are they purely for organization issue?(prevent a clock falling after cover removed)  Or can the competitors use them after inspection?  If the competitors can/can not use them after inspection, I think we need to address that too.

lgarron:
> I have a question. For clock stands, are they purely for organization issue?(prevent a clock falling after cover removed) Or can the competitors use them after inspection? If the competitors can/can not use them after inspection, I think we need to address that too.

We haven't really been able to agree. So I think we should only allow it for the beginning of inspection, and discuss whether to allow using it at the end of inspection throughout the year.

(e.g. get top clock solvers to try it at home to see if they think it's an unfair to require or allow them)

wca-regulations.md
lgarron: I don't really like the way A2b1 does this for now. But I think that it's reasonable to copy for now.

How's this?

```suggestion
    - F2a) The organization team may enforce using a stand for the puzzle to prevent it from falling before the start of the attempt. If these objects are used, this must be announced before the round starts.
```

lgarron: And since this was unclear enough that @hanwu85 asked about it:

```suggestion
    - F2a) The organization team may enforce using a stand for the puzzle to prevent it from falling before the start of the attempt. If these objects are used, this must be announced before the round starts.
        - F2a1) The judge must remove the stand from the mat immediately after the competitor removes the puzzle from it.
```

Rouxles (approved):
Don't see any problems with these - really similar to sq1

wca-regulations.md
```diff
@@ -509,6 +509,8 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 
 - F1) Standard speed solving procedures are followed, as described in [Article A](regulations:article:A) (Speed Solving). Additional regulations that supersede the corresponding procedures in [Article A](regulations:article:A) are described below.
 - F2) The judge places the scrambled puzzle onto the mat in a standing position.
+    - F2a) The organization team may enforce using a stand for the puzzle to prevent it from falling before the start of the attempt. If these objects are used, this must be announced before the round starts.
```
lgarron: More consistent wording:

```suggestion
    - F2a) The organization team may enforce using a stand for the puzzle to prevent it from falling before the start of the attempt. If such stands are used, this must be announced before the round starts.
```

Rouxles (approved):

wca-regulations.md
```diff
@@ -509,6 +509,8 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 
 - F1) Standard speed solving procedures are followed, as described in [Article A](regulations:article:A) (Speed Solving). Additional regulations that supersede the corresponding procedures in [Article A](regulations:article:A) are described below.
 - F2) The judge places the scrambled puzzle onto the mat in a standing position.
+    - F2a) The organization team may enforce using a stand for the puzzle to prevent it from falling before the start of the attempt. If such stands are used, this must be announced before the round starts.
+        - F2a1) The judge must remove the stand from the mat immediately after the competitor removes the puzzle from it.
```
Ivan-Brigidano, line 513: I’m not sure if this should be a “must”

Issue #894, opened 2019-12-02
New A6c/A6e changes doesn't seem to work


xsrvmy:
There was a recent A6c/A6e update in the PRs that changed the discretion from judge to delegate. Apparently the intention was to allow systematic leniency (like that 4x4 WR) but it does not have the effect of doing so. The new wording only slow down comps by requiring the delegate to come over for almost every single A6c/A6e violation and still does not allow waiving the penalty.
(I have chosen not to comment on the PRs since they are merged already)

lgarron:
We already planned to update this after https://www.worldcubeassociation.org/incidents/23

The Delegate is already usually involved in these decisions, and systematic leniency needs to rely on the discretion of the Delegate to be accountable.

I don't expect that we will revert this update, but I think we should be open to it if others feel strongly.

Issue #856, opened 2019-10-12
WCA Regulations — Changes for 2020


lgarron:
Here is a copy of an email I will be sending to WCA staff and organizers shortly.

--------

Dear WCA Community,

The WCA Regulations Committee (WRC) has evaluated current WCA issues, and we are considering the following potential changes to the WCA Regulations, taking effect at the beginning of 2020. The WRC has discussed these and our consensus is in favor of each of these changes. We'd like your feedback on whether there are major concerns that we might not have considered yet.

If you know anyone in your local community who might be affected by these changes, please help us get feedback from them! If you can contribute feedback for each change in the associated GitHub issue link, this is very helpful for us. However, feedback over email is also welcome.

## **Simplifications**

If you can think of other simplifications to the Regulations that do not fundamentally change the nature of our sport, please let us know!

- Clock pins are allowed to be distinguishable. [#759](https://github.com/thewca/wca-regulations/issues/759)
  - Pro: Many competitors have puzzles like this that they would like to use, and there are already various ways to tell a clock's orientation without pins.
- Clock pins are all pushed down at the end of the scramble. #382 
  - Pro: This makes it easier to scramble correctly and transport the scrambled puzzle without affecting the state.
  - Pro: We already ignore pins for the solved state. This effectively simplifies the scramble state to be consistent.
  - Con: This will simplify the work that competitors need to do at the beginning of an attempt. This is not fair to past competitors.
    - Counterpoint: Most competitors will still start their solves by moving pins.
    - Counterpoint: 1/8 of scrambles already had all pins on the same side. If this is a significant advantage, it has already applied to many individual results.
- Solved state: All pieces must be attached and normally placed. [#223](https://github.com/thewca/wca-regulations/issues/223)
  - Pro: there are currently many states between clearly "solved" vs."unsolved" (e.g. 0 vs. 1 vs. 2 slice misalignments, 0 parts detached vs. 1 part with a single color vs. 1 part with multiple colors vs. 2 colored parts) — Regulation 5b5 alone has 5 parts! This significantly simplifies decisions about whether all parts are attached.
  - Con: Competitors would be penalized if this happens by accident at the end of the solve.
    - Counterpoint: The old policy was essentially designed to accommodate center caps that could pop off easily. This is uncommon with modern puzzles.
[...]

PR #897, opened 2019-12-03
Revert A6c/A6e to judge's discretion.

lgarron:
As pointed out in #894, this doesn't completely solve the problem with the incidents we've had. While it's important to have Delegates involved, they usually already are in practice.

While this is fairly minor, it's essentially the only change in `draft` that wasn't in the changes listed at #856.  After discussing it with @Ivan-Brigidano, it should be okay to revert until we figure out where exactly to go with systematic leniency.


xsrvmy:
My point actually was, the +2 vs DNF decision should be made by the judge, but if the penalty is waived  completely then maybe the delegate should be involved.

Ivan-Brigidano (approved):

PR #926, opened 2019-12-08
Deleting also from (see Regulation ...)

Ivan-Brigidano:


lgarron (approved):

Laura-O:
I don't understand this PR.
I am not a native speaker, but in my opinion, "see" and "also see" mean two different things:
see = a reference to the same topic, like a source or a definition
also see = an addition, something which might be important in the context

All of the changed regulations basically intend to do the latter. For example, "If incidents are not covered by the regulations, Delegates must make fair decisions, but they can **also** give a provisional extra if they are not 100% sure.

What's the reason for this change?

EdHollingdale:
I agree with @Laura-O, the also in these cases are saying “there may be another regulation that covers this instance in addition to this one” rather than “this is due to some other regulation, see here”. In my opinion we shouldn’t have any of the second type - the regulations shouldn’t be repeating themselves.

Ivan-Brigidano:
My initial idea was to make a draft PR instead of a PR.
As I'm not a native English speaker, I initially thought that this is a formatting inconsistency. Thanks you both for your explanation!

PR #927, opened 2019-12-08
Modify A2b1 to not require an announcement

Ivan-Brigidano:


Ivan-Brigidano:
@thewca/wrc-team as suggesting by some Delegates. Do you think we should still recommending an announcement in the Guidelines?

Samuel-Baird:
I think an announcement is nice to have and should be kept. Since inserts are optional competitors should know what to expect before they solve.

lgarron:
I know at least one competitor has complained about what paper inserts do to their puzzle, and in general I don't think the announcement is too much to ask.

So even if they're commonly expected, I think we should not take it out of the Regs just yet.

Samuel-Baird:
@Ivan-Brigidano why do you think this should be changed from the current regs?

xsrvmy:
TBH I think the must needs to be changed to should just to avoid the exploit of asking for an extra for the lack of an announcement when the real reason is a bad CS case.

Sixstringcal:
An announcement literally allows people to decide if they will compete based on this.  Due to the harmful nature of inserts to the lubrication, it's entirely reasonable to refuse to compete if inserts will be present.  If it's unknown if inserts will be used, that can be problematic.  It can even decide for a competitor what puzzle to use (ie, if one has DNM in it and the other doesn't, you shouldn't use the one with DNM in it because the DNM will get absorbed if the inserts are absorbent).

julesDesjardin:
Calvin, if you think inserts are a reason valid enough not to compete, nothing prevents you from DNSing the round after the first solve. (and I'm not sure it is "entirely reasonable", but that's subjective. Of course you find it reasonable, but you're in a tiny minority)

What exactly makes DNM different, compared to any other lube?

xsrvmy:
@juleDesjardin Probably because he is using it right before competing for the short term effects, and having a paper insert interferes. Same deal with maru.

Sixstringcal:
The damage is already done once the insert is put in for solve 1.  Additionally, what puzzle people use can be decided on this.  If one is heavily reliant on water-based lubes and another is not, a competitor could use the one not heavily reliant on water based lubes and have the effects of the insert lessened.

DNM is water based.

Ivan-Brigidano:
I want to clarify that this is not a PR but a Draft PR because it hasn't been discussed in the WRC, yet.
I agree with the proposal of moving the announcement from must to should. This way still the Organization Team would need to announce it but if they don't then the regulation wouldn't be violated. So any competitors could claim an unfair extra just because they didn't hear the announcement.
Additionally, any SQ-1 solver should have in mind that papers are typical and accepted this. The regulation is here to avoid accidental miss-scrambles because of low hardware.

lgarron:
I'll leave this open for discussion, but just to be clear, we won't merge this into 2020 Regs!

Ivan-Brigidano:
> I'll leave this open for discussion, but just to be clear, we won't merge this into 2020 Regs!

I've created an issue so I think it is better to continue the discussion there. #935

PR #922, opened 2019-12-06
Fix 2j3 wording

Ivan-Brigidano:

PR #928, opened 2019-12-08
Move 2j3 changes to 2k6

Ivan-Brigidano:
As spotted by @Laura-O at #922 , the A1c movement makes more sense under 2k.

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -358,7 +358,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - A1b) If the time limit for an attempt is greater than 10 minutes, a stopwatch must be used for timekeeping.
         - A1b1) A Stackmat must be used in addition to the stopwatch. Exception: for events without inspection where the competitor is expected to exceed 10 minutes, use of the Stackmat is optional.
         - A1b2) If a time from the Stackmat timer is available, it is the original recorded time. Otherwise, the stopwatch time is the original recorded time.
-    - A1c) A competitor participating in an event must be able to fulfill the event's requirements (see [Regulation 2j3](regulations:regulation:2j3)).
+    - A1c) A competitor participating in an event must be able to fulfill the event's requirements (see [Regulation 2k6](regulations:regulation:2k6)).
 - A2) Scrambling:
     - A2a) When called for a round, the competitor submits the puzzle, in its solved state, to the scrambler. The competitor then waits in their assigned Competitors Area until they are called to compete.
     - A2b) A scrambler scrambles the puzzle according to the regulations in [Article 4](regulations:article:4).
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -90,7 +90,6 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 2j1) If a competitor is disqualified from an event for any reason, they are not eligible for any more attempts in the event.
         - 2j1a) The results of all remaining attempts in the event are recorded as DNF.
     - 2j2) If a competitor is disqualified during the course of an event, their earlier results remain valid. Exception: cheating or defrauding (see [Regulation 2k2a](regulations:regulation:2k2a)).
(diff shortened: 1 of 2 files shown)


wca-regulations.md
```diff
@@ -90,14 +90,14 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
...
-    - 2j3) If a competitor does not fulfill the event's requirements (e.g. know how to solve the puzzle). A competitor must not compete with the expectation of a DNF result or an intentionally poor result.
 - 2k) At the discretion of the WCA Delegate, a competitor may be disqualified from some events (a single event, multiple events, or all events) if the competitor:
     - 2k1) Fails to check in or register in time for the competition.
     - 2k2) Is suspected of cheating or defrauding the officials during the competition.
         - 2k2a) The WCA Delegate may disqualify any suspected results.
     - 2k3) Behaves in a way that is unlawful/violent/indecent, or intentionally damages venue facilities or personal property within the venue.
     - 2k4) Distracts or interferes with others during the competition.
     - 2k5) Fails to abide by WCA Regulations during the competition.
+    - 2k6) Does not fulfill the event's requirements (e.g. not knowing how to solve the puzzle). A competitor must not compete with the expectation of a DNF result or an intentionally poor result.
```
Ivan-Brigidano, line 100: Does the example make sense of being negative?

xsrvmy:
I have a suggestion for splitting 2k6 into 2k6 and 2k7
2k6) Does not fulfill the event's requirements
2k7) competes with the intention of wasting competition resources (eg. competing with the expectation of an intentionally poor result, intentionally makes an FMC result hard to grade)

xsrvmy:
Also there is a problem with placing the change here: this would allow the delegate to disqualify the competitor from the entire competition (but A1c does not)

Ivan-Brigidano:
> I have a suggestion for splitting 2k6 into 2k6 and 2k7
> 2k6) Does not fulfill the event's requirements
> 2k7) competes with the intention of wasting competition resources (eg. competing with the expectation of an intentionally poor result, intentionally makes an FMC result hard to grade)

@xsrvmy I think it makes sense to have both in a single regulation as both phrases are related.

> Also there is a problem with placing the change here: this would allow the delegate to disqualify the competitor from the entire competition (but A1c does not)

I don't think this is a big deal as it's still Delegate discretion. I don't think this might cause any problems as any Delegate would disqualify a competitor from the entire competition because of not knowing how to solve a specific event.

xsrvmy:
We still need to mention the FMC issue.

Ivan-Brigidano:
> We still need to mention the FMC issue.

What do you mean here?

lgarron:
> @xsrvmy I think it makes sense to have both in a single regulation as both phrases are related.

Agreed.

> I don't think this is a big deal as it's still Delegate discretion. I don't think this might cause any problems as any Delegate would disqualify a competitor from the entire competition because of not knowing how to solve a specific event.

I also agree with this. It's very rare for a Delegate to disqualify a competitor from the entire competition, and I think it's unlikely to happen by "accident". However, it may actually be appropriate in this situation, e.g. if a competitor has signed up for 3x3x3 followed by 3x3x3 BLD but clearly can't even solve a cube sighted.

> > We still need to mention the FMC issue.
> 
> What do you mean here?

I think @xsrvmy might mean adding this to 2k7?  I don't think it's really needed.

wca-regulations.md
lgarron: I think it's good to stay similar to the old A1c, since it still works well enough.

lgarron (approved):

xsrvmy:
Just noticed a different issue: A1c allows an individual attempt to be disqualified. 2k only allows disqualification from the event. Which one takes precedence?

Ivan-Brigidano:
> Just noticed a different issue: A1c allows an individual attempt to be disqualified. 2k only allows disqualification from the event. Which one takes precedence?

I don't think there is a problem with that. I think both regulations are compatible

wca-regulations.md
Ivan-Brigidano: Notice that 2k force to make it negative. I think the change makes sense

PR #936, opened 2019-12-12
Clarify how to handle missing scramble signatures.

lgarron:


wca-guidelines.md
```diff
@@ -165,6 +165,9 @@ To be more informative, each Guideline is classified using one of the following
...
 - A1c+) [CLARIFICATION] WCA Delegates should only use their discretion to prevent competitors from being a severe detriment to the competition (e.g. wasting time and/or competition resources). Competitors should not be disqualified for a "poor" result when they are competing to the best of their abilities.
 - A2c1+) [REMINDER] In the past, score sheets have been used to cover puzzles while leaving some sides exposed. This is no longer permitted.
+- A2d1+) [ADDITION] If a missing scramble signature is discovered before an attempt, the puzzle should immediately be returned to a scrambler for verification and a corresponding signature.
+- A2d1++) [ADDITION] An attempt must not be disqualified solely due to a missing scramble signature. If a scramble signature is missing after the beginning of an attempt, the judge should allow the attempt to continue (if it is still in progress) and the WCA Delegate should allow the result to be kept. However, the WCA Delegate may award an extra attempt if they have a reason to suspect that the scramble was incorrect, other than the missing signature.
```
xsrvmy: I believe that "the judge should allow" should be a "must".
Also aren't "must"s supposed to be in the regulations and not the guidelines?

wca-guidelines.md
```diff
@@ -165,6 +165,9 @@ To be more informative, each Guideline is classified using one of the following
 - A1a4+) [REMINDER] If a competitor has accidentally been permitted to exceed the time limit, the time limit must be enforced retroactively, and the judge, competitor, and WCA Delegate should be informed (see [Regulation 1g2](regulations:regulation:1g2)). Judges must always be aware of the time limit for a current attempt (which might depend on previous attempts, in the case of a cumulative time limit).
 - A1c+) [CLARIFICATION] WCA Delegates should only use their discretion to prevent competitors from being a severe detriment to the competition (e.g. wasting time and/or competition resources). Competitors should not be disqualified for a "poor" result when they are competing to the best of their abilities.
 - A2c1+) [REMINDER] In the past, score sheets have been used to cover puzzles while leaving some sides exposed. This is no longer permitted.
+- A2d1+) [ADDITION] If a missing scramble signature is discovered before an attempt, the puzzle should immediately be returned to a scrambler for verification and a corresponding signature.
```
xsrvmy: the puzzle -> the puzzle and the scorecard
Not sure is "a missing scramble signature is discovered" makes sense (can you discover something that is missing?). I would word it as "If the scramble signature is found (to be) missing before an attempt, ..."

wca-guidelines.md
lgarron: There is one important "must" here, which is because some Delegates feel strongly (and no one really seems to be against) forbidding a DNF for a missing signature. That shouldn't really happen, and the Regulations should not imply that this is normal or "permitted", so I think an `ADDITION` Guideline is best.

I'd be alright with changing this to "the judge *must* allow the attempt to continue", but I think it's better to avoid a "must" here unless it's needed. @Ivan-Brigidano, any thoughts?

wca-guidelines.md
```diff
@@ -164,7 +164,8 @@ To be more informative, each Guideline is classified using one of the following
 - A1a4+) [REMINDER] If a competitor has accidentally been permitted to exceed the time limit, the time limit must be enforced retroactively, and the judge, competitor, and WCA Delegate should be informed (see [Regulation 1g2](regulations:regulation:1g2)). Judges must always be aware of the time limit for a current attempt (which might depend on previous attempts, in the case of a cumulative time limit).
 - A2c1+) [REMINDER] In the past, score sheets have been used to cover puzzles while leaving some sides exposed. This is no longer permitted.
 - A2d1+) [ADDITION] If a missing scramble signature is discovered before an attempt, the puzzle should immediately be returned to a scrambler for verification and a corresponding signature.
```
Ivan-Brigidano: What @xsrvmy proposed sounds better to me

```suggestion
- A2d1+) [ADDITION] If the scramble signature is found to be missing before an attempt the puzzle should immediately be returned to a scrambler for verification and a corresponding signature.
```

wca-guidelines.md
```diff
@@ -164,7 +164,8 @@ To be more informative, each Guideline is classified using one of the following
...
-- A2d1++) [ADDITION] If a missing scramble signature is discovered after the beginning of the attempt, the result may be kept, at the discretion of the WCA Delegate. However, the WCA Delegate should not allow such incidents to become common.
+- A2d1++) [ADDITION] An attempt must not be disqualified solely due to a missing scramble signature. If a scramble signature is missing after the beginning of an attempt, the judge should allow the attempt to continue (if it is still in progress) and the WCA Delegate should allow the result to be kept. However, the WCA Delegate may award an extra attempt if they have a reason (other than the missing signature) to suspect that the scramble was incorrect.
```
Ivan-Brigidano: I agree that a `must` is not proper for a guideline. I think a `should` would also work pretty well here.
```suggestion
- A2d1++) [ADDITION] An attempt should not be disqualified solely due to a missing scramble signature. If a scramble signature is missing after the beginning of an attempt, the judge should allow the attempt to continue (if it is still in progress) and the WCA Delegate should allow the result to be kept. However, the WCA Delegate may award an extra attempt if they have a reason (other than the missing signature) to suspect that the scramble was incorrect.
```

Ivan-Brigidano: Actually thinking a little bit more, I'm happy leaving the must here

Goosly: The regulations currently use the verbs "replace", "grant" and "receive" when referring to "extra". Can "award" be changed to "grant" here? Receiving an extra isn't really an award :-)
Also the regulations never use "however", it seems a bit redundant here.

Ivan-Brigidano:
```suggestion
- A2d1++) [ADDITION] An attempt must not be disqualified solely due to a missing scramble signature. If a scramble signature is missing after the beginning of an attempt, the judge should allow the attempt to continue (if it is still in progress) and the WCA Delegate should allow the result to be kept. The WCA Delegate may grant an extra attempt if they have a reason (other than the missing signature) to suspect that the scramble was incorrect.
```

wca-guidelines.md
lgarron: I think it's important to keep the comma, but sounds good!

```suggestion
- A2d1+) [ADDITION] If the scramble signature is found to be missing before an attempt, the puzzle should immediately be returned to a scrambler for verification and a corresponding signature.
```

wca-guidelines.md
lgarron: A few things:
- Given Delegate feedback, I think it's important to keep the "must".
- We do use "However" (once) in the Guidelines already. I think it's valuable to keep it here to emphasize that this is the exceptional case.
- You're right about "award" — I totally forgot that we stick with "grant" now. However, I think that if there is a good reason to use an extra, we should emphasize that it's not optional, so I've changed it to "replace the attempt with an extra attempt" below. (https://www.worldcubeassociation.org/regulations/#11e3 already covers the case where the extra should be provisional.)
- Based on feedback, changing the judge's action to "must".

```suggestion
- A2d1++) [ADDITION] An attempt must not be disqualified solely due to a missing scramble signature. If a scramble signature is missing after the beginning of an attempt, the judge must allow the attempt to continue (if it is still in progress) and the WCA Delegate should allow the result to be kept. However, the WCA Delegate may replace the attempt with an extra attempt if they have a reason (other than the missing signature) to suspect that the scramble was incorrect.
```

wca-guidelines.md
lgarron: I think "the puzzle" is sufficiently clear, and I'd prefer to keep it more concise.
Second suggestion adopted below!

Ivan-Brigidano (approved):

wca-guidelines.md
```diff
@@ -163,8 +163,9 @@ To be more informative, each Guideline is classified using one of the following
...
-- A2d1++) [ADDITION] If a missing scramble signature is discovered after the beginning of the attempt, the result may be kept, at the discretion of the WCA Delegate. However, the WCA Delegate should not allow such incidents to become common.
+- A2d1+) [ADDITION] If the scramble signature is found to be missing before an attempt, the puzzle should immediately be returned to a scrambler for verification and a corresponding signature.
+- A2d1++) [ADDITION] An attempt must not be disqualified solely due to a missing scramble signature. If a scramble signature is missing after the beginning of an attempt, the judge must allow the attempt to continue (if it is still in progress) and the WCA Delegate should allow the result to be kept. However, the WCA Delegate may replace the attempt with an extra attempt if they have a reason (other than the missing signature) to suspect that the scramble was incorrect.
+- A2d1+++) [REMINDER] Although the WCA Delegate may allow several individual attempts with missing scramble signatures to count, they should ensure that missing scramble signatures do not become common.
```
Ivan-Brigidano, line 168: It sounds a bit strange to me to call "they" to the Delegate/s
```suggestion
- A2d1+++) [REMINDER] Although the WCA Delegate may allow several individual attempts with missing scramble signatures to count, the WCA Delegate should ensure that missing scramble signatures do not become common.
```

lgarron: We do use "they" for the WCA Delegate in some places: https://www.worldcubeassociation.org/regulations/#4b2a

Repeating "WCA Delegate" sounds very redundant to my ears, so I'll stick with "they" if that's okay!
(I'm assuming it's okay with you because you already approved it.)

Issue #948, opened 2020-01-06
Minor (but important?) fixes


mintydev789:
12a2 "n may be also be omitted" should be changed to "n may also be omitted"
a2d1 What happens if there isn't a signature from the scrambler/scramble checker? Is that intentionally not stated because this has just been introduced? I can't quite tell if this really is mandatory or not.
a5a This means that if someone other than these people starts talking to a competitor they cannot do anything to ask them to stop other than ask the judge to do so. I recommend adding a point about allowing some involuntary remark that does not aid the competitor in any way at the discretion of the judge.
a6h it should say that noone except for a WCA delegate is allowed to even touch the puzzle in case of a dispute. I heard people say that, but it seems like it's not actually in the regs, you just can't align the puzzle.
a7b1 replace the / from "before/starting" with a space
a7c2 Just a question, is this mandatory? Cause I've never seen this enforced.
a7c4 I have seen multiple instances of delegates allowing a signature to be made after the fact, so the reg should probably be updated to include sportsmanship and what not.
b1b should it say that the blindfold must not be see through?
b4 should it instead say that the memo phase starts and a sight blocker is placed AFTER the first move is applied?
f3a but they can rotate the gears/clocks?
h1a3 shouldn't it be "arranged in A shape"?

hanwu85:
Hi denthebro,

Thank you for the issues.  I will answer you as much as I can.
`12a2` Good catch!
`A2d1` Because there is a "must" in the sentence.  It is mandatory to do so.  However, at this moment, a missing signature is not punished, but the WCA Delegate should do their best to prevent this.  You can refer to the [Guideline](https://www.worldcubeassociation.org/regulations/guidelines.html#A2d1+) for more information.
`A5a` If someone is talking to the competitors during their solves, the judge should stop that.  Or the competitors can just ask for extra attempts.
`A6h` Actually, even the WCA Delegate should not touch the puzzle before it's resolved, so it is not written in the Regulations.
`A7b1` The before or starting is actually correct, because the inspection phase and the "starting" (hand position, touching puzzle...etc) of the solve are written in the same penalty area.  
`A7c2` It is mandatory, and it is enforced.  You must only sign it AFTER the solve.
`A7c4` It is allowed to sign after the attempt but it must be done before the end of the competition.  Please refer to [A7c4+](https://www.worldcubeassociation.org/regulations/guidelines.html#A7c4+)
[...]

PR #952, opened 2020-01-10
Add guideline B4c+ (BLD sight blocker timing)

hanwu85:
The judge should put up the sight blocker after the competitors have applied the first move, because competitors can remove the blindfold if they have not applied any moves.

Thanks @denthebro for pointing out this. #948

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -446,7 +446,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 - B4) Blindfolded phase:
     - B4a) The competitor dons the blindfold to start the blindfolded phase.
     - B4b) The competitor must not apply moves to the puzzle before they have fully donned the blindfold. Penalty: disqualification of the attempt (DNF).
-    - B4c) After the competitor has applied the first move to the puzzle, the judge must ensure that there is a sight blocker (e.g. a sheet of paper or cardboard) between the competitor's face and the puzzle.
+    - B4c) As soon as the competitor applies the first move to the puzzle, the judge must ensure that there is a sight blocker (e.g. a sheet of paper or cardboard) between the competitor's face and the puzzle.
         - B4c1) In all cases, the competitor must wear the blindfold such that their view of the puzzle would still clearly be blocked if the sight blocker were not in the way.
         - B4c3) If the judge and competitor agree beforehand, the competitor may choose to place the puzzle behind a suitable object (e.g. a music stand, the surface of the table) by themselves during the blindfolded phase.
     - B4d) The competitor must not look at the puzzle at any point during the blindfolded phase. Penalty: disqualification of the attempt (DNF).
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -446,9 +446,8 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 - B4) Blindfolded phase:
     - B4a) The competitor dons the blindfold to start the blindfolded phase.
     - B4b) The competitor must not apply moves to the puzzle before they have fully donned the blindfold. Penalty: disqualification of the attempt (DNF).
-    - B4c) The judge must ensure that there is a sight blocker between the competitor's face and the puzzle while the competitor is solving.
+    - B4c) After the competitor has applied the first move to the puzzle, the judge must ensure that there is a sight blocker (e.g. a sheet of paper or cardboard) between the competitor's face and the puzzle.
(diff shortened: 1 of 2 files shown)


Ivan-Brigidano (approved):
It sounds reasonable to me. But I would like to see more opinions from other WRC members before merging it into the Draft

mirandapedro:
Hi.

Taking a closer look at the Regulations, I've found some inconsistencies regarding when should the judge put up the the blocker between the competitor and the puzzle.

We're normally taught that it should only be done after the competitor applies the first move to the puzzle. This is in order to follow B4e, where a competitor may undon the blindfold before they've made the first move.

Following this line of thought, the guideline here being proposed is fine, and IMO should only be and "CLARIFICATION", if we understand it is only following what is in the Regulations.

However, B4c2 states the jughe should place the objetc while the competitor is wearing the blindfold. If we follow this, after the competitor has donned the blindfold, but before he has made his first move, the judge should still be using the blocker.

This is unnecessary, as the competitor can go back to the memorization phase and see the puzzle any time he wants, and clearly doesn't reflect what happens in competitions.

My suggestion is to remove B4c2 and add the proposed guideline as an article in the Regulations, perhaps as an edit to B4c's current wording. It could look like this:

"B4c) After the competitor has applied the first move to the puzzle, the judge must ensure that there is a sight blocker between the competitor's face and the puzzle.

mirandapedro (changes requested):
Hi.

Taking a closer look at the Regulations, I've found some inconsistencies regarding when should the judge put up the the blocker between the competitor and the puzzle.

We're normally taught that it should only be done after the competitor applies the first move to the puzzle. This is in order to follow B4e, where a competitor may undon the blindfold before they've made the first move.

Following this line of thought, the guideline here being proposed is fine, and IMO should only be and "CLARIFICATION", if we understand it is only following what is in the Regulations.

However, B4c2 states the jughe should place the objetc while the competitor is wearing the blindfold. If we follow this, after the competitor has donned the blindfold, but before he has made his first move, the judge should still be using the blocker.

This is unnecessary, as the competitor can go back to the memorization phase and see the puzzle any time he wants, and clearly doesn't reflect what happens in competitions.

My suggestion is to remove B4c2 and add the proposed guideline as an article in the Regulations, perhaps as an edit to B4c's current wording. It could look like this:

"B4c) After the competitor has applied the first move to the puzzle, the judge must ensure that there is a sight blocker between the competitor's face and the puzzle.

mirandapedro (approved):

mirandapedro:
If @Ivan-Brigidano and @hanwu85 agree to these changes I believe we can merge it into the draft.

hanwu85:
I think the purpose of B4c2 is to prevent a DNF when a judge forgets to put the sight blocker for a short period of time. (eg, 1 second)

mirandapedro:
But how does it do that? 

Regulation B4c2 was added on [commit b2802c9](https://github.com/thewca/wca-regulations/commit/b2802c923fc2b88a8d72fc7b0d30940fd691204a).

I believe it's original purpose was to have the judge place the sight blocker between the competitor and the puzzle at all times, even before the competitor has made applied the first move to the puzzle. 

Even then, when the commit was made, B4c and B4c2 were contradictory. I don't think there's a need for the Regulation B4c2.

wca-regulations.md
```diff
@@ -446,9 +446,8 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 - B4) Blindfolded phase:
     - B4a) The competitor dons the blindfold to start the blindfolded phase.
     - B4b) The competitor must not apply moves to the puzzle before they have fully donned the blindfold. Penalty: disqualification of the attempt (DNF).
-    - B4c) The judge must ensure that there is a sight blocker between the competitor's face and the puzzle while the competitor is solving.
+    - B4c) After the competitor has applied the first move to the puzzle, the judge must ensure that there is a sight blocker (e.g. a sheet of paper or cardboard) between the competitor's face and the puzzle.
```
lgarron: A small nit: "after" technically applies to the entire solving period. We can use an alternative that makes it clear that the goal is to cover during as much of the solving period as possible.

```suggestion
    - B4c) As soon as the competitor applies the first move to the puzzle, the judge must ensure that there is a sight blocker (e.g. a sheet of paper or cardboard) between the competitor's face and the puzzle.
```

lgarron (commented):
One small nit, but I'd love to see this landed!

PR #953, opened 2020-01-10
Clarify A1b2

### LINKED ISSUE #616: Possible bug with stopwatches and timer resets
According to A1b2, it seems that if the time limit for an event is over 10 minutes, the competitor is not DNFed for a timer reset, and instead the stopwatch time is used. This is inconsistent with events with time limits under 10 minutes. Is this considered a bug or intended?

Goosly:
Fixes #616

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -357,7 +357,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - A1a5) The time counting towards the time limit for an attempt is either the result after time penalties are applied (if the result was not DNF) or the time elapsed in the solve (if the result was DNF).
     - A1b) If the time limit for an attempt is greater than 10 minutes, a stopwatch must be used for timekeeping.
         - A1b1) A Stackmat must be used in addition to the stopwatch. Exception: for events without inspection where the competitor is expected to exceed 10 minutes, use of the Stackmat is optional.
-        - A1b2) If a time from the Stackmat timer is available, it is the original recorded time. Otherwise (e.g. the Stackmat timer malfunctioned, ran for over 10 minutes or wasn't used at all), the stopwatch time is the original recorded time.
+        - A1b2) If a time from the Stackmat timer is available, it is the original recorded time. Otherwise (e.g. the Stackmat timer malfunctioned, ran for over 10 minutes or was not used at all), the stopwatch time is the original recorded time.
     - A1c) A competitor participating in an event must be able to fulfill the event's requirements (see [Regulation 2k6](regulations:regulation:2k6)).
 - A2) Scrambling:
     - A2a) When called for a round, the competitor submits the puzzle, in its solved state, to the scrambler. The competitor then waits in their assigned Competitors Area until they are called to compete.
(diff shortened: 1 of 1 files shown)


Ivan-Brigidano (commented):
Thanks for submitting this @Goosly!
This could be achieved with an example inside the regulation (as you proposed) or a guideline. Is there a specific reason why you think this is better with an example inside the regulation rather than an independent guideline?

xsrvmy:
I originally reported 616 because I believed this is not the intended interpretation. WRC porbably should comment on this.

Goosly:
@Ivan-Brigidano A1b2 says "If a time from the Stackmat timer is available" but it is not defined what "available" means. The definition of "available" should be in the regulations, not hidden in the guidelines.

@xsrvmy This came up in a report thread recently and the WRC commented there.

wca-regulations.md
```diff
@@ -357,7 +357,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - A1a5) The time counting towards the time limit for an attempt is either the result after time penalties are applied (if the result was not DNF) or the time elapsed in the solve (if the result was DNF).
     - A1b) If the time limit for an attempt is greater than 10 minutes, a stopwatch must be used for timekeeping.
         - A1b1) A Stackmat must be used in addition to the stopwatch. Exception: for events without inspection where the competitor is expected to exceed 10 minutes, use of the Stackmat is optional.
-        - A1b2) If a time from the Stackmat timer is available, it is the original recorded time. Otherwise, the stopwatch time is the original recorded time.
+        - A1b2) If a time from the Stackmat timer is available, it is the original recorded time. Otherwise (e.g. the Stackmat timer malfunctioned or ran for over 10 minutes), the stopwatch time is the original recorded time.
```
Claster: Otherwise (e.g. the Stackmat timer malfunctioned, ran for over 10 minutes **or wasn't used at all**)

Goosly: ✔️

wca-regulations.md
```diff
@@ -357,7 +357,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - A1a5) The time counting towards the time limit for an attempt is either the result after time penalties are applied (if the result was not DNF) or the time elapsed in the solve (if the result was DNF).
     - A1b) If the time limit for an attempt is greater than 10 minutes, a stopwatch must be used for timekeeping.
         - A1b1) A Stackmat must be used in addition to the stopwatch. Exception: for events without inspection where the competitor is expected to exceed 10 minutes, use of the Stackmat is optional.
-        - A1b2) If a time from the Stackmat timer is available, it is the original recorded time. Otherwise, the stopwatch time is the original recorded time.
+        - A1b2) If a time from the Stackmat timer is available, it is the original recorded time. Otherwise (e.g. the Stackmat timer malfunctioned, ran for over 10 minutes or wasn't used at all), the stopwatch time is the original recorded time.
```
```suggestion
        - A1b2) If a time from the Stackmat timer is available, it is the original recorded time. Otherwise (e.g. the Stackmat timer malfunctioned, ran for over 10 minutes or was not used at all), the stopwatch time is the original recorded time.
```

Ivan-Brigidano: As is just a format modification, I'm directly merging this into your PR
Edit: Apparently I can't push to "this repository". Probably because you configured the PR so administrators can't directly submit changes

Goosly: image

Ivan-Brigidano: I was not able to to commit as a suggestion but I was able to commit as a standard commit 🤷‍♂️

Ivan-Brigidano (approved):
LGTM!

xsrvmy:
What happens if the timer does not start? (wasn't used at all would imply using stopwatch, but again inconsistency)

wca-regulations.md
xsrvmy: A stackmat cannot run for over 10 minutes. A better wording would be "stopped after 10 minutes", or "the solve exceeded 10 minutes"

Goosly:
> What happens if the timer does not start?

The same as when there is no stopwatch, that is the whole point of this regulation.

PR #958, opened 2020-01-27
Improve wording of 9g. Fixes #957

### LINKED ISSUE #957: Clarify that a cutoff is a requirement, not a result
As discussed in the report thread of Guildford Open 2020

Goosly:
Fixes #957


Nanush7 (approved):
Hi,
I read the Guildfold Open 2020 thread.
As an addition to the change I would add a mathematical expresion, just like: _"**< 2 minutes** and not **<= 2 minutes**"_. I don't know if I'm being clear.
I suggest this because for some people it would be easier to understand.
I don't even know if using mathematical expresions is a good practise for this kind of documents, it's just a suggestion.

xsrvmy:
I believe in the past, top X places cutoffs were allowed. This would reintroduce that possibility. There should be a clarification that the cutoff must be a result rather than a placement.

Samuel-Baird:
The line that caused the issue at Guildford Open 2020 was A, in order to fix future issues then A should be changed to something more similar to B. Also the current PR draft is conflicting with the use of "better than" and "meets" so it's even more confusing than current regs :P the current wording seems to mean "better than or equal to" so changing it to that (or not changing it from the current regs) should fix the issue

A. Current line in 9g
If the competitor meets the cutoff requirement in at least one of their cutoff phase attempts, they are eligible for the remaining attempts.

B. Improved line in 9g
If the competitor meets or achieves a result better than the cutoff result in at least one of their cutoff phase attempts, they are eligible for the remaining attempts.

Goosly:
"meet" refers to "requirement"
"better than" refers to 2 minutes.
"better than 2 minutes" is a requirement.

It's not conflicting.

sam596:
I agree that this is not conflicting and is a lot more clear than it is currently. Meeting a requirement of less than two minutes is (more or less) the same as beating the result of 2 minutes. Meeting the requirement sounds more "professional" for lack of a better term.

Samuel-Baird:
I think I realized where my confusion has come from, this PR should be fine. I do see an issue though, some scorecards like the first one below use the format < cutoff while others like the second just list a time as the cutoff (cubecomps follows this format iirc). This is why I was thinking that it was conflicting. The second scorecard list 1:15.00 as the cutoff, "meets" would make you think that a result of 1:15.00 or better allows them to finish their average/meanwhile the use of "better than" would allow them to continue only if they get a time better than 1:14.99.

If this PR gets accepted then maybe a 9g1 or 9g+ should be created stating that if a time of X is listed for the cutoff then the cutoff will be interpreted as < X? 

Capture

43567004-f7e82b7a-9662-11e8-8dd3-d8cf3fa6ed4d

Ivan-Brigidano:
What about making an EXAMPLE Guideline?
This would make the Regulation much clearer.

sam596:
@UnderwaterCuber

The problem is not the scoresheets. Scoresheets aren't what define what the cutoff is for an event, The WCA Events page does.

The problem is that the regulations currently define a cutoff as a time you have to meet, not a requirement you have to meet.

@Ivan-Brigidano
I don't agree with making a guideline as this is pivotal to how the cutoff system is defined.

Samuel-Baird:
> @UnderwaterCuber
> 
> The problem is not the scoresheets. Scoresheets aren't what define what the cutoff is for an event, The WCA Events page does.
> 
> The problem is that the regulations currently define a cutoff as a time you have to meet, not a requirement you have to meet.
> 
> @Ivan-Brigidano
> I don't agree with making a guideline as this is pivotal to how the cutoff system is defined.

I am aware of that,I don’t think all of the judges at comps are looking at the WCA events page though :P I think it could be nice to clarify the issue in the regs but ultimately organizers should be responsible for making sure the scorecards make sense

Ivan-Brigidano (approved):
LGTM. @thewca/wrc-team if you are fine with the PR, we can merge it.

Ivan-Brigidano:
See #1011

PR #967, opened 2020-03-05
Rename Vatican City country name

### LINKED ISSUE #962: Holy See should be called Vatican City in list of states
Vatican City is actually the name of the country. Holy See is the jurisdiction of the Bishop of Rome (the pope), kind of like a company or something. It is known as Holy See within places like the UN, Council of Europe, etc. However this operates from and exercises "exclusive dominion" over the independent Vatican City State enclave.

So yeah that's an easy change I think

Ivan-Brigidano:
Fixes #962 
Thanks @Mollerz for spotting it!


hanwu85 (approved):

hanwu85:
Thanks! @mollerz and Ivan!

PR #970, opened 2020-03-31
Add WCA Forum to the Source

Ivan-Brigidano:


wca-regulations.md
```diff
@@ -24,7 +24,7 @@ Original source of the WCA regulations: [www.worldcubeassociation.org/regulation
 WCA Regulations in [PDF format](link:pdf)
 
 ### Source
-Development of the WCA Regulations and Guidelines is public [on GitHub](https://github.com/thewca/wca-regulations).
+Development of the WCA Regulations and Guidelines is public [on GitHub](https://github.com/thewca/wca-regulations) and the discussion is public [on WCA Forum](https://forum.worldcubeassociation.org/c/regulations).
```
lgarron:
```suggestion
Development of the WCA Regulations and Guidelines is public [on GitHub](https://github.com/thewca/wca-regulations) and the discussion is public [on the WCA Forum](https://forum.worldcubeassociation.org/c/regulations).
```

lgarron:
Thanks for drafting this!

If we want to update this between major updates, I think we should:

- Ask the Board for approval for a textual revision.
- Merge this into `official`.
- Merge `official` into `draft`.

Ivan-Brigidano:
> Thanks for drafting this!
> 
> If we want to update this between major updates, I think we should:
> 
> * Ask the Board for approval for a textual revision.
> * Merge this into `official`.
> * Merge `official` into `draft`.

I don't think is strictly necessary to ask the Board because this is not actually a Regulations change. I will ask the Board just in case though.

What do you mean by "Merge `official` into `draft`"? If you make a Pull Request of changes from `official` into `draft` you will have some unwanted changes.

lgarron:
> I don't think is strictly necessary to ask the Board because this is not actually a Regulations change. I will ask the Board just in case though.

Thanks!

> What do you mean by "Merge `official` into `draft`"? If you make a Pull Request of changes from `official` into `draft` you will have some unwanted changes.

Hmm, it looks like there was such a merge recently, so I think this shouldn't be much of an issue?

Ivan-Brigidano:
> > What do you mean by "Merge `official` into `draft`"? If you make a Pull Request of changes from `official` into `draft` you will have some unwanted changes.
> 
> Hmm, it looks like there was such a merge recently, so I think this shouldn't be much of an issue?

Well there are two small commits. So we can do it and then creating a new PR with these small typos fixes.

PR #976, opened 2020-05-06
Adaptation of 9p1 to reflect WRC decision - Fixes #220

### LINKED ISSUE #220: Incidents resulting in fewer competitors advancing.
Occasionally, a retroactive correction/disqualification affects more than just the competitor in question.

Austrian Big Cube Open 2014 had an interesting case [relevant thread on results list](https://groups.google.com/forum/#!searchin/wca-results/austrian/wca-results/JsWBh7AT2jQ/or4p6jPLyUUJ):
- Matěj Mužátko's second result in [round 1](https://www.worldcubeassociation.org/results/c.php?i=AustrianBigCubeOpen2014&allResults=1#e666_d) was changed to DNF.
- As a result, Matěj Mužátko's didn't have a first round average, and didn't qualify for the second round.
- Since this reduced the number of second round competitors from 15 to 14, this means that the final round could only have 10 competitors instead of 11 (due to [91p](https://www.worldcubeassociation.org/regulations/#9p1)).
- Therefore, Viktor Kalmar had to be retroactively removed from the final round.

So, a change in someone's first round result affected someone two rounds later. This sort of thing happens regularly, although it's usually less complicated.

Although we don't want to allow organizers to get around the Regulations, this sort of change still affects competitors who had done nothing wrong.

mirandapedro:
This PR reflects the understanding the WRC adopted on the decision made on [Incident 56](https://www.worldcubeassociation.org/incidents/56) and fixes #220 

With the proposed wording for Regulation 9p1, round results in which a competitor should not have competed in can be removed without affecting the results of other competitors.

This is because the obligatory elimination of 25% of the competitors between rounds will be applied on _competitors with recorded results at the time the round ended_. Therefore, if a competitor found guilty of cheating has their result removed at a later date, the number of competitors able to advance to the following round won't decrease, as the competitor found guilty of cheating did have recorded results at the time the round ended.

EDTI: Adding further explanation of the proposal.

The proposed wording for 9p1 is meant to be a blanket solution to all possible cases of posterior changes to results that result in fewer competitors being able to advance to the next round.

Currently, 9p1 states that "at least 25% of competitors must be eliminated between consecutive rounds of the same event."

Following this wording, when there's a scoretaking mistake or a decision on a disciplinary incident that results in fewer competitors being able to advance to a consecutive round, we always face a dilemma: do we remove the results of innocent competitors (preventing a violation of 9p1) or do we allow innocent competitors to keep their results, even if the obligatory elimination of at least 25% of competitors between rounds would've prevented them from competing in the following round (being guided by fair sportsmanship, as dictates 11d)?

With the proposed wording for Regulation 9p1, we do not have to make this choice!

Here's what my proposal looks like:

"9p1) At least 25% of competitors of competitors _with recorded results at the time the round ended_ must be eliminated between consecutive rounds of the same event."

Let's say a decision on a disciplinary incident forces us to remove the results of a competitor on the second round of an event, with 16 competitors on it.

Following the logic of the current wording of 9p1, this reduces the number of competitors that could advance to the third round. The competitor that placed 12th, not related to the disciplinary incident, would have to have his results removed from the third round of the event, because 9p1 dictates that at most 11 competitors can advance in a round where 15 competitors had competed in.

However, with the proposed wording for 9p1, the innocent competitor would be able to keep his results! This is because the competitor involved in the disciplinary incident, who had his results removed, did have recorded results at the time the round ended. His results on the second round were only removed afterwards.

This makes it so the competitor involved in the disciplinary incident counts towards the 25% obligatory elimination, even if his results were removed.

The proposed wording of 9p1 serves as a blanket solution to any similar situation, and gives competitors not related to these incidents assurance that their results won't be affected by a change in the results of another competitor.

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -258,6 +258,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 9m3) Rounds with 7 or fewer competitors must not have subsequent rounds.
 - 9o) Cutoff rounds count as one round when counting the number of rounds per event.
 - 9p) If an event has multiple rounds, then:
+    - 9p1) At least 25% of competitors that had results when the round ended must be eliminated between consecutive rounds of the same event.
     - 9p2) The competitors who advance to the next round must be determined by ranking (best x competitors) or by result (all competitors with a better result than x) in the preceding round.
         - 9p2a) For each round, advancement conditions must be announced before the round starts, and should not be changed after it has begun. Changes must be made at the discretion of the WCA Delegate, who must carefully consider the fairness of the change.
     - 9p3) If a qualifying competitor withdraws from a round, they may be replaced by the best-ranked non-qualifying competitor from the preceding round.
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -253,12 +253,11 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 - 9k) All competitors may participate in all events of a competition, except in cases specifically approved by the Board.
 - 9l) Each round must be completed before any following round of the same event can start.
 - 9m) Events must have at most four rounds.
-    - 9m1) Rounds with 99 or fewer competitors must have at most two subsequent rounds.
-    - 9m2) Rounds with 15 or fewer competitors must have at most one subsequent round.
+    - 9m1) Rounds with 99 or fewer competitors that had results when the round ended must have at most two subsequent rounds.
+    - 9m2) Rounds with 15 or fewer competitors that had results when the round ended must have at most one subsequent round.
     - 9m3) Rounds with 7 or fewer competitors must not have subsequent rounds.
 - 9o) Cutoff rounds count as one round when counting the number of rounds per event.
 - 9p) If an event has multiple rounds, then:
-    - 9p1) At least 25% of competitors with recorded results at the time the round ended must be eliminated between consecutive rounds of the same event.
     - 9p2) The competitors who advance to the next round must be determined by ranking (best x competitors) or by result (all competitors with a better result than x) in the preceding round.
         - 9p2a) For each round, advancement conditions must be announced before the round starts, and should not be changed after it has begun. Changes must be made at the discretion of the WCA Delegate, who must carefully consider the fairness of the change.
     - 9p3) If a qualifying competitor withdraws from a round, they may be replaced by the best-ranked non-qualifying competitor from the preceding round.
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -258,7 +258,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 9m3) Rounds with 7 or fewer competitors must not have subsequent rounds.
 - 9o) Cutoff rounds count as one round when counting the number of rounds per event.
 - 9p) If an event has multiple rounds, then:
-    - 9p1) At least 25% of competitors must be eliminated between consecutive rounds of the same event.
+    - 9p1) At least 25% of competitors with recorded results at the time the round ended must be eliminated between consecutive rounds of the same event.
     - 9p2) The competitors who advance to the next round must be determined by ranking (best x competitors) or by result (all competitors with a better result than x) in the preceding round.
         - 9p2a) For each round, advancement conditions must be announced before the round starts, and should not be changed after it has begun. Changes must be made at the discretion of the WCA Delegate, who must carefully consider the fairness of the change.
     - 9p3) If a qualifying competitor withdraws from a round, they may be replaced by the best-ranked non-qualifying competitor from the preceding round.


Goosly:
It's not really clear what the new 9p1 is supposed to mean. Wouldn't it be more clear to leave 9p1 as it is, and add a guideline to thoroughly explain what should happen if there are mistakes during scoretaking, found after the subsequent round has started (or ended)?

mirandapedro:
Hi Manu.

The proposed wording for 9p1 is meant to be a blanket solution to all possible cases of posterior changes to results that result in fewer competitors being able to advance to the next round.

Besides being generally opposed to the unnecessary use of the Guidelines, the Regulations should be a complete document, and therefore it's mistakes should not be fixed by adding Guidelines.

Furthermore, I believe a detailed explanation of how to proceed in a specific situation would solve the situation described, but no alll the other situations that are similar to it.

This case is a great example of this. It was initially originated not from a scoretaking mistake, as you mentioned, but from a disciplinary incident solved by the WDC. If I had set out to only solve one situation or the other, we'd be left without answers when a situation not described in the Guidelines came up.

Instead, I tried to come up with a blanket solution to all similar situations, and here's how the wording I proposed does that.

Currently, 9p1 states that "at least 25% of competitors must be eliminated between consecutive rounds of the same event."

Following this wording, when there's a scoretaking mistake or a decision on a disciplinary incident that results in fewer competitors being able to advance to a consecutive round, we always face a dilemma: do we remove the results of innocent competitors (preventing a violation of 9p1) or do we allow innocent competitors to keep their results, even if the obligatory elimination of at least 25% of competitors between rounds would've prevented them from competing in the following round (being guided by fair sportsmanship, as dictates 11d)?

With the proposed wording for Regulation 9p1, we do not have to make this choice!

Here's what my proposal looks like:

"9p1) At least 25% of competitors of competitors _with recorded results at the time the round ended_ must be eliminated between consecutive rounds of the same event."

Let's say a decision on a disciplinary incident forces us to remove the results of a competitor on the second round of an event, with 16 competitors on it.

Following the logic of the current wording of 9p1, this reduces the number of competitors that could advance to the third round. The competitor that placed 12th, not related to the disciplinary incident, would have to have his results removed from the third round of the event, because 9p1 dictates that at most 11 competitors can advance in a round where 15 competitors had competed in.

However, with the proposed wording for 9p1, the innocent competitor would be able to keep his results! This is because the competitor involved in the disciplinary incident, who had his results removed, did have recorded results at the time the round ended. His results on the second round were only removed afterwards.

This makes it so the competitor involved in the disciplinary incident counts towards the 25% obligatory elimination, even if his results were removed.

The proposed wording of 9p1 serves as a blanket solution to any similar situation, and gives competitors not related to these incidents assurance that their results won't be affected by a change in the results of another competitor.

I hope I've been more clear in explaining the impact of the proposed change to Regulation 9p1. Thanks for pointing the relative obscurity with which I initially presented the proposal.

Feel free to make any further comments you feel would contribute to the solution of this issue.

Jambrose777:
Should 9m also be modified for the same reasons?

If Round 1 has 16 people, Round 2 has 8, and Round 3 has 6, and a competitor that made round 2 was disqualified from Round 1. Then the competitor will have DNF results in Round 1, removed results in round 2, causing Round 2 to not be able to have a subsequent round. So Round 3 would have to be removed.

Goosly:
Yeah, I understand your explanation and the reasoning.
And the fact that you need so many words to explain the small change, shows that it does indeed need a more thorough explanation than the current proposal! :-) The change is simply too short to be understandable.

Maybe leave 9p1 as it is and add something that starts with: "In case of posterior changes to results that result in fewer competitors advancing, then ... _<explain what should happen, e.g. don't punish innocent competitor by removing their results>_

Ivan-Brigidano (commented):
Not fully sure is "with recorded results at the time the round ended" is enough to reflect the ruling. If the general ruling is:
- Keep the results if the competitor proceeded thanks to an error in someone else's results
- Remove the results if the competitor proceeded thanks to an error in their own result

This modification would allow the competitor of the second case to maintain their own results.

I like the idea of @Goosly of creating a Guideline.

mirandapedro:
> This modification would allow the competitor of the second case to maintain their own results.

The proposed wording would not allow competitors who advanced due to an error in their own results to keep their times.

This is because when their results are removed, the removal is not based on Regulation 9p1, but in many others, designed for specific scenarios. Just to mention a few:

- 2k2: a competitor may be disqualified from some events if they are suspected of cheating or defrauding the officials during the competition;
- 2k3: a competitor may be disqualified from some events they behave in a way that is unlawful/violent/indecent, or intentionally damages venue facilities or personal property within the venue.
- 2k5: a competitor may be disqualified from some events if they fail to abide by WCA Regulations during the competition;
- 9p2: The results of a competitor may be removed if it is posteriorly clear he did not meet the requirements to have competed in a round.

Therefore whenever there is an l incident and we remove a competitor results in a round, we do it based on various articles of the Regulation, and not on Regulation 9p1.

Allowing competitors that had results removed after the competition to count towards the 25% obligatory elimination quota does not change this.

@Jambrose777 perhaps a change to 9m is also indeed necessary. I'll look into it.

xsrvmy:
I believe we should apply the same fix to 9m.

lgarron:
Thanks for taking the time to start this!

While we have an idea of what we want to achieve, we'll have to be careful with the wording.

For example, we need to make sure that something like "recorded results at the time the round ended" can't be abused if results are misplaced or ignored.

mirandapedro:
I've discussed with @Ivan-Brigidano and improved the wording of this change!

To the Regulations, a competitor has a result after the judge has recorded the result in the scoresheet and both the competitor and judge have signed on it (Regulation A7). 

If competitor has completed this steps, they have results, even if they are not yet on WCA Live, if they were missplaced or missing.

 I changed "with recorded results" to "that had results", because it leaves no room for doubts or loopholes, as you mentioned @lgarron.

Ivan-Brigidano (approved):
I'm happy with the wording now. But we should await for a double check by other WRC members just in case.

mirandapedro:
Does anyone else from the @thewca/wrc-team have some thoughts on this?

Ivan-Brigidano:
See #997

PR #977, opened 2020-05-11
Count individual puzzles as unsolved in Multi-Blind accidents

Ivan-Brigidano:
Fixes [this Incident Log](https://www.worldcubeassociation.org/incidents/26) and [this WCA Forum thread](https://forum.worldcubeassociation.org/t/disqualify-individual-multi-blind-puzzles-if-a-competitor-makes-a-move/3142).

@Claster you brought some concerns in the WCA Forum thread. Are you happy with the wording?


xsrvmy:
I think it's possible to interpret this wording as making 2 moves = DNF. Don't think that's a good idea because accidental slice moves can probably happen.

wca-guidelines.md
```diff
@@ -222,3 +222,4 @@ To be more informative, each Guideline is classified using one of the following
 - H1b1++) [ADDITION] The judge may permit the competitor to continue the attempt unofficially, but the attempt must be stopped and judged first, in full accordance with the Regulations.
 - H1d+) [EXAMPLE] Example: Suppose a competitor attempts 10 cubes, stops the solve with a time of 59:57, and has two time penalties. The time of the final result is 59:57 + 2*2 = 60:01 (also see [Regulation A1a5](regulations:regulation:A1a5)).
 - H1d++) [EXAMPLE] Example: Suppose a competitor attempts 10 cubes, the judge stops the competitor at 60 minutes, and the attempt has two time penalties. The time of the final result is 60:00 + 2*2 = 60:04.
+- H1d+++) [CLARIFICATION] If a competitor accidentally applies a move to a single puzzle during the memorization phase, the puzzle may be considered unsolved instead of the whole attempt, at the discretion of the WCA Delegate.
```
lgarron: Nit: it's not quite the attempt that is unsolved
```suggestion
- H1d+++) [CLARIFICATION] If a competitor accidentally applies any moves to one or more puzzles during the memorization phase, those puzzles may be considered unsolved at the end of the attempt (instead of immediately disqualifying the entire attempt), at the discretion of the WCA Delegate.
```

lgarron (approved):
Thanks for catching that we need to update this!

mirandapedro (commented):
I believe this should be in the Regulations instead of the Guidelines.

It does not make sense to have it as a clarification of Regulation H1d, as it refers to time penalties, not disqualifications.

If we want the Regulations to be "considered a complete document", we should not add articles to the Guidelines that are complete innovations.

Ivan-Brigidano:
> I believe this should be in the Regulations instead of the Guidelines.
> 
> It does not make sense to have it as a clarification of Regulation H1d, as it refers to time penalties, not disqualifications.
> 
> If we want the Regulations to be "considered a complete document", we should not add articles to the Guidelines that are complete innovations.

Do you think it would be better as a Regulation H1e? If so, I can easily make the changes.

mirandapedro:
Yes, I believe adding it as Regulation H1e is more fitting.

mirandapedro (approved):

Claster:
Looks good to me, thanks for this change!

PR #978, opened 2020-05-11
Rename Article H of the Guidelines

Ivan-Brigidano:
It doesn't make much sense to have a different name than the Regulations.


mirandapedro (approved):
Good catch!

Looks like they missed this on [commit 74424b5](https://github.com/thewca/wca-regulations/commit/74424b5c45819f36ef25504a95a56109927a6cf8).

Looks good to me! You can go ahead with merging.

lgarron (approved):

PR #980, opened 2020-07-20
Merge 'mirandapedro-temporary-regulations-covid-19' into 'official' - version 2020-08-01.

mirandapedro:
This amendment was written in light of the WCA Competition Safety Policy and the COVID-19 pandemic, and are meant to help the WCA in hosting safe competitions with as low risk as possible.

Here's a summary of the included changes:

- A new article Y is added to the Regulations which complements the Safety Policy. It is expected that the changes will be reverted once it is safe to do so.
- WCA Delegates can impose additional requirements on attendees like wearing masks or not having COVID symptoms, if it is necessary and if it is clearly announced before the competition (Regulations Y3 and Y3a through Y3d).
- The competitor resets the timer instead of the judge. The judge reminds the competitor to do so (Regulation Y7d1).
- Competitors who behave unsafely might be disqualified by the Delegate (Regulation 2k3) and requested to leave the venue (Regulation 11h).

Some additions that make the Regulations more inline with the Safety Policy:
- Competitors should bring their own pens (Regulation Y4), judges may bring their own stopwatches (Regulation Y5).
- Competitors, without touching the scoresheet, must sign the result recorded by their judge using their own pen (Regulation Y7e1).
- There should be fixed seats and there shouldn’t be a Competitor Area if unnecessary (Regulations Y6a, Y7a).
- The competitor places the puzzle inside the cover after the attempt (Regulation Y7f).

### DIFF
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -1,6 +1,6 @@
 # <wca-title>WCA Guidelines
 
-<version>Version: January 1, 2020
+<version>Version: August 1, 2020
 
 
 ## Notes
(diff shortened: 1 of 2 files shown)


lgarron (approved):

mirandapedro:
Shouldn't this branch be deleted?

PR #981, opened 2020-07-20
Merge 'temporary-regulations-covid-19-draft' into 'draft' - version 2020-08-01

Ivan-Brigidano:
This PR intends to add the changes of the Temporary Regulations amendment to `draft`. This can't be merged automatically and require solving some conflicts.

I've made it from a new branch that have all the changes from `mirandapredro-temporary-regulations-covid-19`. This way we can commit new changes into this branch without any problems.

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -24,7 +24,7 @@ Original source of the WCA regulations: [www.worldcubeassociation.org/regulation
 WCA Regulations in [PDF format](link:pdf)
 
 ### Source
-Development of the WCA Regulations and Guidelines is public [on GitHub](https://github.com/thewca/wca-regulations).
+Development of the WCA Regulations and Guidelines is public [on GitHub](https://github.com/thewca/wca-regulations) and the discussion is public [on the WCA Forum](https://forum.worldcubeassociation.org/c/regulations).
 
 ### Contact
 For questions and feedback, please contact the [WCA Regulations Committee (WRC)](mailto:wrc@worldcubeassociation.org).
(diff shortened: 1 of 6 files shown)


Ivan-Brigidano:
I've solved all the conflicts so I should be ready to be merged. But please, whoever reviews it, make sure that I have not accidentally removed any changes that are included in #980 (and that I didn't remove any commits for 2021 Regulations).

wca-regulations.md
```diff
@@ -297,7 +297,8 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 11e2) If a competitor is granted an extra attempt, the extra attempt should be done right after the attempt that caused it, and must replace the original regularly numbered attempt.
     - 11e3) If it is unclear whether an incident should result in an extra attempt, the competitor may be granted a provisional extra attempt that will be used only if it is later found that an extra attempt was appropriate (e.g. by a decision of the WRC).
 - 11f) Decisions about an incident may be supported with video or photographic analysis, at the discretion of the WCA Delegate.
-- 11g) The WCA Delegate must ensure that copies of the Regulations and Guidelines are available (e.g. printed, digital, or accessible via internet) to officials and competitors for consultation on any incidents.
+- 11g) The WCA Delegate must ensure that copies of the Regulations and Guidelines are available (e.g. printed, digital, or accessible via internet) to consult for any incidents.
```
```suggestion
- 11g) The WCA Delegate must ensure that copies of the Regulations and Guidelines are available (e.g. printed, digital, or accessible via internet) to officials and competitors for consultation on any incidents.
```

xsrvmy:
There might be an unintended issue in A7e1: The word "must" also applies to the recommendation of using the competitor's own pen, implying that the result is invalid if the competitor signs with the judge's pen. This is probably not intended, since Y4 uses "should". I think it makes more sense to make the safety recommendation in a separate sentence: "The competitor should use their own pen."
Similar issues in A7f: a safety recommendation is worded with "must".

Ivan-Brigidano:
> There might be an unintended issue in A7e1: The word "must" also applies to the recommendation of using the competitor's own pen, implying that the result is invalid if the competitor signs with the judge's pen. This is probably not intended, since Y4 uses "should". I think it makes more sense to make the safety recommendation in a separate sentence: "The competitor should use their own pen."

For the first case I assume that you are referring to Y7e1 and not A7e1 (which doesn't exist). I don't think there is any problem with current wording because the organizing team could give a pen to each competitor on the registration and ask them to always use that pen (which is what was done at Lonely Denmark 2020).

> Similar issues in A7f: a safety recommendation is worded with "must".

And for the second case I guess that you are actually referring to Y7f instead of A7f. I don't think there is any problem with using a "must" at all.

In any case you can give your feedback on the [Forum topic](https://forum.worldcubeassociation.org/t/wca-regulations-temporary-regulations/7346). But this version is the one approved by the Board so there is no intention to introduce big changes unless a big error is spotted.

mirandapedro (approved):
LGTM. I think @lgarron should also take a look at this to make sure everything is fine.

lgarron (approved):

Issue #534, opened 2018-01-13
Regulation A6e) Revisement


Jambrose777:
> A6e) After releasing the puzzle, the competitor must not touch or move the puzzle until the judge has inspected the puzzle. Penalty: disqualification of the attempt (DNF). Exception: If no moves have been applied, a time penalty (+2 seconds) may be assigned instead, at the discretion of the judge.

There are 2 issues with this regulation:

1) "until the judge has inspected the puzzle" is very ambiguous. 
When does the judge inspect the puzzle? Are they able to inspect the puzzle without seeing all sides? How long does it take? When can the competitor be sure that the judge has inspected the puzzle? This creates unfair solving conditions as some judges may be more attentive than others and can inspect the puzzle faster than others.
Example: At CubingUSA Nationals 2017 3x3 Finals, Lucas Etter's solve 4: https://www.youtube.com/watch?v=YSo8w1yPetE&feature=youtu.be&t=1h22m12s the judge stated that he inspected the puzzle in the half second before the competitor touched the puzzle, no penalty was awarded for A6e.

One possible soultion would be to change the regulation to until the judge has _signed the scorecard_.

2) "If no moves have been applied, a time penalty (+2 seconds) may be assigned instead"
This regulation should read "If no moves have been applied, a time penalty (+2 seconds) may be assigned **instead of a DNF**"
_Many_ delegates have misinterpreted this regulation in this way and award no penalty because the +2 may be awarded.

xsrvmy:
1. I believe that "inspected" the puzzle refers to the judge telling the competitor that the cube is "OK" (A7a). However, that regulation is often not followed unless the result is a +2 or if the event is BLD.\
I think the best way to handle this is to allow the competitor to touch the puzzle after the judge starts recording the time, after he calls "OK", or after he explicitly allows the competitor to do so (in case of an incident).
2. The fix is definitely needed. Also, there appears to be ambiguity in what the phrase "at the discretion of the judge" modifies - Does it modify the entire sentence, or just which penalty to assign? The latter interpretation does not seem correct. Perhaps "at the discretion of the judge" should be moved to the start of the sentence.

julesDesjardin:
1. Clarifying this can indeed be a good way to ensure fairness and consistency between judges (I personally would have given Lucas a +2 without any doubt since I am quite strict on this rule).
[...]

Issue #23, opened 2013-02-08
A3c2) "limits 10f"


lgarron:

lgarron:
Fixed in b3133530e1969464c54b35420080aa038a185bec.

PR #983, opened 2020-08-06
Update Regulation A6e to WRC understanding

### LINKED ISSUE #200: Picking up the puzzle before the end of the attempt. (A6e)
In [A6e](https://www.worldcubeassociation.org/regulations/#A6e), the phrase "until the judge has inspected the puzzle" is not clear.

> A6e) The competitor must not touch or move the puzzle until the judge has inspected the puzzle. Penalty: disqualification of the attempt (DNF). Exception: If no moves have been applied, a time penalty (+2 seconds) may be assigned instead, at the discretion of the judge.

Discussion on Delegates List (in the US Nationals 2014 thread): https://groups.google.com/forum/#!topic/wca-delegates/g1WhtgfSbrs

mirandapedro:
Update Regulation A6e to WRC understanding by specifying the requirements for lifting the time penalty for touching the puzzle after releasing it.

Fixes #200, #534, [this](https://forum.worldcubeassociation.org/t/regulation-a6e-revisement/3225/3) Forum post and makes the adjustment mentioned on Incident Log [#23](https://www.worldcubeassociation.org/incidents/23).

There are three main issues that should be addressed on this update:
1. When can the competitor touch the puzzle after stopping the timer - the current wording "until the judge has inspected the puzzle" is not very clear;
2. On what circumstances can the time penalty for touching the puzzle after stopping the timer be lifted; and
3. Whether a puzzle falling on the competitor's lap should still be penalized - currently made clear by Guideline A6e+.

The solution for each one of these issues is:

1. Making it so the competitor can only touch the puzzle after the attempt ends (after the competitor has signed the scoresheet, as per Regulation A7c) or after being explicitly allowed to do so by the judge.

2. Instituting the following three requirements for lifting the time penalty for touching the puzzle after releasing it (also present in the above mentioned Incident Log):
- The touch must happen immediately after stopping the timer - this Regulation aims at being lenient toward instinctual or automatic reactions (like the one made on [this solve](https://www.youtube.com/watch?v=YSo8w1yPetE&feature=youtu.be&t=1h22m12s) by Lucas Etter), and not touch to the puzzle that happens when the competitor is calm or can think things through;
- The touch must be brief - for the same reason as above, a prolonged touch to the puzzle constitutes a clear violation of Regulation A6e; and
- The touch must not affect the state of the puzzle - if it does, the attempt should be disqualified, as per the proposed Regulation A6e1. To be able to determine whether this happened or not, the WCA Delegate may use video evidence, as per Regulation 11f.

This commit also adds the explicit need for WCA Delegate discretion for applying the exception, as judges shouldn't be the ones making these decisions.

I believe an update do A6e shouldn't get into the merit of the touch being intentional or not. Besides it being subjective to determine intention (you could argue, for example, that Lucas' touch in the above-linked video was intentional), it isn't the best criteria - sometimes an unintentional touch should be penalized (e.g. a puzzle falling on a competitor's lap) and an "intentional" (or instinctual) touch shouldn't.

3. As a puzzle falling on a competitor's lap does not configure a brief touch - thus not meeting the requirements defined in Regulation A6e2 - it should still be penalized. That's why Guideline A6e+ should still stand as it is.

Finally, there shouldn't be any discretion in determining whether a violation of Regulation A6e results in a DNF or a +2. That's why these two situations were treated separately and made different.


xsrvmy:
Wording problem: the time penalty *may not* be applied - "may not" is not in RFC2119. Probably should be "should not" or "must not".

---

Rest is comment on the proposal itself:

Is there a reason why the WRC has decided to make A6e stricter when almost everyone thinks its unnecessarily strict as it is? Wasn't there a proposal to loosen A6e so that the +2 is not given as long as there is no potential misalignment penalty that need to be resolved? Why isn't allowing the competitor to touch the puzzle after the judge signs enough?

I think I need to voice a pretty strong objection to the first criteria:
There are at least three common actions that can lead to the puzzle being accidentally touched: reactions, reaching for the camera, signing (this one is new). The first criteria does not cover the third case at all, and only partially covers the second case. The competitor might also accidentally make contact with the puzzle while the judge is holding it.

Here is an example where the cube is touched when picking up the camera (former 6x6 WR): https://www.youtube.com/watch?v=RrSGmPKRdr4

Also two questions arise about signing:
- What if the competitor signs, and while signing, touches the puzzle? The regs do not allow adding a +2 to the scoresheet at this point.
- Does this void A7c4+?

KitClement:
Is there any reason the new A6e2 is at the discretion of the delegate? The delegate is typically not the one that will be viewing the solve in the first place, it seems more reasonable to have this at the discretion of the judge.

wca-regulations.md
```diff
@@ -405,7 +405,9 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
...
     - A6c) The competitor must fully release the puzzle before stopping the timer. Penalty: disqualification of the attempt (DNF). Exception: If no moves have been applied after stopping the timer and before releasing the puzzle, a time penalty (+2 seconds) may be assigned instead, at the discretion of the judge.
     - A6d) The competitor must stop the timer using both hands, placed flat on the sensors with palms down. Penalty: time penalty (+2 seconds).
-    - A6e) After releasing the puzzle, the competitor must not touch or move the puzzle until the judge has inspected the puzzle. Penalty: disqualification of the attempt (DNF). Exception: If no moves have been applied, a time penalty (+2 seconds) may be assigned instead, at the discretion of the judge.
+    - A6e) After stopping the timer, the competitor must not touch or move the puzzle until the attempt has ended (see [Regulation A7c](regulations:regulation:A7c)) or after the judge has explicitly allowed them to do so.
+            - A6e1) Penalty for moving the puzzle after stopping the timer: disqualification of the attempt (DNF).
```
SAuroux: I am certain that _moving the puzzle_ is not a valid synonym for _applying moves to the puzzle_.

SAuroux:
I see a lot of issues with this proposal and I really do not think that the current wording is practically applicable at all. Most of all, changing the scope of this regulation from _until the puzzle has been inspected_ to _until the end of the attempt_ is a significant scope extension that introduces a lot of additional strictness which I believe to be unnecessary. 

Especially, I do not think that all implications have been thought through well enough. For instance:
* Pushing aside the puzzle to create space on the mat for signing as a competitor is now a DNF according to the current wording of A6e1 as the puzzle has been _moved_ out of place. But also after fixing the wording error in A6e1 to _applying moves to the puzzle_, the same action is still a clear +2 according to A6e2. Surely, this cannot be intended.
* Situation being absolutely common these days would now be clear DNFs, e.g.:
  * A competitor aligning their puzzle from +2 state to solved state before signing the result (with +2) already correctly recorded and signed by the judge
  * A competitor swiftly practicing a previously messed up PLL before signing a result already correctly recorded and signed by the judge

For those reasons and also to prevent adding unnecessary strictness, I think that the regulation's scope should not be extended. If anything, it should rather be narrowed. Two weeks ago I shared my opinion in the corresponding [WCA form thread](https://forum.worldcubeassociation.org/t/regulation-a6e-revisement/3225/7?u=sebastien_auroux) and questioned where is the sense in applying a +2 penalty if it is already clear that the final state of the puzzle has not been affected? 

I have always been convinced that the ultimate intention behind regulation A6e is to ensure that if a competitor prevents the judge from inspecting the final state of the puzzle, they get DNF'd. Isn't this what we actually want? If yes, I really do not think that so much has to be changed starting from the existing regulation. Here is my attempt:

> A6e) After releasing the puzzle and stopping the timer, the competitor should not interact with the puzzle until the judge has inspected the puzzle and determined its final state. Penalty for any interaction of the competitor with the puzzle due to which the judge is unable to capture the final state of the puzzle when the timer was stopped: disqualification of the attempt (DNF), at the discretion of the judge.

EdHollingdale:
I completely agree with @SAuroux here and I like the wording he has proposed. Extending the regulation in this way adds more unnecessary strictness, and the +2 rule is confusing.

I also don't get the +2 rule and exception - it feels like it has been worded to not change too much while fitting with prior WRC rulings. In what instances would the +2 rule be applied? If the judge thinks there was a move applied then it is DNF, if there was no move applied and the competitor touched after stopping the solve then it is most likely no penalty. Or at least I would say that if not for the "immediately" which is why it like it is just there to fix a noteworthy WRC decision. 

Applying a few real world tests should show the weakness of the proposed change: what if I brush the cube while reaching to sign the scorecard - is that at +2 now? What if I forget to sign the scorecard and the score-taker finds me later, is that an automatic +2? Or did I touch the cube "immediately" after enough for it to be covered by the exception? In which case, why even have the +2 as a rule?

mirandapedro:
Thank you to everyone who contributed to this PR!

Ivan-Brigidano (approved):

PR #982, opened 2020-08-01
fix inconsistent link

oyyq99999:

PR #984, opened 2020-08-07
Fix inconsistent link in A7c

Ivan-Brigidano:
This is the equivalent of #982 for the `draft` branch.


mirandapedro:
Just saw this today! Thanks for handling this, Iván!

My only concern is that by doing this, we'd be rewriting the history of the `draft` branch and eventually rewriting the history of the `official` branch itself once we have a new Regulations ammendment.

I think the most appropriate way to handle this would be to create a PR from `official` (with the incorrect link already fixed) to `draft`. From our [wiki page](https://github.com/thewca/wca-regulations/wiki/On-submitting-pull-requests):

> If your pull request adresses only minor changes, without an influence of the Regulations meaning, then it is known as a textual revision, and should be submitted using `official` as the base branch.
> [...]
> We need to make textual revisions against `official` because once the changes are merged into `official`, the WRC will merge `official` into `draft` to port your changes to the current work-in-progress version.
> It's necessary, as doing it the other way around (merging first in `draft`, then backporting to `official`) would also bring all the current major changes from `draft` to `official`, which is definitely not something we want.
> Other options (such as cherry-picking) are not viable, as we want `draft` and `official` to share the same history (and not rewrite it!).

@lgarron could you please confirm this?

mirandapedro (changes requested):
Just saw this today! Thanks for handling this, Iván!

My only concern is that by doing this, we'd be rewriting the history of the `draft` branch and eventually rewriting the history of the `official` branch itself once we have a new Regulations ammendment.

I think the most appropriate way to handle this would be to create a PR from `official` (with the incorrect link already fixed) to `draft`. From our [wiki page](https://github.com/thewca/wca-regulations/wiki/On-submitting-pull-requests):

> If your pull request adresses only minor changes, without an influence of the Regulations meaning, then it is known as a textual revision, and should be submitted using `official` as the base branch.
> [...]
> We need to make textual revisions against `official` because once the changes are merged into `official`, the WRC will merge `official` into `draft` to port your changes to the current work-in-progress version.
> It's necessary, as doing it the other way around (merging first in `draft`, then backporting to `official`) would also bring all the current major changes from `draft` to `official`, which is definitely not something we want.
> Other options (such as cherry-picking) are not viable, as we want `draft` and `official` to share the same history (and not rewrite it!).

@lgarron could you please confirm this?

It's also good to point that there probably won't be any conflicts on merging `official` into `draft`, so this should be really easy.

lgarron:
> @lgarron could you please confirm this?
> 
> It's also good to point that there probably won't be any conflicts on merging `official` into `draft`, so this should be really easy.

Yeah, It's not completely trivial, `git` seems to think think there's a merge conflict, probably because of previous merge directions.

I think our best options are:

- Consider 023c899f16f880b5747206d85e5ec73e2b6d3f39 (the change from #982) a clerical error and delete it from `official` (rewrite the history by rewinding one commit).
- Merge `official` into `draft`.

If we don't want to avoid rewriting history (generally not a good idea):

- Revert 023c899f16f880b5747206d85e5ec73e2b6d3f39 (the change from #982) via #987.
- Merge `official` into `draft` via #989.

If I were leader, I'd probably just make a decision, but I can follow through with either (assuming #987 and #989 are approved).

Ivan-Brigidano:
Thanks you for handling this, Lucas.
I've just approved both PR. You are the git expert and I rely on you. So please, proceed merging the PRs as needed.

lgarron:
> Thanks you for handling this, Lucas.
> I've just approved both PR. You are the git expert and I rely on you. So please, proceed merging the PRs as needed.

Thanks!

I think we should be good now. `official` matches the release from last August, and everything from `official` is merged into `draft`.

lgarron:
You'll probably need to rebase this or open a new version based on the latest `draft`.

Ivan-Brigidano:
@lgarron can you please approve and rebase this PR? (Or create a new PR, whatever you think it's better for the commit history)

PR #993, opened 2021-02-26
Explicitly allow the use of correction fluid in FMC

Rouxles:
Forum Link: https://forum.worldcubeassociation.org/t/making-e3a-less-strict/1297

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -492,6 +492,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - E3c) Stickers (self-supplied).
     - E3d) Stopwatch or watch (self-supplied) for keeping track of the elapsed time, if it is approved by the WCA Delegate.
     - E3e) Other non-electronic aids that do not give an unfair advantage, at the discretion of the WCA Delegate (see [Regulation 2i1](regulations:regulation:2i1)).
+    - E3f) Correction fluid/tape or similar (self-supplied).
 - E4) The WCA Delegate may choose to publish the solutions (e.g. photo of the score sheets or transcriptions of solutions).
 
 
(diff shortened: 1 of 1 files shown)


Ivan-Brigidano:
I'm fine with using a new Regulation. Including it on E3e also sounds good to me.

Ivan-Brigidano (approved):

lgarron (commented):
Would it make sense to include erasers as an example, as suggested in the forum post?

We already mention pencils directly in E3a.

lgarron (changes requested):
I don't mean to block this change (feel free to dismiss in the UI), but marking it as "changes requested" to discuss whether to include eraser.

Ivan-Brigidano:
> I don't mean to block this change (feel free to dismiss in the UI), but marking it as "changes requested" to discuss whether to include eraser.

Given that pencils are allowed, I think allowing erasers is perfectly fine.

Jambrose777:
Should E3f1 instead be a guideline? All other example regulations are placed in the guidelines.

lgarron:
> Should E3f1 instead be a guideline? All other example regulations are placed in the guidelines.

Good point. We usually use "(e.g. ...)" for examples in the Regs. If someone can think of a way to do this inlein without two parentheses next to each other, I think that'd b ebest!

wca-regulations.md
```diff
@@ -492,6 +492,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - E3c) Stickers (self-supplied).
     - E3d) Stopwatch or watch (self-supplied) for keeping track of the elapsed time, if it is approved by the WCA Delegate.
     - E3e) Other non-electronic aids that do not give an unfair advantage, at the discretion of the WCA Delegate (see [Regulation 2i1](regulations:regulation:2i1)).
+    - E3f) Tools to remove markings (self-supplied) such as correction fluid, correction tape, or erasers.
```
Rouxles: @lgarron @Jambrose777 Do you think this wording works fine? I think it solves Lucas' issue (although this wording doesn't exist in the regulations).

IMO, it's important enough to make clear what these tools are in the regulations rather than the guidelines for clarity, so I think using a completely new wording rather than the guidelines is important enough here.

lgarron: I think this works well!

I'd add a tiny comma, although it's not strictly needed:
```suggestion
    - E3f) Tools to remove markings (self-supplied), such as correction fluid, correction tape, or erasers.
```

lgarron (approved):

wca-regulations.md
Rouxles: I think it looks better with the comma - I'll commit the suggestion

Issue #58, opened 2013-08-03
5b3b: What about big cube defects?


sarahstrong314:
If a big cube pop involves two corners, this might result in CP parity, which requires more than 4 pieces to be taken out to fix. 

Suggestion: Allow 8 pieces for 4x4x4,  11 pieces for 5x5x5, 14 for 6x6x6, and 17 for 7x7x7. Or limit how many pieces are allowed to pop (DNF if >N pieces pop out) so you can't get CP parity when the puzzle is being repaired.

lgarron:
Question: how much of this can be solved by leaving it at the judge's/Delegate's discretion? This goes under the umbrella of "how do we avoid hurting competitors who aren't trying to cheat, while still preventing cheating". (Or replace "cheating" with "gaining an unfair advantage".

lgarron:
@Sa967St: If you'd like to do something about this, could you take the initiative?

SAuroux:
Keep it simple...

## Proposal

5b3b) If, after repairing the puzzle but before the end of the attempt, the competitor finds that the puzzle is unsolvable, he may disassemble and reassemble a reasonable amount of pieces to make the puzzle solvable, at the discretion of the delegate.

lgarron:
How about at the discretion of the judge?

Making this a Delegate responsibility is a bit weird. Are you supposed to
stop and call over the Delegate before proceeding?

»Lucas Garron

SAuroux:
no, you're right. just replace "delegate" by "judge".

2013/12/11 Lucas Garron notifications@github.com

Laura-O:
This hasn't been updated in nearly 4 years.
Closing this for now as this has obviously never been a problem in practice.

PR #996, opened 2021-03-01
Clarify how to handle results of disqualified attempts

mirandapedro:
This PR change clarifies how should the results of disqualified attempts be handled.

It comes as a result of [incident log #58](https://www.worldcubeassociation.org/incidents/58), in which a competitor was disqualified at the start of a round before having finished any attempts. In alignment with Regulation 2j1a, the Delegate recorded all of his attempts as DNFs.

However, this artificially increases the number of competitors in a round (see Regulation 9m1-9m3) and misrepresents what happened in practice.

The proposed wording for 2j1a makes it clear that the rest of the remaining attempts should only be recorded as DNFs if the competitor has finished one or more attempts. Otherwise, if the competitor has finished no attempts, as is the case in the aforementioned incident log, no results should be recorded.

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -88,7 +88,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - 2i2b) The competitor must not interact with (e.g. operate, hold, wear) any active camera. Exception: the competitor may wear a camera mounted on their head, as long as it is out of their sight and it is clear that they are not interacting with it (apart from wearing it).
 - 2j) The WCA Delegate may disqualify a competitor from a specific event.
     - 2j1) If a competitor is disqualified from an event for any reason, they are not eligible for any more attempts in the event.
-        - 2j1a) If the competitor had already started at least one attempt in a round before being disqualified, the results of all remaining attempts in the event are recorded as DNF. If the competitor had started no attempts, no results should be recorded.
+        - 2j1a) If the competitor had already started at least one attempt in a round before being disqualified, the results of all remaining attempts in the event are recorded as DNF. If the competitor had started no attempts, no results are recorded.
     - 2j2) If a competitor is disqualified during the course of an event, their earlier results remain valid. Exception: cheating or defrauding (see [Regulation 2k2a](regulations:regulation:2k2a)).
 - 2k) At the discretion of the WCA Delegate, a competitor may be disqualified from some events (a single event, multiple events, or all events) if the competitor:
     - 2k1) Fails to check in or register in time for the competition.
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -88,7 +88,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - 2i2b) The competitor must not interact with (e.g. operate, hold, wear) any active camera. Exception: the competitor may wear a camera mounted on their head, as long as it is out of their sight and it is clear that they are not interacting with it (apart from wearing it).
 - 2j) The WCA Delegate may disqualify a competitor from a specific event.
     - 2j1) If a competitor is disqualified from an event for any reason, they are not eligible for any more attempts in the event.
-        - 2j1a) The results of all remaining attempts in the event are recorded as DNF.
+        - 2j1a) If the competitor has already finished one or more attempts in a round after being disqualified, the results of all remaining attempts in the event are recorded as DNF. If the competitor has finished no attempts, no results should be recorded.
     - 2j2) If a competitor is disqualified during the course of an event, their earlier results remain valid. Exception: cheating or defrauding (see [Regulation 2k2a](regulations:regulation:2k2a)).
 - 2k) At the discretion of the WCA Delegate, a competitor may be disqualified from some events (a single event, multiple events, or all events) if the competitor:
     - 2k1) Fails to check in or register in time for the competition.
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -88,7 +88,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - 2i2b) The competitor must not interact with (e.g. operate, hold, wear) any active camera. Exception: the competitor may wear a camera mounted on their head, as long as it is out of their sight and it is clear that they are not interacting with it (apart from wearing it).
 - 2j) The WCA Delegate may disqualify a competitor from a specific event.
     - 2j1) If a competitor is disqualified from an event for any reason, they are not eligible for any more attempts in the event.
-        - 2j1a) If the competitor had already started at least one attempt in a round before being disqualified, the results of all remaining attempts in the event are recorded as DNF. If the competitor had started no attempts, no results are recorded.
+        - 2j1a) If the competitor has already started at least one attempt in a round before being disqualified from it, the results of all remaining attempts in the event are recorded as DNF. If the competitor has started no attempts (i.e. has no results or only has DNS results), no results are recorded.
     - 2j2) If a competitor is disqualified during the course of an event, their earlier results remain valid. Exception: cheating or defrauding (see [Regulation 2k2a](regulations:regulation:2k2a)).
 - 2k) At the discretion of the WCA Delegate, a competitor may be disqualified from some events (a single event, multiple events, or all events) if the competitor:
     - 2k1) Fails to check in or register in time for the competition.
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -88,7 +88,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - 2i2b) The competitor must not interact with (e.g. operate, hold, wear) any active camera. Exception: the competitor may wear a camera mounted on their head, as long as it is out of their sight and it is clear that they are not interacting with it (apart from wearing it).
 - 2j) The WCA Delegate may disqualify a competitor from a specific event.
     - 2j1) If a competitor is disqualified from an event for any reason, they are not eligible for any more attempts in the event.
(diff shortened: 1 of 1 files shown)


xsrvmy:
I think this leaves one issue open for discussion: what should be recorded if the competitor is disqualified in the middle for their first attempt (eg. cheating)?

mirandapedro:
Perhaps the wording on the PR should changed from "finished" to "started" so that, if a competitor has started an attempt and is disqualified in the middle of it, the attempt that already has started and the following attempts get recorded as DNF.


This is in alignment with the definition of Regulation 9f4 of what constitutes a DNF:


`9f4) The result of an attempt is recorded as DNF (Did Not Finish) if the attempt is disqualified or unsolved/unfinished`


What do you think?

xsrvmy:
Yeah started makes more sense. (I'm only a regular competitor btw)

wca-regulations.md
```diff
@@ -88,7 +88,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - 2i2b) The competitor must not interact with (e.g. operate, hold, wear) any active camera. Exception: the competitor may wear a camera mounted on their head, as long as it is out of their sight and it is clear that they are not interacting with it (apart from wearing it).
 - 2j) The WCA Delegate may disqualify a competitor from a specific event.
     - 2j1) If a competitor is disqualified from an event for any reason, they are not eligible for any more attempts in the event.
-        - 2j1a) The results of all remaining attempts in the event are recorded as DNF.
+        - 2j1a) If the competitor has already finished one or more attempts in a round after being disqualified, the results of all remaining attempts in the event are recorded as DNF. If the competitor has finished no attempts, no results should be recorded.
```
```suggestion
        - 2j1a) If the competitor has already finished one or more attempts in a round after being disqualified, the results of all remaining attempts in the event are recorded as DNF. If did not start any attempt, no results are recorded.
```

I think this would solve @xsrvmy's concerns.

Ivan-Brigidano (changes requested):

wca-regulations.md
xsrvmy: The first part of the sentence should also started?

Ivan-Brigidano: I'm sorry but I don't understand what you mean. Do you mean about my proposal or the original proposal (Pedro's one)?

mirandapedro: I think what @xsrvmy means is that in `If the competitor has already finished one or more attempts in a round after being disqualified` (the first part of my initial proposal), "finished" should also be altered to "started". I'll push a commit that consolidates both of these changes.

mirandapedro:
Thank you Iván for your help and @xsrvmy for your contributions! It doesn't matter if you're a regular competitor as long as you're willing to help out and improve the Regulations. Comments like these are very appreciated!

wca-regulations.md
```diff
@@ -88,7 +88,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - 2i2b) The competitor must not interact with (e.g. operate, hold, wear) any active camera. Exception: the competitor may wear a camera mounted on their head, as long as it is out of their sight and it is clear that they are not interacting with it (apart from wearing it).
 - 2j) The WCA Delegate may disqualify a competitor from a specific event.
     - 2j1) If a competitor is disqualified from an event for any reason, they are not eligible for any more attempts in the event.
-        - 2j1a) The results of all remaining attempts in the event are recorded as DNF.
+        - 2j1a) If the competitor had already started at least one attempt in a round before being disqualified, the results of all remaining attempts in the event are recorded as DNF. If the competitor had started no attempts, no results should be recorded.
```
Ivan-Brigidano: One small question before merging this PR:

> If the competitor had started no attempts, no results _should_ be recorded.

What is the purpose of this "should"? Wouldn't it be enough to phrase is similar to the previous sentence?

> If the competitor had started no attempts, no results _are recorded_.

mirandapedro: You're right! I'll push another commit fixing this.

Ivan-Brigidano (approved):
LGTM!

wca-regulations.md
```diff
@@ -88,7 +88,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - 2i2b) The competitor must not interact with (e.g. operate, hold, wear) any active camera. Exception: the competitor may wear a camera mounted on their head, as long as it is out of their sight and it is clear that they are not interacting with it (apart from wearing it).
 - 2j) The WCA Delegate may disqualify a competitor from a specific event.
     - 2j1) If a competitor is disqualified from an event for any reason, they are not eligible for any more attempts in the event.
-        - 2j1a) The results of all remaining attempts in the event are recorded as DNF.
+        - 2j1a) If the competitor had already started at least one attempt in a round before being disqualified, the results of all remaining attempts in the event are recorded as DNF. If the competitor had started no attempts, no results are recorded.
```
lgarron: I have two small suggestions:

- We can avoid some passive voice.
- It's possible that the competitor has "started one attempt", but this attempt doesn't count, if they were awared an extra attempt. I *think* we don't allow round results that are all DNS?

```suggestion
        - 2j1a) If the competitor has already started at least one attempt in a round before being disqualified, the results of all remaining attempts in the event are recorded as DNF. If the competitor has started no attempts (or only has DNS results), no results are recorded.
```

mirandapedro: I agree with your first suggestion, but I'm not sure if the second is necessary. If the competitor started one attempt, but was granted an extra for it and was disqualified before starting the extra, his result should be recorded as a DNF and not a DNS. As per the definition of a DNS in Regulation 9f5, the competitor did not decline the attempt, so it should be recorded as a DNF instead of a DNS.

It is true that we don't record results that are all DNS, but inserting this on 2j1a does not make sense, as this Regulation deals specifically with the results of attempts in which a competitor was disqualified.

What do you think?

SAuroux: > If the competitor started one attempt, but was granted an extra for it and was disqualified before starting the extra, his result should be recorded as a DNF and not a DNS. 

I disagree with this. As soon as the competitor is granted the extra attempt, the started attempt is voided, so the competitor goes back to the state of having started no attempt.
The definition of a DNS in Regulation 9f5 is not relevant here, since 2j1a) is already superceeding this no matter what, i.e.  the granted extra attempt is no different from the remaining attempts of the round which would also not be voluntarily declined in case of a disqualification.

But for the exact same logic I also disagree with @lgarron 's second suggestion, as having only DNS results is already part of having started no attempts. In my view, the following is more explicit: 

> [...] If the competitor has started no attempts, i.e. has no results or only has DNS results, no results are recorded.

Completely unrelated, I also just realized that the first sentence as is currently leaves a loophole because _a round_ is not associated to the round the competitor is disqualified for. Therefore the first sentence should be:

> 2j1a) If the competitor has already started at least one attempt in a round before being disqualified **from it**, the results of all remaining attempts in the event are recorded as DNF. [...]

mirandapedro: > since 2j1a) is already superceeding this no matter what, i.e. the granted extra attempt is no different from the remaining attempts of the round which would also not be voluntarily declined in case of a disqualification.

By bringing up Regulation 2j1a, you agree with my point that the extra attempt should be recorded as a DNF instead of a DNS, as "The results of all remaining attempts in the event are recorded as DNF" includes the extra attempt. 

This is a different situation than what happened in the incident log where a competitor was disqualified before the round has even started. It is true that the attempt is voided whan an extra is granted, but going by the PRs original intent to try and represent in the results of a competition the closest possible to what actually happenned, if a competitor:
- Started their first attempt in a round
- Were granted and extra attempt, and 
- Were disqualified from the round before starting the extra attempt granted

They participated in the round in a general sense (even if the attempt they actually started was voided), and their results for the round should be recorded as DNF instead of not having results at all.

What do you think?

> Completely unrelated, I also just realized that the first sentence as is currently leaves a loophole because a round is not associated to the round the competitor is disqualified for. Therefore the first sentence should be:

`2j1a) If the competitor has already started at least one attempt in a round before being disqualified from it, the results of all remaining attempts in the event are recorded as DNF. [...]`

Nice improvement! I'll incorporate it in the next commit once we've agreed on the previous discussion.

SAuroux: > By bringing up Regulation 2j1a, you agree with my point that the extra attempt should be recorded as a DNF instead of a DNS, as <current 2ja1>

No. You realize that 2ja1 is the regulation discussed here? :)
I was only pointing out that the new version of 2ja1 is no matter what going to include scenarios where not voluntarily declined attempts are recorded as DNS, i.e. when a competitor is disqualified before the round has started.

> What do you think?

You have effectively not given any new argument, so I do of course still disagree. For some reasons that I do not comprehend, you keep insisting that the existence of some voided attempt in the temporal past should make a difference, which I am not subscribing to, as this is not in line with any other logic used in the regulations.

Instead, I insist on considering voided attempts as voided. I therefore disagree with your statement _They participated in the round in a general sense_ and I continue to say that it should not matter if a competitor with zero completed attempts in a round has previously received zero, one or a hundred extra attempts in that round. If you drew a state diagram of the competitor in that round, all three diagrams would look exactly the same.

Ivan-Brigidano: As soon as an extra attempt is granted, the original result is discarded. Therefore, if the competitor discard the extras, this should be considered DNS. This [Incident Log](https://www.worldcubeassociation.org/incidents/41) describe this exact situation.

> [...] If the competitor has started no attempts, i.e. has no results or only has DNS results, no results are recorded.

I like this suggestion. Only that I prefer using parenthesis rather than commas.

```suggestion
        - 2j1a) If the competitor has already started at least one attempt in a round before being disqualified from it, the results of all remaining attempts in the event are recorded as DNF. If the competitor has started no attempts (i.e. has no results or only has DNS results), no results are recorded.
```

PR #997, opened 2021-03-03
Deal with scoretaking mistakes/retroactive penalties, and how to affect other rounds

julesDesjardin:
When a scoretaking mistake is discovered, or a penalty is applied to a solve after video review, this can affect competitors' advancement in the rounds.
[This incident](https://www.worldcubeassociation.org/incidents/48) is an example of such an issue, and contains the decision we took:
- Keep the results if the competitor proceeded thanks to an error in someone else's results
- Remove the results if the competitor proceeded thanks to an error in their own results
This can cascade due to 9p1, and/or 9m1, 9m2, 9m3, which rely on the number of competitors that were in the round. This can impact innocent competitors. For fairness, we decided that rounds can overrule these Regulations. (see [This incident](https://www.worldcubeassociation.org/incidents/56)).

This PR adds these decisions to the Regulations, with an example for 9t2 which can be difficult to understand.


xsrvmy:
I think 9t1 needs to be reworded - the consequence of applying 9p1 retrospectively would be the removal of the results of a competitor, not the removal of a round.

Jambrose777:
> I think 9t1 needs to be reworded - the consequence of applying 9p1 retrospectively would be the removal of the results of a competitor, not the removal of a round.

Actually, both can happen as a result of someone’s results being removed.
Say for instance, round 2 has 8 competitors and round 3 has 6. And someone in round 1 was disqualified and had their results removed from round 2, this would in turn have to remove 6 people’s round 3 results and ultimately change the final round to round 2 instead (where multiple scramble groups may have been applied) without this regulation

julesDesjardin:
Thanks for your comments, I did not catch this problem!
I rewrote 9t1 in 2 separate rules (one for results of competitors because of 75%, the other for subsequent rounds). I also added a 9t+ to explain what these rules actually mean.

Jambrose777:
Looks good! The only /issue/ I foresee is some Delegate may read one of the 9t1, 9t2, or 9t3 without the context of 9t, meaning that they may not apply the time period of after the competition when applying those regulations, which could cause many problems.

For 9t3's time period: If 3 rounds of an event have occurred and at some point before the competition the scorecards are checked and find that this situation has occurred, then the same actions in 9t3 should be followed as to not break 9l and possibly 1h+. And if only 2 rounds occurred such that the Delegate could try to resolve such a problem, we would need to ensure that the competitor who already competed does not have their results removed. Overall, I'm not sure how to word this grey area to give the most clarity for Delegates.

wca-guidelines.md
```diff
@@ -131,6 +131,8 @@ To be more informative, each Guideline is classified using one of the following
...
 - 9p2+) [CLARIFICATION] Result refers to the single or average that determines the ranking of the corresponding round.
 - 9q+) [RECOMMENDATION] Events and rounds should have at least 2 competitors.
+- 9t+) [EXPLANATION] Because of [9t1](regulations:regulain:9t1) and [9t2](regulations:regulain:9t2), rounds may look like they don't comply with [9p1](regulations:regulation:9m1), [9m1](regulations:regulation:9m1), [9m2](regulations:regulation:9m2) or [9m3](regulations:regulation:9m3). However, makings changes to comply to these rules would affect innocent competitors, which is why the results must stay the same.
+- 9t3+) [EXAMPLE] Suppose competitor A get a 10.00 average and competitor B get a 10.05 average in the first round of an event, such that competitor A advances to the final round but not B. If a score-taking mistake is discovered, and competitor B actually got a 9.95 average (and should have proceeded instead of A), competitor A keeps their result for the final round.
```
SAuroux: Shouldn't it be either _gets_ or _got_ in the first sentence in both occurrences of _get a ... average_?

julesDesjardin: Indeed, nice catch!

julesDesjardin:
@Jambrose777 The time period is indeed maybe not the best one, and this kind of thing can happen during the competition. Defining an exact time period seems pretty hard because of potential cascades.
We can add a guideline to say that this can also happen during a competition, to the discretion of the Delegate, to give them the opportunity to keep the spirit of this Regulation, and not have to remove rounds or results.

EdHollingdale:
@julesDesjardin if the spirit of the regulation is to allow results that were thought to be official at the time they were set to stand, then we should add a guideline - although I don't have any great wording suggestions in mind. 

I can easily see situations where delegates checking scorecards on a Sunday afternoon could get a "fairer" solution by pretending they didn't find the error until a few hours later. In situations like that I'd like to not have to encourage duplicity.

julesDesjardin:
Here is a suggestion of such a guideline. Using the WCA Delegate's discretion should be enough to cover the cases where these rules make sense.

wca-guidelines.md
```diff
@@ -131,6 +131,9 @@ To be more informative, each Guideline is classified using one of the following
 - 9f5+) [CLARIFICATION] The result of an attempt is DNS if the competitor was eligible for the attempt and did not start it (see [Regulation A3b2](regulations:regulation:A3b2)). If the competitor did not qualify for an attempt (e.g. in a cutoff round), they have no result for the attempt.
 - 9p2+) [CLARIFICATION] Result refers to the single or average that determines the ranking of the corresponding round.
 - 9q+) [RECOMMENDATION] Events and rounds should have at least 2 competitors.
+- 9t+) [EXPLANATION] Because of [9t1](regulations:regulain:9t1) and [9t2](regulations:regulain:9t2), rounds may look like they don't comply with [9p1](regulations:regulation:9m1), [9m1](regulations:regulation:9m1), [9m2](regulations:regulation:9m2) or [9m3](regulations:regulation:9m3). However, makings changes to comply to these rules would affect innocent competitors, which is why the results must stay the same.
```
cubewhiz:
```suggestion
- 9t+) [EXPLANATION] Because of [9t1](regulations:regulain:9t1) and [9t2](regulations:regulain:9t2), rounds may look like they don't comply with [9p1](regulations:regulation:9m1), [9m1](regulations:regulation:9m1), [9m2](regulations:regulation:9m2) or [9m3](regulations:regulation:9m3). However, making changes to comply to these rules would affect innocent competitors, which is why the results must stay the same.
```

Ivan-Brigidano (approved):

PR #998, opened 2021-03-04
Clarify who can approve not allowing competitiors to participate in all events of a competition

mirandapedro:
This PR clarifies who can approve competitors not participating in all events of a competition.

As highlighted by @Epride [here](https://forum.worldcubeassociation.org/t/inaccuracy-of-regulation-9k/8926),  Regulation 9k currently inaccurately states that the WCA Board is the one responsible for approving that competitors aren't able to participate in all events of a competition.

However, in almost all cases, this approval is not carried out by the Board, but by the WCAT.

This PR, therefore, makes it clear that the Board specifically doesn't need to give the mentioned approval.

We chose not to change the wording in the Regulation mentioning specifically the WCAT so that the Regulations doesn't need to be updated if the Team changes its name or if anoother entity in the WCA becomes responsible for the approval.

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -250,7 +250,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 9i2) All the results of a round are considered to take place on the last calendar date of the round. If a regional record is broken multiple times on the same calendar date, only the best result is recognized as breaking that regional record.
     - 9i3) If the WCA Regulations for an event are changed, existing regional records stand until they are broken under the new WCA Regulations.
 - 9j) Each event must be held at most once per competition.
-- 9k) All competitors may participate in all events of a competition, except in cases specifically approved by the Board.
+- 9k) All competitors may participate in all events of a competition, except in cases specifically approved by the WCA.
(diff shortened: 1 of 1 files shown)


Ivan-Brigidano (approved):
LGTM!

Jambrose777:
I don't agree with specifying "WCA". A Delegate represents the WCA and via the regulation change can make such restrictions, which isn't the intention of the regulation.

mirandapedro:
This wording was taken from what was proposed by Ethan in the Forum post. Do you perhaps have a suggestion for a better term?

Jambrose777:
> This wording was taken from what was proposed by Ethan in the Forum post. Do you perhaps have a suggestion for a better term?

I think explicitly stating WCAT is fine. IMO The pros of being specific outway the cons. If the entity ever changes, then another regulation change would be required, but there won't be any misunderstanding as to who "WCA" is referring to.

mirandapedro:
I agree with that. We also already mention a WCA team in the Regulations (Regulation 2c2), so it is in alignment with what is in practice.

I'll try and write the PR later today. Thanks Jacob for the suggestion!

PR #1000, opened 2021-03-05
Clarification regarding competitors wearing a smartphone on their pockets

Ivan-Brigidano:
Forum discussion: [https://forum.worldcubeassociation.org/t/when-does-2i-apply-exactly-and-other-comment/3203](https://forum.worldcubeassociation.org/t/when-does-2i-apply-exactly-and-other-comment/3203).

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -86,6 +86,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 2i2) Competitors may use cameras at the solving station at the discretion of the WCA Delegate, but the following restrictions apply from the start of the attempt until the competitor stops the solve. Penalty for breaking a restriction: disqualification of the attempt (DNF).
         - 2i2a) Each camera monitor must be blank or out of sight of the competitor (see [Regulation A5b](regulations:regulation:A5b)).
         - 2i2b) The competitor must not interact with (e.g. operate, hold, wear) any active camera. Exception: the competitor may wear a camera mounted on their head, as long as it is out of their sight and it is clear that they are not interacting with it (apart from wearing it).
+    - 2i3) Exception: the competitor may wear a cell phone on their pocket, as long as it is out of their sight and it is clear that they are not interacting with it (apart from wearing it).
 - 2j) The WCA Delegate may disqualify a competitor from a specific event.
     - 2j1) If a competitor is disqualified from an event for any reason, they are not eligible for any more attempts in the event.
         - 2j1a) The results of all remaining attempts in the event are recorded as DNF.
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -86,7 +86,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 2i2) Competitors may use cameras at the solving station at the discretion of the WCA Delegate, but the following restrictions apply from the start of the attempt until the competitor stops the solve. Penalty for breaking a restriction: disqualification of the attempt (DNF).
         - 2i2a) Each camera monitor must be blank or out of sight of the competitor (see [Regulation A5b](regulations:regulation:A5b)).
         - 2i2b) The competitor must not interact with (e.g. operate, hold, wear) any active camera. Exception: the competitor may wear a camera mounted on their head, as long as it is out of their sight and it is clear that they are not interacting with it (apart from wearing it).
-    - 2i3) Exception: the competitor may have a cell phone in their pocket, as long as it is out of their sight and it is clear that they are not interacting with it (apart from it being on their person).
+    - 2i3) The competitor may have a cell phone in their pocket, as long as it is clear that they are not interacting with it (apart from it being on their person).
 - 2j) The WCA Delegate may disqualify a competitor from a specific event.
     - 2j1) If a competitor is disqualified from an event for any reason, they are not eligible for any more attempts in the event.
         - 2j1a) The results of all remaining attempts in the event are recorded as DNF.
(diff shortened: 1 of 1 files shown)


wca-regulations.md
```diff
@@ -86,6 +86,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 2i2) Competitors may use cameras at the solving station at the discretion of the WCA Delegate, but the following restrictions apply from the start of the attempt until the competitor stops the solve. Penalty for breaking a restriction: disqualification of the attempt (DNF).
         - 2i2a) Each camera monitor must be blank or out of sight of the competitor (see [Regulation A5b](regulations:regulation:A5b)).
         - 2i2b) The competitor must not interact with (e.g. operate, hold, wear) any active camera. Exception: the competitor may wear a camera mounted on their head, as long as it is out of their sight and it is clear that they are not interacting with it (apart from wearing it).
+    - 2i3) Exception: the competitor may wear a cell phone on their pocket, as long as it is out of their sight and it is clear that they are not interacting with it (apart from wearing it).
```
Jambrose777:
```suggestion
    - 2i3) Exception: the competitor may have a cell phone in their pocket, as long as it is out of their sight and it is clear that they are not interacting with it (apart from it being on their person).
```

The original doesn't make sense as you typically don't wear a cell phone.

Ivan-Brigidano: Thank you!

mirandapedro (approved):
LGTM!

SAuroux:
I am quite significantly confused by this current formulation. 

My first thought was: What would be a scenario where a competitor has a smartphone in their pocket but at the same time it is not out of sight? Then I realized that larger smartphones might partially stick out of pockets and thus be partially visible and therefore in sight. But certainly that is nothing we see as a problem, right?

So: What would be a _problematic_ scenario where a competitor has a smartphone in their pocket but at the same time it is not out of sight?

I really can't see any that would and I therefore think that bringing "being in sight" into this regulation creates far more problems than it solves. My simple proposal:

> 2i3) Exception: the competitor may have a cell phone in their pocket, as long as it is clear that they are not interacting with it (apart from it being on their person).

Goosly:
2i3 starts with "Exception", but I really don't understand how 2i3 is an exception to 2i. Rather, 2i3 is a clarification on 2i. (An unnecessary one, in my opinion)

Ivan-Brigidano:
> 2i3 starts with "Exception", but I really don't understand how 2i3 is an exception to 2i. Rather, 2i3 is a clarification on 2i. (An unnecessary one, in my opinion)

I agree that this is not strictly a "Exception". However, I do think that it's worth clarifying.

wca-regulations.md
```diff
@@ -86,6 +86,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 2i2) Competitors may use cameras at the solving station at the discretion of the WCA Delegate, but the following restrictions apply from the start of the attempt until the competitor stops the solve. Penalty for breaking a restriction: disqualification of the attempt (DNF).
         - 2i2a) Each camera monitor must be blank or out of sight of the competitor (see [Regulation A5b](regulations:regulation:A5b)).
         - 2i2b) The competitor must not interact with (e.g. operate, hold, wear) any active camera. Exception: the competitor may wear a camera mounted on their head, as long as it is out of their sight and it is clear that they are not interacting with it (apart from wearing it).
+    - 2i3) Exception: the competitor may have a cell phone in their pocket, as long as it is out of their sight and it is clear that they are not interacting with it (apart from it being on their person).
```
```suggestion
    - 2i3) The competitor may have a cell phone in their pocket, as long as it is clear that they are not interacting with it (apart from it being on their person).
```

wca-regulations.md
```diff
@@ -86,7 +86,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 2i2) Competitors may use cameras at the solving station at the discretion of the WCA Delegate, but the following restrictions apply from the start of the attempt until the competitor stops the solve. Penalty for breaking a restriction: disqualification of the attempt (DNF).
         - 2i2a) Each camera monitor must be blank or out of sight of the competitor (see [Regulation A5b](regulations:regulation:A5b)).
         - 2i2b) The competitor must not interact with (e.g. operate, hold, wear) any active camera. Exception: the competitor may wear a camera mounted on their head, as long as it is out of their sight and it is clear that they are not interacting with it (apart from wearing it).
-    - 2i3) Exception: the competitor may have a cell phone in their pocket, as long as it is out of their sight and it is clear that they are not interacting with it (apart from it being on their person).
+    - 2i3) The competitor may have a cell phone in their pocket, as long as it is clear that they are not interacting with it (apart from it being on their person).
```
lgarron: I'd suggest keeping this simple:
```suggestion
    - 2i3) The competitor may have a cell phone in a pocket, as long as it is clear that they are not interacting with it (apart from having it in a pocket).
```

Could also do "as long as it is clear that they are not otherwise interacting with it."

EdHollingdale:
I agree that this is more of a clarification rather than an exception. However isn't it better then to make it slightly broader - what if a competitor has another electronic item in their pocket? If we think it is worth clarifying for a phone shouldn't we clarify more generally?

Ivan-Brigidano:
> what if a competitor has another electronic item in their pocket?

Maybe changing it to "a electronic device (e.g. a cell phone)" could work.

wca-regulations.md
cubewhiz: I prefer the alternative you suggested:
"as long as it is clear that they are not otherwise interacting with it."

wca-regulations.md
```diff
@@ -86,6 +86,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 2i2) Competitors may use cameras at the solving station at the discretion of the WCA Delegate, but the following restrictions apply from the start of the attempt until the competitor stops the solve. Penalty for breaking a restriction: disqualification of the attempt (DNF).
         - 2i2a) Each camera monitor must be blank or out of sight of the competitor (see [Regulation A5b](regulations:regulation:A5b)).
         - 2i2b) The competitor must not interact with (e.g. operate, hold, wear) any active camera. Exception: the competitor may wear a camera mounted on their head, as long as it is out of their sight and it is clear that they are not interacting with it (apart from wearing it).
+    - 2i3) The competitor may have a cell phone in their pocket, as long as it is clear that they are not interacting with it (apart from it being on their person).
```
```suggestion
    - 2i3) The competitor may have a cell phone in their pocket, as long as it is clear that they are not otherwise interacting with it.
```

PR #1002, opened 2021-03-06
Make puzzle related announcement not strictly mandatory

hanwu85:
Making SQ-1 inserts and clock stands announcement not mandatory.

[Discussion on the WCA forum](https://forum.worldcubeassociation.org/t/modify-a2b1-and-f2a-to-not-require-an-announcement/1291)


xsrvmy:
Is there a particular reason that it is being changed to a "may" instead of a "should"?

hanwu85:
Hi,

This is  open for discussion, currently this is only a draft.
What do you think of MAY vs. SHOULD in this situation?

Ivan-Brigidano:
I think it's better to use "should" so that the announcement still needs to be done but it's not strictly mandatory.

hanwu85:
I have changed the MAYs to SHOULDs. Thanks!

Ivan-Brigidano (approved):

PR #1017, opened 2021-05-02
Fix typos

Gelma:
Hi,
   and thanks a lot for your work and the idea to keep your rules available on GitHub. That's really smart and cool.

   This is just a stupid contribution on two typos (one maybe not, I dunno...).

Thanks again,
Gelma

### DIFF
--- a/style-guide.md
+++ b/style-guide.md
@@ -9,7 +9,7 @@ This document is meant to clarify some of the implicit conventions adopted in WC
   - The Guidelines have an irregular correspondence to numbering in the Regulations, thus Guidelines have no indentation in the source.
 - The division into Regulations Guidelines roughly mirrors the normative vs. non-normative (informative) distinction employed by W3C standards.
   - The use of the words "must (not)", "shall (not)", "should (not)", and "may (not)" (requirement levels) conform to RFC 2119. In particular, note that "shall" indicates a (non-optional) requirement.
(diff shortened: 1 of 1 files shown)


style-guide.md
```diff
@@ -122,4 +122,4 @@ This document is meant to clarify some of the implicit conventions adopted in WC
 - Ranking/Rank: An ordinal number representing a result's order in a list of results (e.g. ranking in a round, world rank).
 - Record: A best-ranked result (e.g. of comparable results for a region, or of results for a competitor)
 - Part: A subset of puzzle components (in particular, any individual piece, cap, sticker, etc.).
-- Piece: A part of the puzzle that stays together during normal use of a puzzle (e.g. a cubie with its stickers).
+- Piece: A part of the puzzle that stays together during normal use of a puzzle (e.g. a cube with its stickers).
```
lgarron: This is specifically using the cuber name of "cubie" for a cubic piece to give an example. It should stay.
```suggestion
- Piece: A part of the puzzle that stays together during normal use of a puzzle (e.g. a cubie with its stickers).
```

lgarron (commented):
Thanks!

I don't think we'll be able to apply it to the current release, but it's never too early to start on the next one!

EDIT: actually, looks like this might just be reduced to a style guide fix. But I'm glad we're getting more eyes on everything. :-D

lgarron (approved):

PR #1018, opened 2021-05-03
Shorten the initial README paragraph.

lgarron:
This contained some redundant information, and I think we ended up up that way mostly by accident. Here's a way to shorten it.


Ivan-Brigidano (approved):
LGTM!

PR #1019, opened 2021-05-06
Fixing typo in link

hanwu85:
Fixing some link error due to typo.


lgarron (approved):
Thanks for catching this!

I'd say this warrants a typo fix update, what do you think?

Ivan-Brigidano:
We have Board approval for merging these typo fixes on the `official` branch. @lgarron @hanwu85 can you please handle this?

lgarron:
Merged and deployed!

As a side note, I meant to note that we haven't shortened the text of cross-links in the past — this changed fixed the links but not did't update the text. I think it would be clearest (and simpler to decide in future places) to always use "Regulation ___", but if there's any place that warrants shortening, here seems alright.

PR #1026, opened 2021-10-24
Clarification on how disqualifications affect results

Ivan-Brigidano:
See the [WCA Forum discussion](https://forum.worldcubeassociation.org/t/2j2-clarity/13699).


mirandapedro (approved):
LGTM! I especially like that we were able to implement this in an already existing regulation.

lgarron (approved):

PR #1028, opened 2021-10-24
Clarification on Clock hardware, which is approved at WCA Delegate discretion

Ivan-Brigidano:
See the [WCA Forum discussion](https://forum.worldcubeassociation.org/t/proposal-anything-goes-for-clock/3118).

### DIFF
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -72,6 +72,7 @@ To be more informative, each Guideline is classified using one of the following 
 - 3h2++) [CLARIFICATION] "Stickerless" puzzles that significantly differ from most mass-produced "stickerless" puzzles are only permitted at the discretion of the WCA Delegate.
 - 3j+) [REMINDER] In the past, engraved/embossed parts have been permitted. This is no longer permitted.
 - 3j++) [CLARIFICATION] On Clock, pins may be distinguishable from any other pin of the same side.
+- 3j1+) [REMINDER] On Clock, loose pins are not considered a reasonable wear, at the discretion of the WCA Delegate.
 - 3j2+) [CLARIFICATION] On Clock, the nine inner clock faces on one side are considered similar pieces.
 - 3k2+) [ADDITION] The WCA Delegate should not apply a listed exception if they believe the competitor tried to use a non-permitted puzzle on purpose.
 - 3k2b+) [CLARIFICATION] If a non-permitted puzzle is found during the course of a 3x3x3 Multi-Blind attempt, the puzzle must not be exchanged or removed from the attempt and must be counted as unsolved if the entire attempt is not disqualified.
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -72,7 +72,7 @@ To be more informative, each Guideline is classified using one of the following 
 - 3h2++) [CLARIFICATION] "Stickerless" puzzles that significantly differ from most mass-produced "stickerless" puzzles are only permitted at the discretion of the WCA Delegate.
 - 3j+) [REMINDER] In the past, engraved/embossed parts have been permitted. This is no longer permitted.
 - 3j++) [CLARIFICATION] On Clock, pins may be distinguishable from any other pin of the same side.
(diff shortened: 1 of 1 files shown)


wca-guidelines.md
```diff
@@ -72,6 +72,7 @@ To be more informative, each Guideline is classified using one of the following
 - 3h2++) [CLARIFICATION] "Stickerless" puzzles that significantly differ from most mass-produced "stickerless" puzzles are only permitted at the discretion of the WCA Delegate.
 - 3j+) [REMINDER] In the past, engraved/embossed parts have been permitted. This is no longer permitted.
 - 3j++) [CLARIFICATION] On Clock, pins may be distinguishable from any other pin of the same side.
+- 3j1+) [REMINDER] On Clock, loose pins are not considered a reasonable wear, at the discretion of the WCA Delegate.
```
```suggestion
- 3j1+) [REMINDER] On Clock, loose pins are not considered reasonable wear, at the discretion of the WCA Delegate.
```

Jambrose777:
```suggestion
- 3j1+) [REMINDER] On Clock, loose pins are not considered reasonable wear, at the discretion of the WCA Delegate.
```

the aritcle is not needed here.

mirandapedro (approved):
LGTM!

Rouxles (approved):

wca-guidelines.md
Jambrose777: Then you shouldn't like the comma in the other 24 places we have one for this phrase in the regulations. I agree that it's not grammatically needed, but it is needed for consistency with the other regulations.

Rouxles: Yep, you're right. Initially I checked the regulations to see whether there were any instances of *at the discretion of the WCA delegate* without a comma which did end up existing, but I didn't look much further to see that the ones without commas needed that clause for the sentence to make sense - I'll go update the suggestion.

wca-guidelines.md
```diff
@@ -72,6 +72,7 @@ To be more informative, each Guideline is classified using one of the following
 - 3h2++) [CLARIFICATION] "Stickerless" puzzles that significantly differ from most mass-produced "stickerless" puzzles are only permitted at the discretion of the WCA Delegate.
 - 3j+) [REMINDER] In the past, engraved/embossed parts have been permitted. This is no longer permitted.
 - 3j++) [CLARIFICATION] On Clock, pins may be distinguishable from any other pin of the same side.
+- 3j1+) [REMINDER] On Clock, loose pins are not considered reasonable wear, at the discretion of the WCA Delegate.
```
lgarron: I can't help but feel like this seems to say "loose pins are allowed if the Delegate is fine with it" rather than "loose pins are not allowed, and the Delegate decides if it's not clear". Given that we don't use "at the discretion of the WCA Delegate" for e.g. 3j+++, could we leave that out and lean on 3k instead?
```suggestion
- 3j1+) [REMINDER] On Clock, loose pins (i.e. pins that can toggle or recede using gravity instead of being directly pressed)  are not considered reasonable wear.
```

lgarron (approved):

wca-guidelines.md
```diff
@@ -72,7 +72,7 @@ To be more informative, each Guideline is classified using one of the following
 - 3h2++) [CLARIFICATION] "Stickerless" puzzles that significantly differ from most mass-produced "stickerless" puzzles are only permitted at the discretion of the WCA Delegate.
 - 3j+) [REMINDER] In the past, engraved/embossed parts have been permitted. This is no longer permitted.
 - 3j++) [CLARIFICATION] On Clock, pins may be distinguishable from any other pin of the same side.
-- 3j1+) [REMINDER] On Clock, loose pins are not considered reasonable wear, at the discretion of the WCA Delegate.
+- 3j1+) [REMINDER] On Clock, loose pins (i.e. pins that can toggle or recede using gravity instead of being directly pressed)  are not considered reasonable wear.
```
lgarron, line 75: Sorry, it looks like I added an extra space here.

Ivan-Brigidano: We need a new PR if we want to fix it, right?

lgarron: Yeah.

PR #1046, opened 2022-10-02
Clarify A7g

CarterKoala:
Essentially takes out the word "new" to clarify the competitor does not have to be a new competitor and gives specific examples for when this applies.


Nanush7 (approved):

wca-guidelines.md
```diff
@@ -169,9 +169,10 @@ To be more informative, each Guideline is classified using one of the following
...
-- A7g++) [CLARIFICATION] Individual penalties cannot be waived for new competitors. A full extra attempt must be granted.
-- A7g+++) [RECOMMENDATION] The WCA Delegate should use stronger discretion when granting more than one extra attempt to the same new competitor. If more than one extra attempt is granted, it should be for a different incident.
+- A7g+) [EXAMPLE] Examples of incidents that could grant a competitor an extra attempt due to inexperience: not waiting for the green light, improperly starting or stopping the timer, going over inspection time due to misunderstanding how to start a solve, or other procedural time penalties.
+- A7g++) [CLARIFICATION] Individual penalties cannot be waived. A full extra attempt must be granted.
+- A7g+++) [RECOMMENDATION] The WCA Delegate should use stronger discretion when granting more than one extra attempt to the same competitor. If more than one extra attempt is granted, it should be for a different incident.
```
lgarron: This detail always bothered me — good chance to fix it.
```suggestion
- A7g+++) [RECOMMENDATION] The WCA Delegate should use stronger discretion when granting more than one extra attempt to the same competitor. If more than one extra attempt is granted, it should be for a different type of incident. Example: if the WCA Delegate already awarded an extra attempt for a competitor who did not wait long enough for the green light, they should not award another extra attempt if that competitor fails to wait long enough for the green light again.
```

wca-guidelines.md
```diff
@@ -169,9 +169,10 @@ To be more informative, each Guideline is classified using one of the following
...
+- A7g++++) [CLARIFICATION] A competitor may be considered inexperienced if they are a new competitor, it is their first time competing in a specific event, or they have not competed in a specific event for a significant period of time.
```
lgarron: I think we need to be very careful about this, especially if https://github.com/thewca/wca-regulations/pull/1059 is also adopted.

Could we 1) clarify that these are the only valid cases (the first two could be collapsed), and 2) the inexperience must be related to the incident?
I think I'd also be more comfortable if we put a minimum on "significant period" — a year at the very least, or take into account Regulation changes.

I'd also consider moving these to the Regs at this point.

wca-regulations.md
```diff
@@ -437,7 +437,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - A7c4) The competitor should not sign an attempt before the judge has recorded and signed the attempt. The competitor is ultimately responsible for ensuring that they and the judge have signed an attempt. If an attempt is found to be missing one or both signatures after the judge delivered the score sheet to the score taker (see [Regulation A7f](regulations:regulation:A7f)), it will be considered unfinished (DNF).
         - A7c5) Missing signatures found before the end of the competition may be added by the judge and/or the competitor, at the discretion of the WCA Delegate.
     - A7f) When a competitor's score sheet for a round is complete, the judge delivers the score sheet to the score taker.
-    - A7g) At the discretion of the WCA Delegate, an incident or penalty caused by a new competitor's inexperience may be replaced with an extra attempt.
+    - A7g) At the discretion of the WCA Delegate, an incident or penalty caused by a competitor's inexperience may be replaced with an extra attempt.
```
```suggestion
    - A7g) At the discretion of the WCA Delegate, an incident or penalty caused by a competitor's inexperience may be replaced with an extra attempt.
```

wca-guidelines.md
```suggestion
- A7g++++) [CLARIFICATION] A competitor may be considered inexperienced if they are a new competitor, it is their first time competing in a specific event, or they have not competed in a specific event for a significant period of time. The inexperience must be related to the incident.
```

PR #1047, opened 2022-10-02
Square-1 Misalignment

CarterKoala:
Clarifies the limits of Square one misalignments in inspection.


Nanush7 (changes requested):
"Exception" makes me think it should be an [ADDITION]. WDYT?

CarterKoala:
I think that makes sense. Though I was looking at exceptions in the guidelines and they seem to be only accompanied by other guidelines - I think it makes more sense to have this in the regulations instead.

Goosly:
Why does this exception only apply to square-1?
If I make an accidental move less than 45° in 3x3, also within the limit of 10f, then the penalty still applies?

Nevseros:
I think Manu's comment makes sense. We shouldn't probably make it as exception, but as a clarification for all puzzles.

Randomno:
> Why does this exception only apply to square-1? If I make an accidental move less than 45° in 3x3, also within the limit of 10f, then the penalty still applies?

No, because you have not applied any moves, so none of the regulation applies.

A3c1 is very unclear as it stands, see [this forum thread](https://forum.worldcubeassociation.org/t/definition-of-move-in-a3c1/16632).

If this reg is changed to allow any turns within the limits of 10f, there are at least a couple of situations to consider:

- A competitor intentionally turning the puzzle less than the limit to gain an advantage
- A competitor receives the puzzle misaligned, and makes a turn greater than the limit, but has not performed a move (e.g. 3x3 comes with a third of a U' done, and they turn it 60 degrees clockwise)

Anyway that's pretty out of scope of this PR, but I do not believe this addition for Square-1 is the real issue here.

Nevseros:
> No, because you have not applied any moves, so none of the regulation applies.

Moves are basically defined in the article 12, so as I see for now, even if you made a 60° turn, it _shouldn't_ be penalized, but it's interpreted differently in fact - `moves` in A3c1 are interpreted as limits from 10f. 

As you pointed yourself, this regulation is unclear and it was discussed before, so I'm all for clarifying A3c1 at least with linking it to 10f.

wca-regulations.md
```diff
@@ -387,7 +387,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - A3b1) When the judge believes the competitor is ready, the judge asks "READY?". The competitor must be ready to start the attempt within one minute from the moment they are asked, or else the competitor forfeits the attempt (DNS), at the discretion of the judge.
         - A3b2) The competitor starts the attempt by confirming that they are ready, and the judge uncovers the puzzle and starts timing the inspection.
     - A3c) The competitor may pick up the puzzle during inspection.
-        - A3c1) The competitor must not apply moves during inspection. Penalty: disqualification of the attempt (DNF).
+        - A3c1) The competitor must not apply moves during inspection. Penalty: disqualification of the attempt (DNF). Exception: For Square-1, the competitor should not be penalized for any accidental move made within the limits of Regulation 10f](regulations:regulation:10f).
```
Nevseros: There is a `[` missing before `Regulation 10f]` so markdown doesn't work ;).

```suggestion
        - A3c1) The competitor must not apply moves during inspection. Penalty: disqualification of the attempt (DNF). Exception: For Square-1, the competitor should not be penalized for any accidental move made within the limits of [Regulation 10f](regulations:regulation:10f).
```

Nanush7 (approved):

wca-regulations.md
```diff
@@ -387,7 +387,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - A3b1) When the judge believes the competitor is ready, the judge asks "READY?". The competitor must be ready to start the attempt within one minute from the moment they are asked, or else the competitor forfeits the attempt (DNS), at the discretion of the judge.
         - A3b2) The competitor starts the attempt by confirming that they are ready, and the judge uncovers the puzzle and starts timing the inspection.
     - A3c) The competitor may pick up the puzzle during inspection.
-        - A3c1) The competitor must not apply moves during inspection. Penalty: disqualification of the attempt (DNF).
+        - A3c1) The competitor must not apply moves during inspection. Penalty: disqualification of the attempt (DNF). Exception: For Square-1, the competitor should not be penalized for any accidental move made within the limits of [Regulation 10f](regulations:regulation:10f).
```
lgarron, line 390: Is there any reason for "should" instead of "must"?

Nevseros: I think it's for the reason that accidential itself is not clearly defined thing, so there's no need in must.

lgarron: Hmm, normally we'd solve this by adding "discretion of the judge" in some way, but I agree that there's no need here.

PR #1048, opened 2022-10-02
Add Exception to 9l

CarterKoala:
I don't really like the wording of this and the addition of 9l1 and 9l2 make this more complicated then I originally thought. Would appreciate suggestions here

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -257,7 +257,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 9i3) If the WCA Regulations for an event are changed, existing regional records stand until they are broken under the new WCA Regulations.
 - 9j) Each event must be held at most once per competition.
 - 9k) All competitors may participate in all events of a competition, except in cases specifically approved by the WCA Competition Announcement Team.
-- 9l) Each round must be completed before any following round of the same event can start. Exception: extra attempts may be given at any point during a competition to replace an attempt that was initially invalid. If this causes a change to the advancement conditions to a round that has already started, the following regulations apply:
+- 9l) Each round must be completed before any following round of the same event can start. Exception: Extra attempts may be given at any point during a competition to replace an attempt that was initially invalid.
     - 9l1) If the competitor's new result causes another competitor to no longer qualify for a subsequent round, that competitor's results for the subsequent round must be kept.
     - 9l2) If the competitor's new result causes themself to no longer qualify for a subsequent round, their results for the subsequent round must be removed.
 - 9m) Events must have at most four rounds.
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -257,7 +257,9 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 9i3) If the WCA Regulations for an event are changed, existing regional records stand until they are broken under the new WCA Regulations.
 - 9j) Each event must be held at most once per competition.
 - 9k) All competitors may participate in all events of a competition, except in cases specifically approved by the WCA Competition Announcement Team.
-- 9l) Each round must be completed before any following round of the same event can start.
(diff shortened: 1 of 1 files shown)


Randomno:
Can't 9t just be expanded?

Nevseros (commented):
Liked the idea of adding 9l1 and 9l2 to make this exception consequences clear!

Nit: to make it consistent with somewhat similar 9t and its subsequents, maybe it's worth wording it similarly? I'd proposed:

```
9l) Each round must be completed before any following round of the same event can start. Exception: extra attempts may be given at any point during a competition to replace an attempt that was initially invalid, at the discretion of the WCA Delegate. 
9l1) If any change occurs to the results of a competitor, such that the competitor should not have advanced to the next round, all results of this competitor for all subsequent rounds must be removed.
9l2) If any change occurs to the result of a competitor, such that another competitor should not have advanced to the next round, the result of the second competitor must not be removed.
```

Randomno:
Would it not cover all situations to do this?:

`9l) Each round must be completed before any following round of the same event can start. Exception: An extra attempt may be given at any point during a competition to replace an attempt that was initially invalid, at the discretion of the WCA Delegate.`

`9t) If any change occurs to the results of a competitor after the next round has started (e.g. score-taking mistake, retroactive penalty), such that the competitor should not have advanced to the next round (or was not allowed to take part in the round), all results of this competitor for all subsequent rounds must be removed.`

I'm not clear on the exact circumstances leading to this change, but to me it seems like there will otherwise be unnecessary repetition with 9l and 9t subsections.

wca-regulations.md
```diff
@@ -257,7 +257,9 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 9i3) If the WCA Regulations for an event are changed, existing regional records stand until they are broken under the new WCA Regulations.
 - 9j) Each event must be held at most once per competition.
 - 9k) All competitors may participate in all events of a competition, except in cases specifically approved by the WCA Competition Announcement Team.
-- 9l) Each round must be completed before any following round of the same event can start.
+- 9l) Each round must be completed before any following round of the same event can start. Exception: extra attempts may be given at any point during a competition to replace an attempt that was initially invalid. If this causes a change to the advancement conditions to a round that has already started, the following regulations apply:
```
```suggestion
- 9l) Each round must be completed before any following round of the same event can start. Exception: Extra attempts may be given at any point during a competition to replace an attempt that was initially invalid.
```

Jambrose777:
Hey WRC,

I have some concerns with this regulation:

Please take a look at the following scenario:
- Round 1 takes place, 75% of competitors proceed to the next round.
- Round 2 takes place
- A Competitor A who was DNFed in Round 1 shows the Delegate video of an incident resulting in an extra (ie. judge reset the timer). 
- Competitor A receives an extra and is now fast enough for top-75% in round 2.
- Now Competitor A should compete in Round 2 and there is a competitor who has already competed whose results are now invalidated.

Another concern of mine is that this may have a slight issue with Record assignments. Records are recorded on the last calendar date of the round. So please take a look at the following scenario
- Round 1 takes place on Jan 1. Competitor A Breaks the WR of 10.00 with a 9.50.
- Round 2 takes place on Jan 1. Competitor A achieves a 9.75 average (better than the previous WR, but slower than the WR set in Round 1).
- On Day 2 of the competition (Jan 2), Competitor B shows the Delegates an incident causing an extra (ie a misscramble) and is awarded an extra. 
In the above scenario, Round 1's date is technically now Jan 2 and Round 2 is still on Jan 1, meaning that both rounds would be considered WR, which is kinda crazy.

As well, this kind of goes against [11e](https://www.worldcubeassociation.org/regulations/#11e) as we'd be granting extras for report incidents well after their completion instead of at the time of the incident.

Jambrose777:
> As well, this kind of goes against [11e](https://www.worldcubeassociation.org/regulations/#11e) as we'd be granting extras for report incidents well after their completion instead of at the time of the incident.

I see https://github.com/thewca/wca-regulations/pull/1060/files now, which adjusts 11e regulation to a should, which would negate this concern with 11e contradiction. :)

Rouxles (approved):

PR #1049, opened 2022-10-06
Re-add A1a2+++ as A1a2+++++

Nevseros:
Guideline A1a2+++ was removed [here](https://github.com/thewca/wca-regulations/commit/ea2d719fe5d7eb11ce210a9499aa1c5f4627560d). See the original discussion [here](https://forum.worldcubeassociation.org/t/readdition-of-guideline-a1a2/18218).


Rouxles (approved):
LGTM

wca-guidelines.md
```diff
@@ -154,6 +154,7 @@ To be more informative, each Guideline is classified using one of the following
...
 - A1a2++) [EXAMPLE] If a round has a cumulative time limit of 30 minutes and the competitor's results of the first two attempts are 6:00 and DNF (10:00), then the time limit for the third attempt is 30:00 - 6:00 - 10:00 = 14:00. For this particular example, note that only the final result for an attempt counts. Therefore, a third attempt with an original recorded time of 13:59 without penalties meets the time limit because the final result is 13:59 (see [Regulation 9f2](regulations:regulation:9f2)), while a third attempt with a result of 13:59 + 2 = 14:01 does not meet the time limit.
 - A1a2+++) [ADDITION] If the result from an attempt is unavailable (e.g. the timer was reset), the judge should consult the WCA Delegate and make an estimate of the elapsed time for the attempt. This estimate counts towards the time limit, but must not be used as the basis of an official result for the individual attempt.
+- A1a2++++) [CLARIFICATION] In case of a cumulative time limit, the result of each remaining attempt in the round is recorded as DNS (Did Not Start) after a competitor reaches the time limit.
```
Samuel-Baird: what is currently done to handle this? It seems a bit strange to record it as a DNS since a DNS is defined as a competitor being eligible for an attempt but declining it. The competitor simply isn't eligible for the extra attempt since they have run out of time. 

If this clarification is added then should we also record attempts that the competitor is not eligible to do, due to missing cutoffs, as DNSs?

SAuroux: > The competitor simply isn't eligible for the extra attempt since they have run out of time.

This implication is simply not correct and not backed up by the regulations. In a round without cutoff, every competitor is always eligible for all attempts.

Nevseros: This guideline was indeed proposed to be re-added as without it the way these attempts are stored in the DB doesn't match the regulations, there is no mistake, check the original Forum thread for more information, please.

```suggestion
- A1a2+++++) [CLARIFICATION] In case of a cumulative time limit, the result of each remaining attempt in the round is recorded as DNS (Did Not Start) after a competitor reaches the cumulative time limit.
```

Rouxles: Potentially adding cumulative here because the addition of having rounds with both a cutoff and a cumulative time limit here might change things a bit. (I don't think this will end up mattering, but just putting it out there in case somebody has something to say)

PR #1051, opened 2022-10-10
Add clarification on judges forgetting to put a sight blocker

Nevseros:
Adds Guideline B4c+ based on the recent rulings.

Completely unsure about the wording and don't like the fact it looks almost the same as the [guideline](https://www.worldcubeassociation.org/regulations/guidelines.html#A2d1++) about scramble signatures thus making it look to have the same level of importance.


Nevseros:
Apologies, my inexperience with git led to this little mess up. Going to close this PR and open the correct one.

PR #1052, opened 2022-10-10
Add clarification on judges forgetting to put a sight blocker

Nevseros:
Adds Guideline B4c+ based on the recent rulings.

Completely unsure about the wording and don't like the fact it looks almost the same as the [guideline](https://www.worldcubeassociation.org/regulations/guidelines.html#A2d1++) about scramble signatures thus making it look to have the same level of importance.

### DIFF
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -180,6 +180,7 @@ To be more informative, each Guideline is classified using one of the following 
 - B1+) [REMINDER] The competitor must use a puzzle without textures, markings, or other features that distinguish similar pieces (see [Regulation 3k](regulations:regulation:3k)). This should be given special attention for Blindfolded Solving.
 - B1b+) [RECOMMENDATION] Blindfolds should be checked by the WCA Delegate before use in the competition.
 - B2d+) [ADDITION] By default, the competitor starts the solve the first time that they start the timer after the judge has indicated that they are ready (see [Regulation B2a](regulations:regulation:B2a)). If they want to check that the timer is in working order, they must (ask and) receive confirmation from the judge each time before starting/resetting the timer during this phase. Penalty for starting and resetting a timer without confirmation from the judge: disqualification of the attempt (DNF).
+- B4c+) [ADDITION] An attempt should not be disqualified solely due to the judge not putting a sight blocker. However, the WCA Delegate may replace the attempt with an extra attempt if they have a reason to suspect that the competitor gained an unfair advantage.
 
 
 ## <article-C><one-handed><onehandedsolving> Article C: One-Handed Solving
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -180,7 +180,8 @@ To be more informative, each Guideline is classified using one of the following 
 - B1+) [REMINDER] The competitor must use a puzzle without textures, markings, or other features that distinguish similar pieces (see [Regulation 3k](regulations:regulation:3k)). This should be given special attention for Blindfolded Solving.
 - B1b+) [RECOMMENDATION] Blindfolds should be checked by the WCA Delegate before use in the competition.
 - B2d+) [ADDITION] By default, the competitor starts the solve the first time that they start the timer after the judge has indicated that they are ready (see [Regulation B2a](regulations:regulation:B2a)). If they want to check that the timer is in working order, they must (ask and) receive confirmation from the judge each time before starting/resetting the timer during this phase. Penalty for starting and resetting a timer without confirmation from the judge: disqualification of the attempt (DNF).
-- B4c+) [ADDITION] An attempt should not be disqualified or replaced with an extra attempt solely due to the judge not putting a sight blocker. However, the WCA Delegate may replace the attempt with an extra attempt if they have a reason to suspect that the competitor gained an unfair advantage. Exception: if the competitor had to put a sight blocker themselves (see [Regulation B4c3](regulations:regulation:B4c3)), the attempt should be disqualified, at the discretion of the WCA Delegate.
+- B4c+) [ADDITION] if the judge forgot to put a sight blocker (or did not put it immediately), the WCA Delegate should always grant an extra attempt.
+- B4c++) [ADDITION] If the WCA Delegate does not suspect any infraction of the regulations by the competitor, the original attempt may stand.
 
 
 ## <article-C><one-handed><onehandedsolving> Article C: One-Handed Solving
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -181,7 +181,7 @@ To be more informative, each Guideline is classified using one of the following 
 - B1b+) [RECOMMENDATION] Blindfolds should be checked by the WCA Delegate before use in the competition.
 - B2d+) [ADDITION] By default, the competitor starts the solve the first time that they start the timer after the judge has indicated that they are ready (see [Regulation B2a](regulations:regulation:B2a)). If they want to check that the timer is in working order, they must (ask and) receive confirmation from the judge each time before starting/resetting the timer during this phase. Penalty for starting and resetting a timer without confirmation from the judge: disqualification of the attempt (DNF).
 - B4c+) [ADDITION] If the judge forgot to put a sight blocker, or places it in a way that the competitor's view of the puzzle is not fully blocked, the WCA Delegate should grant an extra attempt.
-- B4c++) [ADDITION] If the WCA Delegate does not suspect any infraction of the regulations by the competitor, the original attempt may stand.
+- B4c++) [ADDITION] If the WCA Delegate does not suspect that the competitor intentionally violated any WCA Regulations, the original attempt may stand.
 
 
 ## <article-C><one-handed><onehandedsolving> Article C: One-Handed Solving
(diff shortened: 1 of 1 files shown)


Randomno:
Is the purpose of this guideline to say that a missing sight blocker should not result in a DNF, or just that it should not result in an extra unless there is a suspected advantage?

Nevseros:
Missing sight blocker should not result in an extra unless an advantage was gained. I'm not sure we should we have to emphasize the 1st point you mentioned.

wca-guidelines.md
```diff
@@ -180,6 +180,7 @@ To be more informative, each Guideline is classified using one of the following
 - B1+) [REMINDER] The competitor must use a puzzle without textures, markings, or other features that distinguish similar pieces (see [Regulation 3k](regulations:regulation:3k)). This should be given special attention for Blindfolded Solving.
 - B1b+) [RECOMMENDATION] Blindfolds should be checked by the WCA Delegate before use in the competition.
 - B2d+) [ADDITION] By default, the competitor starts the solve the first time that they start the timer after the judge has indicated that they are ready (see [Regulation B2a](regulations:regulation:B2a)). If they want to check that the timer is in working order, they must (ask and) receive confirmation from the judge each time before starting/resetting the timer during this phase. Penalty for starting and resetting a timer without confirmation from the judge: disqualification of the attempt (DNF).
+- B4c+) [ADDITION] An attempt should not be disqualified solely due to the judge not putting a sight blocker. However, the WCA Delegate may replace the attempt with an extra attempt if they have a reason to suspect that the competitor gained an unfair advantage.
```
lgarron: I think this needs to make explicit that this does *not* apply if https://www.worldcubeassociation.org/regulations/#B4c3 is in effect (in which case the judge was not expected to use a sight blocker, and the competitor is responsible).

Nevseros: I actually thought about it, but I can't figure out how to clarify it correctly. Maybe an exception is needed here, like:

```
... the competitor gained an unfair advantage. Exception: if the competitor had to put a sight blocker themselves (see [Regulation B4c3](regulations:regulation:B4c3)), the attempt should be disqualified, at the discretion of the WCA Delegate.
````

And I still don't like this wording.

Comment (Nanush7) [reply wca-guidelines.md]: > thus making it look to have the same level of importance.

What would be the problem with that?

Comment (Nevseros) [reply wca-guidelines.md]: > What would be the problem with that?

A competition with all scramblers signatures missing won't probably cause a debate on whether the results should stay or not, but I don't think it would be fine to have the same for the sight blockers.

Comment (Nanush7) [reply wca-guidelines.md]: Understood. What about changing "should not be disqualified" to "May remain valid" (or something like that)? With that wording, it would be a possibility, instead of a recommendation.

Comment (Nevseros) [reply wca-guidelines.md]: It would sound a way better. But maybe it should even be reworded to make it an exception? To clearly define that it's not a normal case and should be trated seriously.

**Review by Nanush7 — DISMISSED** [review]
What do you think about this approach? Focus on the idea and not the (bad) wording :p
```
B4c+) [ADDITION] if the judge forgot to put a sight blocker (or did not put it immediately), the WCA Delegate should always grant an extra attempt.
B4c++) [ADDITION] If an extra attempt is not available, the result may remain valid under exceptional circumstances, at the discretion of the WCA Delegate.
```

**Comment by Nevseros** [comment]
It does not fit the rulings. I feel it's needed to be discussed internally, not here.

File: wca-guidelines.md [hunk 2]
```diff
@@ -180,6 +180,7 @@ To be more informative, each Guideline is classified using one of the following
...
 - B1b+) [RECOMMENDATION] Blindfolds should be checked by the WCA Delegate before use in the competition.
 - B2d+) [ADDITION] By default, the competitor starts the solve the first time that they start the timer after the judge has indicated that they are ready (see [Regulation B2a](regulations:regulation:B2a)). If they want to check that the timer is in working order, they must (ask and) receive confirmation from the judge each time before starting/resetting the timer during this phase. Penalty for starting and resetting a timer without confirmation from the judge: disqualification of the attempt (DNF).
+- B4c+) [ADDITION] An attempt should not be disqualified or replaced with an extra attempt solely due to the judge not putting a sight blocker. However, the WCA Delegate may replace the attempt with an extra attempt if they have a reason to suspect that the competitor gained an unfair advantage. Exception: if the competitor had to put a sight blocker themselves (see [Regulation B4c3](regulations:regulation:B4c3)), the attempt should be disqualified, at the discretion of the WCA Delegate.
```
```suggestion
- B4c+) [ADDITION] if the judge forgot to put a sight blocker (or did not put it immediately), the WCA Delegate should always grant an extra attempt.
- B4c++) [ADDITION] If the WCA Delegate does not suspect any infraction of the regulations by the competitor, the original attempt may stand.
```

Rouxles: @lgarron can you think of alternate wording here? i don't enjoy the wording of this too much, but i think the idea is there

OJCubing: "... does not suspect any infraction of the regulations..." might sound less clunky

Samuel-Baird:
The current use of "put" and the specific language of the judge forgetting is a bit strange. Also, the current language uses "should always" and needs to be changed to "should" or "must" for clarity. 

It would also be great to expand this to include when sight blockers are used but in a way that they do not properly block the competitor's view of the puzzle. How is something like the following:

`If a sight blocker is not properly placed between the competitor's face and the puzzle by the judge (such that the competitor's view of the puzzle would still clearly be blocked if the blindfold was not properly worn) an extra attempt should be granted by the WCA Delegate. `

wca-guidelines.md
```diff
@@ -180,6 +180,8 @@ To be more informative, each Guideline is classified using one of the following
 - B1+) [REMINDER] The competitor must use a puzzle without textures, markings, or other features that distinguish similar pieces (see [Regulation 3k](regulations:regulation:3k)). This should be given special attention for Blindfolded Solving.
 - B1b+) [RECOMMENDATION] Blindfolds should be checked by the WCA Delegate before use in the competition.
 - B2d+) [ADDITION] By default, the competitor starts the solve the first time that they start the timer after the judge has indicated that they are ready (see [Regulation B2a](regulations:regulation:B2a)). If they want to check that the timer is in working order, they must (ask and) receive confirmation from the judge each time before starting/resetting the timer during this phase. Penalty for starting and resetting a timer without confirmation from the judge: disqualification of the attempt (DNF).
+- B4c+) [ADDITION] If the judge forgot to put a sight blocker (or did not put it immediately), the WCA Delegate should always grant an extra attempt.
```
```suggestion
- B4c+) [ADDITION] If the judge forgot to put a sight blocker, or places it in a way that the competitor's view of the puzzle is not fully blocked, the WCA Delegate should grant an extra attempt.
```

Rouxles: changed must -> should to account for the fact that it's not a very good idea to give an extra if the competitor DNFs anyways

EdHollingdale: I much prefer this wording. The "immediately" doesn't really matter except for top 3bld solvers, and the "always" doesn't sit well with me.

wca-guidelines.md
```diff
@@ -180,6 +180,8 @@ To be more informative, each Guideline is classified using one of the following
 - B1+) [REMINDER] The competitor must use a puzzle without textures, markings, or other features that distinguish similar pieces (see [Regulation 3k](regulations:regulation:3k)). This should be given special attention for Blindfolded Solving.
 - B1b+) [RECOMMENDATION] Blindfolds should be checked by the WCA Delegate before use in the competition.
 - B2d+) [ADDITION] By default, the competitor starts the solve the first time that they start the timer after the judge has indicated that they are ready (see [Regulation B2a](regulations:regulation:B2a)). If they want to check that the timer is in working order, they must (ask and) receive confirmation from the judge each time before starting/resetting the timer during this phase. Penalty for starting and resetting a timer without confirmation from the judge: disqualification of the attempt (DNF).
+- B4c+) [ADDITION] If the judge forgot to put a sight blocker, or places it in a way that the competitor's view of the puzzle is not fully blocked, the WCA Delegate should grant an extra attempt.
+- B4c++) [ADDITION] If the WCA Delegate does not suspect any infraction of the regulations by the competitor, the original attempt may stand.
```
lgarron: Nit: I'd avoid "infraction"  for international audiences, in favor of something more direct:
```suggestion
- B4c++) [ADDITION] If the WCA Delegate does not suspect that the competitor intentionally violated any WCA Regulations, the original attempt may stand.
```

Nevseros: Shouldn't `any Regulations` here be downcase, i.e. `any regulations`?

lgarron: Either is technically fine (sort of how a Delegate is also a delegate), I tend to uppercase.

Nevseros: But for the delegate we always use the uppercase, while for multiple/undefined regulations downcase (e.g. [5b5](https://www.worldcubeassociation.org/regulations/#5b5)).

Rouxles: Updated comment to say WCA Regulations here. (I have seen capital R be preceded by WCA more often than not)

Rouxles (approved):

PR #1053, opened 2022-10-17
Update 2j/2k

Nanush7:
Update disqualification-related regulations, based on WRT's opinion.

### DIFF
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -57,6 +57,7 @@ To be more informative, each Guideline is classified using one of the following 
 - 2i2a+) [ADDITION] A monitor of a camera can be visible to the competitor as long as it does not show the live feed of the camera.
 - 2j2+) [EXAMPLE] For example, if a competitor is disqualified from an event for failing to show up for the final round, their results from earlier rounds remain valid.
 - 2j2++) [EXAMPLE] If the WCA Delegate disqualifies a competitor during their third attempt in a round, only the third attempt and all following attempts in that event are disqualified, even if the circumstances that originated the disqualification occurred prior to this attempt.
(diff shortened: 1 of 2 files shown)


wca-guidelines.md
```diff
@@ -57,6 +57,7 @@ To be more informative, each Guideline is classified using one of the following
 - 2i2a+) [ADDITION] A monitor of a camera can be visible to the competitor as long as it does not show the live feed of the camera.
 - 2j2+) [EXAMPLE] For example, if a competitor is disqualified from an event for failing to show up for the final round, their results from earlier rounds remain valid.
 - 2j2++) [EXAMPLE] If the WCA Delegate disqualifies a competitor during their third attempt in a round, only the third attempt and all following attempts in that event are disqualified, even if the circumstances that originated the disqualification occurred prior to this attempt.
+- 2j2+++) [CLARIFICATION] Specific completed attempts may be disqualified under the circumstances described in [Regulation 2k](regulations:regulation:2k), only if the relevant incident ocurred during the attempt.
```
Nanush7: @SAuroux Please, let me know if this is in line with what has been discussed.

SAuroux: Looks ok to me @Nanush7 .

@thewca/wrt can at least 1-2 others double-check please?

lgarron: I think this is a strong enough override that I think it should go into the Regs. What do you think about folding it into 2j, or as 2j3?

Nanush7: Agreed 👍

ohexter: I'm happy with the sentiment, but this seems like an exception to the updated 2j2, not merely a clarification. I would suggest removing the word "only" and moving this items to 2j2a within the regulations.

Nanush7: Thank you for your comment! But you are replying to an outdated commit, your concern was already addressed in the latest commit.

JoaoViniciusSantos (approved):

PR #1059, opened 2022-10-17
Change A7g++ so penalties can be waived

Rouxles:


Rouxles:
For reference, I don't feel very strongly in favor of this, but I'm putting this down here just so there's an idea that can possibly be used

JoaoViniciusSantos (approved):

Nanush7 (approved):

wca-guidelines.md
```diff
@@ -170,7 +170,7 @@ To be more informative, each Guideline is classified using one of the following
...
 - A7c+) [ADDITION] If a result signed by a competitor is found to be incomplete or unreadable, it should be interpreted the worst reasonable way (e.g. "1:05" will be considered 1:05.99, "25.X3" will be considered 25.73 if X could be either a 1 or a 7). Given that the Stackmat timer displays a time of X:0Y.ss as X:Y.ss, a missing tens digit should be considered more likely than a missing ones digit (e.g. "1:2.27" will be considered 1:02.27). The judge who signed an ambiguous result may be requested to clarify it, at the discretion of the score taker.
 - A7g+) [EXAMPLE] Examples of incidents that could grant a new competitor an extra attempt: not waiting for the green light, improperly starting or stopping the timer, going over inspection time due to misunderstanding how to start a solve, or other procedural time penalties.
-- A7g++) [CLARIFICATION] Individual penalties cannot be waived for new competitors. A full extra attempt must be granted.
+- A7g++) [CLARIFICATION] Individual penalties should not be waived for new competitors unless it is impossible to grant a full extra attempt.
```
Rouxles, line 173: Possible new wording? I feel like must not would still communicate the same idea because it's paired with unless

~~I do remember something about `unless` not being in the regulations from the last cycle, but I need to double check that~~ Double checked, this should be fine

```suggestion
- A7g++) [CLARIFICATION] Individual penalties for inexperienced competitors are not permitted to be waived unless it is impossible to grant a full extra attempt.
```

Rouxles:
Not being implemented due to the possible abuse this could cause

PR #1062, opened 2022-10-19
Define part in 5b6

Rouxles:
Copied directly from the style guide, probably needs rewording/rethinking

Also perhaps `piece` needs to be defined somewhere else as well?

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -192,6 +192,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - 5b5c) If more than one part with one colored face is affected, the puzzle is considered unsolved (DNF).
         - 5b5d) If one or more parts with more than one colored face are affected, the puzzle is considered unsolved (DNF).
         - 5b5e) Regulations [5b5c](regulations:regulation:5b5c) and [5b5d](regulations:regulation:5b5d) supersede [5b5a](regulations:regulation:5b5a) and [5b5b](regulations:regulation:5b5b).
+    - 5b6) Definition: A part is a subset of puzzle components (in particular, any individual piece, cap, sticker, etc.).
 
 
 ## <article-7><environment><environment> Article 7: Environment
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -192,7 +192,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - 5b5c) If more than one part with one colored face is affected, the puzzle is considered unsolved (DNF).
         - 5b5d) If one or more parts with more than one colored face are affected, the puzzle is considered unsolved (DNF).
         - 5b5e) Regulations [5b5c](regulations:regulation:5b5c) and [5b5d](regulations:regulation:5b5d) supersede [5b5a](regulations:regulation:5b5a) and [5b5b](regulations:regulation:5b5b).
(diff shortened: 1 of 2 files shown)


wca-regulations.md
```diff
@@ -192,6 +192,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - 5b5c) If more than one part with one colored face is affected, the puzzle is considered unsolved (DNF).
         - 5b5d) If one or more parts with more than one colored face are affected, the puzzle is considered unsolved (DNF).
         - 5b5e) Regulations [5b5c](regulations:regulation:5b5c) and [5b5d](regulations:regulation:5b5d) supersede [5b5a](regulations:regulation:5b5a) and [5b5b](regulations:regulation:5b5b).
+    - 5b6) Definition: A part is a subset of puzzle components (in particular, any individual piece, cap, sticker, etc.).
```
lgarron: This looks good, but we don't use "in particular". I'd go for this:
```suggestion
    - 5b6) Definition: A part is a subset of puzzle components that can physically hold together (e.g. an individual piece, cap, single sticker, or sticker attached to another part).
```
I think we should also have an example like "2/3 of a corner", but I can't think of a good way to phrase that.

xsrvmy:
I think there is another ambiguity here: if a center pieces pops on a stickerless cube, and the cap seperates from the stem, this should be considered one colored face, but it could be understood as two. It needs to be made clear that only colored faces that matter for the solved state should be counted.

wca-regulations.md
```diff
@@ -192,6 +192,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - 5b5c) If more than one part with one colored face is affected, the puzzle is considered unsolved (DNF).
         - 5b5d) If one or more parts with more than one colored face are affected, the puzzle is considered unsolved (DNF).
         - 5b5e) Regulations [5b5c](regulations:regulation:5b5c) and [5b5d](regulations:regulation:5b5d) supersede [5b5a](regulations:regulation:5b5a) and [5b5b](regulations:regulation:5b5b).
+    - 5b6) Definition: A part is a subset of puzzle components that can physically hold together (e.g. an individual piece, cap, single sticker, or sticker attached to another part).
```
ohexter:
```suggestion
    - 5b6) Definition: A part is an individual puzzle component (e.g. a single sticker, a 3x3x3 center cap) or a group of puzzle components that remain connected and do not move relative to each other during ordinary operation of the puzzle (e.g. a 3x3 corner piece, a pyraminx tip).
```

I think this is a little clearer in terms of what the change appears trying to accomplish. I am also concerned about how considering both smaller components and larger parts as 'parts' could impact the application of 5b5, where a defect could be analysed differently depending on the level considered.

Rouxles: I think this is pretty clear - there's still a weird edge case of having 2/3 of a corner piece, which is still considered a part though. Do you have any alternative wording for that?

ohexter:
```suggestion
    - 5b6) Definition: A part is one or more fully connected components from a group of puzzle components that remain connected and do not move relative to each other during ordinary operation of the puzzle (e.g. a 3x3x3 centre cap, a pyraminx tip, a 3x3x3 corner piece missing one face, one face of a 3x3x3 corner piece).
```

This is getting super-awkward, but I've given it a go anyway.

Nevseros: Maybe it's worth putting it into a separate guideline? I feel like there are too much examples for the regulation, but at the same time they all are important to understand the meaning.

Rouxles: i think having a lot of examples is fine for this

Nevseros: Yeah, I mean the examples are fine, but here's a bunch of them. To not make regs longer and longer, maybe just create a guideline with the label `[EXAMPLE]`?

CarterKoala (approved):

PR #1064, opened 2022-10-20
License

### LINKED ISSUE #639: License
Something just reminded me that the Regulations don't have a license.

We should probably use a Creative Commons license: https://creativecommons.org/choose/

--------

If we want others (say, Red Bull) to be able to adapt our wording:  https://creativecommons.org/licenses/by-sa/2.0/

If we want to allow copies of the WCA Regulations to be distributed but attributed (and not modified): https://creativecommons.org/licenses/by-nc-nd/4.0/

Note: we should also figure out how this applies to translations. Contributions on GitHub fall under terms that are equivalent to a CLA, but we need to be able to handle 1) contributions through other media 2) the fact that we don't have formal licenses for past contributions.

Nanush7:
Closes #639 

Add [CC BY-NC-ND 4.0](https://creativecommons.org/licenses/by-nc-nd/4.0/legalcode) license.

This is strongly linked to thewca/wca-regulations-translations#129. Please, have a look at that PR for further information.

Requesting review from @thewca/board, as this is legal stuff that may concern you.

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -26,6 +26,9 @@ WCA Regulations in [PDF format](link:pdf)
 ### Source
 Development of the WCA Regulations and Guidelines is public [on GitHub](https://github.com/thewca/wca-regulations) and the discussion is public [on the WCA Forum](https://forum.worldcubeassociation.org/c/regulations).
 
+### License
+This document is licensed under a [CC BY-NC-ND 4.0](https://creativecommons.org/licenses/by-nc-nd/4.0/) license. Permissions beyond the scope of this license may be available at [thewca/wca-regulations-translations](https://github.com/thewca/wca-regulations-translations#license)
+
 ### Contact
 For questions and feedback, please contact the [WCA Regulations Committee (WRC)](mailto:wrc@worldcubeassociation.org).
 
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -9,6 +9,9 @@
 
 The WCA Guidelines supplement the [WCA Regulations](regulations:top). Please see the Regulations for more information about the WCA.
 
+### License
+This document is licensed under a [CC BY-NC-ND 4.0](https://creativecommons.org/licenses/by-nc-nd/4.0/) license. Permissions beyond the scope of this license may be available at [thewca/wca-regulations-translations](https://github.com/thewca/wca-regulations-translations#license).
+
 ### Numbering
 
 Guidelines are numbered in correspondence with related regulations. Note that multiple Guidelines may correspond to the same Regulation, and some Guidelines correspond to Regulations that do not exist anymore.


wca-regulations.md
```diff
@@ -26,6 +26,9 @@ WCA Regulations in [PDF format](link:pdf)
 ### Source
 Development of the WCA Regulations and Guidelines is public [on GitHub](https://github.com/thewca/wca-regulations) and the discussion is public [on the WCA Forum](https://forum.worldcubeassociation.org/c/regulations).
 
+### License
+This document is licensed under a [CC BY-NC-ND 4.0](https://creativecommons.org/licenses/by-nc-nd/4.0/) license. Permissions beyond the scope of this license may be available at [thewca/wca-regulations-translations](https://github.com/thewca/wca-regulations-translations#license)
```
lgarron, line 30: Since this text appears on the main Regulations, I think we should avoid linking to something else that can imply the availability of a looser license.

```suggestion
This document is licensed under a [CC BY-NC-ND 4.0](https://creativecommons.org/licenses/by-nc-nd/4.0/) license.
```

If we want translations to specify looser languages, we could generalize the mechanism we use for "this translation is not an official version" at https://github.com/thewca/wca-regulations-translations#translation-instructions

lgarron:
I agree that this is a good license to use; I am happy to license my past and future contributions under it.

"Non-commercial" could in principle clash with competitions that make any profit, but 1) this is more regulated now, 2) excerpts are almost certainly fair use all around the world, and 3) if this prohibits someone from hosting a duplicate copy of the full Regulations for some commercial purpose, that's probably actually a good thing (they should be linking to an official version, or contacting the WCA for alternatives).

We should probably also consider a full assignment of rights to the WCA of any contributions.

wca-regulations.md
Rouxles: is this a change for this regulations cycle iyo? @lgarron

Nanush7: This is the common procedure that Creative Commons provides to grant more rights (in this case, the right to distribute a translation). See "Simplest CC+ Example" [here](https://wiki.creativecommons.org/wiki/CCPlus).

However, we may simply use the CC BY-NC-ND 4.0 and only allow translations to be posted in our website.

Nanush7:
@lgarron 
> We should probably also consider a full assignment of rights to the WCA of any contributions.

Can we do that alongside a CC license?

lgarron:
> @lgarron
> 
> > We should probably also consider a full assignment of rights to the WCA of any contributions.
> 
> Can we do that alongside a CC license?

Yeah, that's usually through a separate (but related) like a [CLA](https://en.wikipedia.org/wiki/Contributor_License_Agreement). GitHub's terms automatically include a CLA for pull requests for the project's current license, but we can ask for contribution rights to be assigned to the WCA so that we can relicense in the future if we need to.

Nanush7:
> Yeah, that's usually through a separate (but related) like a [CLA](https://en.wikipedia.org/wiki/Contributor_License_Agreement). GitHub's terms automatically include a CLA for pull requests for the project's current license, but we can ask for contribution rights to be assigned to the WCA so that we can relicense in the future if we need to.

Ok, I think it may be a good idea. I'm not familiar with CLAs, though. Would you handle that?

PR #1065, opened 2022-10-21
O-ring guideline

### LINKED ISSUE #265: O-Ring Information
Would it be appropriate to put the oring size information on the guidelines? I feel like this would make the information more readily available and I wouldn't have to search my email for it.

Nanush7:
Add O-ring size recommendation.

Closes #265


lgarron:
Could this maybe go into the competition requirements policy? The size is a detail that competitors and judges don't need to worry about.

Nanush7:
> The size is a detail that competitors and judges don't need to worry about.

Hmm, you're right. But I'm not sure if the WCRP is the correct place either. Maybe in the organiser guidelines? @thewca/wqac

Nanush7:
The WQAC will handle this.

PR #1068, opened 2022-10-25
Added clarifying regulation for FMC

Rouxles:
Added as a new regulation (E5) rather than a clarification because I think this is significant enough of a change.

This also needs rewording - someone give me a suggestion for wording and we'll go from there

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -507,6 +507,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - E3e) Other non-electronic aids that do not give an unfair advantage, at the discretion of the WCA Delegate (see [Regulation 2i1](regulations:regulation:2i1)).
     - E3f) Tools to remove markings (self-supplied), such as correction fluid, correction tape, or erasers.
 - E4) The WCA Delegate may choose to publish the solutions (e.g. photo of the score sheets or transcriptions of solutions).
+- E5) The competitor may communicate with the WCA delegate during the attempt. The WCA delegate may answer questions from competitors as long as the contents are strictly based on the WCA Regulations.
 
 
 ## <article-F><clock><clocksolving> Article F: Clock Solving
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -507,7 +507,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - E3e) Other non-electronic aids that do not give an unfair advantage, at the discretion of the WCA Delegate (see [Regulation 2i1](regulations:regulation:2i1)).
     - E3f) Tools to remove markings (self-supplied), such as correction fluid, correction tape, or erasers.
 - E4) The WCA Delegate may choose to publish the solutions (e.g. photo of the score sheets or transcriptions of solutions).
-- E5) The competitor may communicate with the WCA delegate during the attempt. The WCA delegate may answer questions from competitors as long as the contents are strictly based on the WCA Regulations.
+- E5) The competitor is permitted to communicate with the WCA Delegate during the attempt. The WCA Delegate may provide information to competitors as long as the information can be found in the WCA Regulations.
 
 
 ## <article-F><clock><clocksolving> Article F: Clock Solving
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -507,7 +507,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - E3e) Other non-electronic aids that do not give an unfair advantage, at the discretion of the WCA Delegate (see [Regulation 2i1](regulations:regulation:2i1)).
     - E3f) Tools to remove markings (self-supplied), such as correction fluid, correction tape, or erasers.
 - E4) The WCA Delegate may choose to publish the solutions (e.g. photo of the score sheets or transcriptions of solutions).
-- E5) The competitor must not communicate with anyone other than the WCA Delegate. This communication should be about the WCA Regulations and the competitor must not receive assistance deriving their solution. Penalty: disqualification of the attempt (DNF).
+- E5) The competitor must not communicate with anyone other than the WCA Delegate or the judge. This communication should be about the WCA Regulations and the competitor must not receive assistance deriving their solution. Penalty: disqualification of the attempt (DNF).
 
 
 ## <article-F><clock><clocksolving> Article F: Clock Solving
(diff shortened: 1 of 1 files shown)


JoaoViniciusSantos:
Maybe: The competitor is permitted to ask questions or communicate with the WCA Delegate during the attempt. The WCA Delegate may answer questions from competitors as long as the contents are strictly based on the WCA Regulations and do not give an unfair advantage.

Nevseros:
A minor note: `WCA Delegate` must be capitalized.

> The competitor is permitted to ask questions or communicate with the WCA Delegate during the attempt.

Asking questions $\in$ communicating, I don't think it's necessary to mention `asking questions`.  

FWIW, I think it also makes sense to allow competitors to communicate with the judge, to allow the Delegates to answer not only about the Regulations, but about other WCA documents too (i.e. about the CoC when the competitor hears another competitor possibly violating it). I pretty much like the wording with an `unfair advantage`, and I'd even leave only this part (i.e. without anything about WCA Regulations) and add an `[EXAMPLE]` guideline.

wca-regulations.md
```diff
@@ -507,6 +507,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - E3e) Other non-electronic aids that do not give an unfair advantage, at the discretion of the WCA Delegate (see [Regulation 2i1](regulations:regulation:2i1)).
     - E3f) Tools to remove markings (self-supplied), such as correction fluid, correction tape, or erasers.
 - E4) The WCA Delegate may choose to publish the solutions (e.g. photo of the score sheets or transcriptions of solutions).
+- E5) The competitor may communicate with the WCA delegate during the attempt. The WCA delegate may answer questions from competitors as long as the contents are strictly based on the WCA Regulations.
```
```suggestion
- E5) The competitor is permitted to communicate with the WCA Delegate during the attempt. The WCA Delegate may provide information to competitors as long as the information can be found in the WCA Regulations.
```

```suggestion
- E5) The competitor is permitted to communicate with the WCA Delegate during the attempt. The WCA Delegate may answer questions from competitors as long as the contents are strictly based on the WCA Regulations.
```

```suggestion
- E5) The competitor is permitted to communicate with the WCA Delegate during the attempt. The WCA Delegate may provide information that can be found in the WCA Regulations to competitors.
```

Rouxles: these are alternate wordings that might not necessarily be the ones that are used

Rouxles: these are alternate wordings that might not necessarily be the ones that are used

wca-regulations.md
```diff
@@ -507,7 +507,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - E3e) Other non-electronic aids that do not give an unfair advantage, at the discretion of the WCA Delegate (see [Regulation 2i1](regulations:regulation:2i1)).
     - E3f) Tools to remove markings (self-supplied), such as correction fluid, correction tape, or erasers.
 - E4) The WCA Delegate may choose to publish the solutions (e.g. photo of the score sheets or transcriptions of solutions).
-- E5) The competitor may communicate with the WCA delegate during the attempt. The WCA delegate may answer questions from competitors as long as the contents are strictly based on the WCA Regulations.
+- E5) The competitor is permitted to communicate with the WCA Delegate during the attempt. The WCA Delegate may provide information to competitors as long as the information can be found in the WCA Regulations.
```
```suggestion
- E5) The competitor must not communicate with anyone other than the WCA Delegate. This communication must be about the WCA Regulations. Penalty: disqualification of the attempt (DNF).
```
Another one :p
(see [A5a](https://www.worldcubeassociation.org/regulations/#A5a))

sam596:
How about: "The competitor must not receive assistance from anyone (or any object)* other than the WCA Delegate. Such assistance must only be with regards to the content of the WCA Regulations"

\* (maybe omit these 3 words?)

This wording puts the first half more in line with A5b, and I find "this communication must be about" a bit of an awkward arrangement.

wca-regulations.md
```diff
@@ -507,6 +507,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - E3e) Other non-electronic aids that do not give an unfair advantage, at the discretion of the WCA Delegate (see [Regulation 2i1](regulations:regulation:2i1)).
     - E3f) Tools to remove markings (self-supplied), such as correction fluid, correction tape, or erasers.
 - E4) The WCA Delegate may choose to publish the solutions (e.g. photo of the score sheets or transcriptions of solutions).
+- E5) The competitor must not communicate with anyone other than the WCA Delegate. This communication must be about the WCA Regulations. Penalty: disqualification of the attempt (DNF).
```
JonEsparaz: If the competitor asked, "can you [the Delegate] please close the door?" or "can I use the washroom", I don't think these should warrant a DNF, but the communication is technically _not_ about the WCA Regulations.

Perhaps the following makes the intent clearer:

```suggestion
- E5) The competitor must not communicate with anyone other than the WCA Delegate. This communication should be about the WCA Regulations and the competitor must not receive assistance deriving their solution. Penalty: disqualification of the attempt (DNF).
```

This wording could be tricky. For example, a Delegate can clarify notation to a competitor, which assists them in writing their solution, but not deriving it.

lgarron:
I think this is a good change, although we may want to find something just a bit more accommodating.

If someone goes up to the ask the Delegate about notation and then also asks "multi-blind is the same room immediately after this, right?" before returning to their seat, I don't think anyone would argue that it would be unfair for the Delegate to provide that information. (I wouldn't *encourage* it, but I'd like to avoid the need for discussions about such incidents where the competitor clearly did not obtain an advantage for their FMC solution.)

Rouxles:
> I think this is a good change, although we may want to find something just a bit more accommodating.
> 
> If someone goes up to the ask the Delegate about notation and then also asks "multi-blind is the same room immediately after this, right?" before returning to their seat, I don't think anyone would argue that it would be unfair for the Delegate to provide that information. (I wouldn't _encourage_ it, but I'd like to avoid the need for discussions about such incidents where the competitor clearly did not obtain an advantage for their FMC solution.)

Hmm... I think the last part is fine mainly because of *should*, rather than *must* being used for the first clause, and the second clause only for deriving a solution (which knowing the schedule for, or something similar to that, won't be of any help)

wca-regulations.md
```diff
@@ -507,6 +507,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - E3e) Other non-electronic aids that do not give an unfair advantage, at the discretion of the WCA Delegate (see [Regulation 2i1](regulations:regulation:2i1)).
     - E3f) Tools to remove markings (self-supplied), such as correction fluid, correction tape, or erasers.
 - E4) The WCA Delegate may choose to publish the solutions (e.g. photo of the score sheets or transcriptions of solutions).
+- E5) The competitor must not communicate with anyone other than the WCA Delegate. This communication should be about the WCA Regulations and the competitor must not receive assistance deriving their solution. Penalty: disqualification of the attempt (DNF).
```
```suggestion
- E5) The competitor must not communicate with anyone other than the WCA Delegate or the judge. This communication should be about the WCA Regulations and the competitor must not receive assistance deriving their solution. Penalty: disqualification of the attempt (DNF).
```

Shouldn't we also allow the competitors to communicate with the judge of the attempt? It seems reasonable for i.e. asking about time left.

Rouxles: I think it's pretty reasonable here

Nevseros (approved):

PR #1070, opened 2022-10-25
Add 4g2 for clock pins

Rouxles:
This one is more of a controversial change, but I'm not fully against it.

### DIFF
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -94,6 +94,9 @@ To be more informative, each Guideline is classified using one of the following 
 - 4d++) [ADDITION] It is permitted for the puzzle to change its orientation when it is moved from the scrambler to the solving station, as long as no one is attempting to influence the randomness of the orientation (see [Regulation A2e1](regulations:regulation:A2e1)).
 - 4f+) [RECOMMENDATION] The WCA Delegate should generate sufficient scramble sequences for the entire competition ahead of time, including spare scramble sequences for extra attempts.
 - 4f++) [REMINDER] If the WCA Delegate generates any additional scramble sequences during the competition, the scramble sequences must be saved.
+- 4g2+) [CLARIFICATION] The scrambler must still verify that the clock is scrambled correctly.
+- 4g2++) [ADDITION] If the clock falls over during uncovering, an extra must be given.
+- 4g2+++) [RECOMMENDATION] The scrambler should leave score sheets outside the cube cover instead of placing them inside the cube cover.
 
 
 ## <article-5><puzzle-defects><puzzledefects> Article 5: Puzzle Defects
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -95,7 +95,7 @@ To be more informative, each Guideline is classified using one of the following 
 - 4f+) [RECOMMENDATION] The WCA Delegate should generate sufficient scramble sequences for the entire competition ahead of time, including spare scramble sequences for extra attempts.
 - 4f++) [REMINDER] If the WCA Delegate generates any additional scramble sequences during the competition, the scramble sequences must be saved.
 - 4g2+) [CLARIFICATION] The scrambler must still verify that the clock is scrambled correctly.
-- 4g2++) [ADDITION] If the clock falls over during uncovering, an extra must be given.
+- 4g2++) [CLARIFICATION] If the clock falls over during uncovering, an extra must be given.
 - 4g2+++) [RECOMMENDATION] The scrambler should leave score sheets outside the cube cover instead of placing them inside the cube cover.
 
 
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -173,7 +173,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 - 4f) Competition scramble sequences must be generated using a current official version of an official WCA scramble program (available [via the WCA website](https://www.worldcubeassociation.org/regulations/scrambles/)).
 - 4g) After scrambling a puzzle, the scrambler must verify that the puzzle is scrambled correctly. If the puzzle state is wrong, the scrambler must correct it (e.g. by solving the puzzle and applying the scramble sequence again).
     - 4g1) Exception: For the 6x6x6 Cube, 7x7x7 Cube, and Megaminx, it is not necessary to correct the puzzle state, at the discretion of the WCA Delegate.
-    - 4g2) Exception: For Clock, if the state of the pins are affected during transportation or uncovering, it is not necessary to correct the pins, at the discretion of the WCA Delegate.
+    - 4g2) Exception: For Clock, if the state of the pins are affected during transportation or uncovering, the pins should not be corrected.
 
 
 ## <article-5><puzzle-defects><puzzledefects> Article 5: Puzzle Defects
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -173,6 +173,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 - 4f) Competition scramble sequences must be generated using a current official version of an official WCA scramble program (available [via the WCA website](https://www.worldcubeassociation.org/regulations/scrambles/)).
 - 4g) After scrambling a puzzle, the scrambler must verify that the puzzle is scrambled correctly. If the puzzle state is wrong, the scrambler must correct it (e.g. by solving the puzzle and applying the scramble sequence again).
     - 4g1) Exception: For the 6x6x6 Cube, 7x7x7 Cube, and Megaminx, it is not necessary to correct the puzzle state, at the discretion of the WCA Delegate.
+    - 4g2) Exception: For Clock, if the state of the pins are wrong, the scrambler should correct the pins, at the discretion of the WCA Delegate.
 
 
 ## <article-5><puzzle-defects><puzzledefects> Article 5: Puzzle Defects
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -173,7 +173,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 - 4f) Competition scramble sequences must be generated using a current official version of an official WCA scramble program (available [via the WCA website](https://www.worldcubeassociation.org/regulations/scrambles/)).
 - 4g) After scrambling a puzzle, the scrambler must verify that the puzzle is scrambled correctly. If the puzzle state is wrong, the scrambler must correct it (e.g. by solving the puzzle and applying the scramble sequence again).
     - 4g1) Exception: For the 6x6x6 Cube, 7x7x7 Cube, and Megaminx, it is not necessary to correct the puzzle state, at the discretion of the WCA Delegate.
-    - 4g2) Exception: For Clock, if the state of the pins are wrong, the scrambler should correct the pins, at the discretion of the WCA Delegate.
+    - 4g2) Exception: For Clock, if the state of the pins are affected during transportation or uncovering, it is not necessary to correct the pins, at the discretion of the WCA Delegate.
 
 
 ## <article-5><puzzle-defects><puzzledefects> Article 5: Puzzle Defects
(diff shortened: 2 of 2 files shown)


Randomno:
I brought this up to a few people and they were in favour of it, however the wording is quite confusing and not strictly accurate if I understand it (since the pins could also be moved by the runner or judge). I haven't thought how else to codify allowing pin misscrambles though.

timreyn:
Hi! Quick question about this PR...

I'm very in favor of relaxing regulations about clock pin position, so I'm very excited to see this being considered for the 2023 regulations! Though I'm a little confused about this particular proposal, it sounds like it's saying scramblers are allowed to send out pin misscrambles only if they confirm with the Delegate. That's not usually the issue with pins -- scramblers don't knowingly send out pin misscrambles, and if they see an issue they'll fix it rather than asking a Delegate. It's more likely that there are issues while running or lifting the cover.

As a (possibly controversial) alternative, would you consider just dropping Article F, or replacing F2 and F3 with a statement that "the pin position is not considered part of the scramble state, and the competitor is permitted to change the position of pins before the start of the solve"? That would mean that:
-The scramble doesn't need to specify a pin position
-Standard covers can be used, rather than clock-specific ones
-During inspection, the competitor is permitted to set the pin position for their first turn
-The competitor can put the puzzle standing, or lying down, but either way they won't be DNFed if the pins change

I think this would be more effective at reducing the challenges of organizing Clock. What do you think?

Thank you!

Rouxles:
> I brought this up to a few people and they were in favour of it, however the wording is quite confusing and not strictly accurate if I understand it (since the pins could also be moved by the runner or judge). I haven't thought how else to codify allowing pin misscrambles though.

I agree that it's pretty confusing at the moment, but we'll be coming up with a better wording for this as time passes.

Rouxles:
> Hi! Quick question about this PR...
> 
> I'm very in favor of relaxing regulations about clock pin position, so I'm very excited to see this being considered for the 2023 regulations! Though I'm a little confused about this particular proposal, it sounds like it's saying scramblers are allowed to send out pin misscrambles only if they confirm with the Delegate. That's not usually the issue with pins -- scramblers don't knowingly send out pin misscrambles, and if they see an issue they'll fix it rather than asking a Delegate. It's more likely that there are issues while running or lifting the cover.
> 
> As a (possibly controversial) alternative, would you consider just dropping Article F, or replacing F2 and F3 with a statement that "the pin position is not considered part of the scramble state, and the competitor is permitted to change the position of pins before the start of the solve"? That would mean that: -The scramble doesn't need to specify a pin position -Standard covers can be used, rather than clock-specific ones -During inspection, the competitor is permitted to set the pin position for their first turn -The competitor can put the puzzle standing, or lying down, but either way they won't be DNFed if the pins change
> 
> I think this would be more effective at reducing the challenges of organizing Clock. What do you think?
> 
> Thank you!

We voted internally on allowing competitors to move pins during inspection, and generally speaking on average we were against this change. I think the main issue is that the sound a pin makes can be pretty similar to the sound of a dial turning, so it's pretty possible to turn dials without judges noticing. We did consider this at some point though!

The wording of the proposal right now _will_ be changed - at the moment the intention with these PRs is to get some discussion going initially before the WRC sends out drafts to people

JoaoViniciusSantos:
@Rouxles Don't you think what I proposed on slack would be less abrupt? Like, wouldn't it be better to require pins to be correct during scrambling, but to allow results in which, during the transportation or the uncovering of the puzzles, the state of the pins change?

Rouxles:
> @Rouxles Don't you think what I proposed on slack would be less abrupt? Like, wouldn't it be better to require pins to be correct during scrambling, but to allow results in which, during the transportation or the uncovering of the puzzles, the state of the pins change?

Yes, I agree - I'm going to make a suggestion in this thread fairly soon, but will need to think about wording a bit more first

wca-regulations.md
```diff
@@ -173,6 +173,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 - 4f) Competition scramble sequences must be generated using a current official version of an official WCA scramble program (available [via the WCA website](https://www.worldcubeassociation.org/regulations/scrambles/)).
 - 4g) After scrambling a puzzle, the scrambler must verify that the puzzle is scrambled correctly. If the puzzle state is wrong, the scrambler must correct it (e.g. by solving the puzzle and applying the scramble sequence again).
     - 4g1) Exception: For the 6x6x6 Cube, 7x7x7 Cube, and Megaminx, it is not necessary to correct the puzzle state, at the discretion of the WCA Delegate.
+    - 4g2) Exception: For Clock, if the state of the pins are wrong, the scrambler should correct the pins, at the discretion of the WCA Delegate.
```
```suggestion
    - 4g2) Exception: For Clock, if the state of the pins are affected during transportation or uncovering, the scrambler should correct the pins.
```

Rouxles: How does this sound? @JoaoViniciusSantos

Rouxles: For me personally, I still don't feel too happy with this, but I think it's far clearer than previous versions

JoaoViniciusSantos: Perhaps using the same wording as 4g1? "Exception: For Clock, if the state of the pins is affected during transportation or uncovering, it is not necessary to correct the pins state, at the discretion of the WCA Delegate."

Nanush7: I agree with João.

```suggestion
    - 4g2) Exception: For Clock, if the state of the pins are affected during transportation or uncovering, it is not necessary to correct the pins, at the discretion of the WCA Delegate.
```

Rouxles: I'll add this suggestion to the commits for now

wca-regulations.md
```diff
@@ -173,6 +173,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 - 4f) Competition scramble sequences must be generated using a current official version of an official WCA scramble program (available [via the WCA website](https://www.worldcubeassociation.org/regulations/scrambles/)).
 - 4g) After scrambling a puzzle, the scrambler must verify that the puzzle is scrambled correctly. If the puzzle state is wrong, the scrambler must correct it (e.g. by solving the puzzle and applying the scramble sequence again).
     - 4g1) Exception: For the 6x6x6 Cube, 7x7x7 Cube, and Megaminx, it is not necessary to correct the puzzle state, at the discretion of the WCA Delegate.
+    - 4g2) Exception: For Clock, if the state of the pins are affected during transportation or uncovering, it is not necessary to correct the pins, at the discretion of the WCA Delegate.
```
Nevseros: How can pins be corrected after failed uncovering? What if pins were scrambled incorrectly by a scrambler, would this attempt be investigated by a Delegate? Maybe it's better to mention that if pins misscramble was found after the beginning of the attempt (e.g. during the inspection, after the solve, after the attempt), then the attempt should stay and should not be DNF'ed/extra should not be given?

JoaoViniciusSantos: After all, all the solutions we're getting to are sounding arbitrary and very open to abuse. I guess in the end the best solution would be to press all pins down or up :/

Randomno: I think making all pins up or down at the end of a scramble is much less favoured, so it would be unfortunate to settle for it because wording the alternative is difficult.

I think part of the problem is a lack of foundational regs regarding misscrambles. For example there is no reg that says the competitor must receive the correct scramble (only that the scrambler must scramble correctly).

Samuel-Baird:
> We voted internally on allowing competitors to move pins during inspection, and generally speaking on average we were against this change. I think the main issue is that the sound a pin makes can be pretty similar to the sound of a dial turning, so it's pretty possible to turn dials without judges noticing. We did consider this at some point though!

It seems like for this sort of scenario to happen the judge would need to be inattentive enough to not see the competitor making turns or grabbing the pins and twisting them. It seems plausible to me that competitors with a judge who has similar levels of inattentiveness could reasonably make moves in inspection/memorization periods of other events. Do you think that it would be significantly easier for a competitor to make moves during the inspection period of clock as opposed to making moves during the inspection/memorization period of other events? 

I have already tried out some incredible quiet clocks (in regards to both turning and pins) and if we already have judges with that level of inattentiveness then I believe it is already quite possible for competitors to make moves during inspection. Has the WRC discussed the problem of competitors making moves during inspection and looked into reasonable solutions for it?

Perhaps I am just too much of a fan of the idea of removing article F to simplify the regulations and help competition flow, but I feel like clock is getting a bit of discrimination for being easier to cheat than other events :P

Samuel-Baird:
> We voted internally on allowing competitors to move pins during inspection, and generally speaking on average we were against this change. I think the main issue is that the sound a pin makes can be pretty similar to the sound of a dial turning, so it's pretty possible to turn dials without judges noticing. We did consider this at some point though!

It seems like for this sort of scenario to happen the judge would need to be inattentive enough to not see the competitor making turns or grabbing the pins and twisting them. It seems plausible to me that competitors with a judge who has similar levels of inattentiveness could reasonably make moves in inspection/memorization periods of other events. Do you think that it would be significantly easier for a competitor to make moves during the inspection period of clock as opposed to making moves during the inspection/memorization period of other events?

I have already tried out some incredible quiet clocks (in regards to both turning and pins) and if we already have judges with that level of inattentiveness then I believe it is already quite possible for competitors to make moves during inspection. Has the WRC discussed the problem of competitors making moves during inspection and looked into reasonable solutions for it?

Perhaps I am just too much of a fan of the idea of removing article F to simplify the regulations and help competition flow, but I feel like clock is getting a bit of discrimination for being easier to cheat than other events :P

wca-regulations.md
Nevseros: I don't like delegate's discretion part here. As said many times by the delegates during the feedback period, it's unclear what they should do here. Maybe:

```suggestion
    - 4g2) Exception: For Clock, if the state of the pins are affected during transportation or uncovering, pins position should not be corrected.
```

wca-guidelines.md
```diff
@@ -94,6 +94,9 @@ To be more informative, each Guideline is classified using one of the following
 - 4d++) [ADDITION] It is permitted for the puzzle to change its orientation when it is moved from the scrambler to the solving station, as long as no one is attempting to influence the randomness of the orientation (see [Regulation A2e1](regulations:regulation:A2e1)).
 - 4f+) [RECOMMENDATION] The WCA Delegate should generate sufficient scramble sequences for the entire competition ahead of time, including spare scramble sequences for extra attempts.
 - 4f++) [REMINDER] If the WCA Delegate generates any additional scramble sequences during the competition, the scramble sequences must be saved.
+- 4g2+) [CLARIFICATION] The scrambler must still verify that the clock is scrambled correctly.
+- 4g2++) [ADDITION] If the clock falls over during uncovering, an extra must be given.
```
```suggestion
- 4g2++) [CLARIFICATION] If the clock falls over during uncovering, an extra must be given.
```

wca-regulations.md
```diff
@@ -173,6 +173,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 - 4f) Competition scramble sequences must be generated using a current official version of an official WCA scramble program (available [via the WCA website](https://www.worldcubeassociation.org/regulations/scrambles/)).
 - 4g) After scrambling a puzzle, the scrambler must verify that the puzzle is scrambled correctly. If the puzzle state is wrong, the scrambler must correct it (e.g. by solving the puzzle and applying the scramble sequence again).
     - 4g1) Exception: For the 6x6x6 Cube, 7x7x7 Cube, and Megaminx, it is not necessary to correct the puzzle state, at the discretion of the WCA Delegate.
+    - 4g2) Exception: For Clock, if the state of the pins are affected during transportation or uncovering, the pins should not be corrected.
```
```suggestion
    - 4g2) Exception: For Clock, if the state of the pins are affected during transportation or uncovering, the pins may be left uncorrected.
```

Rouxles: Comment to (hopefully) make this visible in the main thread

wca-guidelines.md
Rouxles: Comment to (hopefully) make this visible in the main thread

abunickabhi:
At a recent competition, the competitor demanded resolves for all 5 solves, as the pins got misscrambled when it was carried by the runner and/or put on the solving table by the judge.

When the competition is running at full pace, there can be inattentiveness from the scrambler, runner, judge and a pin misscramble can occur. Making a resolve for such situation just delays the round and adds to the confusion.

I strongly think this proposal should be considered in 2023 regulations.

CarterKoala (approved):

PR #1060, opened 2022-10-19
Add 11e+++ for extras

Rouxles:
Wording probably needs adjusting

PR #1074, opened 2022-10-30
1h+++

Rouxles:


wca-guidelines.md
```diff
@@ -38,6 +38,7 @@ To be more informative, each Guideline is classified using one of the following
 - 1h++) [RECOMMENDATION] All attempts in the final round of an event, as well as all 3x3x3 Fewest Moves attempts, should have the same scramble sequences for all competitors (i.e. only 1 group).
 - 1h1+) [CLARIFICATION] Scramblers/judges should only scramble for/judge other competitors in the same group if it is important for competition logistics.
 - 1h1++) [CLARIFICATION] For 3x3x3 Fewest Moves, judges may judge competitors in their own group before they have finished their attempt (i.e. while they are competing).
+- 1h1+++) [ADDITION] Competitors are permitted to compete earlier/later than the scheduled time, at the discretion of the WCA Delegate.
```
```suggestion
- 1h1+++) [ADDITION] Competitors may request to compete earlier or later than the scheduled time for the round in a separate group. Competitors requesting such accommodations should contact the organization team and WCA Delegate before the competition.
```

OJCubing:
I think we should make this much clearer that it's well within the delegates' right to not accommodate for this, at their own discretion. The way it's currently worded makes me think people will read it and assume they can always move into a later group as long as they tell the delegates beforehand (which I know isn't how it's actually worded, but I just can see it happening).

wca-guidelines.md
```diff
@@ -38,6 +38,7 @@ To be more informative, each Guideline is classified using one of the following
 - 1h++) [RECOMMENDATION] All attempts in the final round of an event, as well as all 3x3x3 Fewest Moves attempts, should have the same scramble sequences for all competitors (i.e. only 1 group).
 - 1h1+) [CLARIFICATION] Scramblers/judges should only scramble for/judge other competitors in the same group if it is important for competition logistics.
 - 1h1++) [CLARIFICATION] For 3x3x3 Fewest Moves, judges may judge competitors in their own group before they have finished their attempt (i.e. while they are competing).
+- 1h1+++) [ADDITION] Competitors may request to compete earlier or later than the scheduled time for the round in a separate group. Competitors requesting such accommodations should contact the organization team and WCA Delegate before the competition.
```
```suggestion
- 1h1+++) [ADDITION] Competitors may request to compete at a time different to the scheduled time for the round in a separate group, at the discretion of the WCA Delegate. Competitors requesting such accommodations should contact the organization team and WCA Delegate before the competition.
```

OJCubing: I'm not a fan of "earlier or later". I think something like "at a different time" would be better. Perhaps:
`Competitors may request to compete at a time different to the scheduled time...`

Nevseros:
I has just realised that it doesn't _clearly_ resolve the issue that essentially started this conversation - organizer failing to show up on time due to their organizer duties cannot contact the organizing team before the competition. But technically it's fine since we use `should`.

Jambrose777:
A couple of comments that may be helpful:
- Can probably change ", at the discretion of the WCA Delegate." to ". The WCA Delegate should carefully consider the fairness of the situation before approving any such requests.". This would show that the competitor may be told no in some cases, as that is a concern.
- Should probably include a reminder about 9l when allowing competitors to compete late.
- It may be good to provide examples of use cases: Not being able to compete on a certain calendar date due to religious reasons, having a Flight get delayed.
- It may be good to change how this affects the timing of a round (in relation to records). Ie. if a competitor is allowed to compete on a later day than when the original round happened, the round is still considered to have occurred on the original calendar date. Might be worth relooking at 9i2 if WRC wants to take that approach.

Rouxles:
> It may be good to change how this affects the timing of a round (in relation to records). Ie. if a competitor is allowed to compete on a later day than when the original round happened, the round is still considered to have occurred on the original calendar date. Might be worth relooking at 9i2 if WRC wants to take that approach.

I feel like this can also work with how extras are defined? (#1060)

I am for this change though

wca-guidelines.md
```diff
@@ -38,6 +38,7 @@ To be more informative, each Guideline is classified using one of the following
 - 1h++) [RECOMMENDATION] All attempts in the final round of an event, as well as all 3x3x3 Fewest Moves attempts, should have the same scramble sequences for all competitors (i.e. only 1 group).
 - 1h1+) [CLARIFICATION] Scramblers/judges should only scramble for/judge other competitors in the same group if it is important for competition logistics.
 - 1h1++) [CLARIFICATION] For 3x3x3 Fewest Moves, judges may judge competitors in their own group before they have finished their attempt (i.e. while they are competing).
+- 1h1+++) [ADDITION] Competitors may request to compete at a time different to the scheduled time for the round in a separate group, at the discretion of the WCA Delegate. Competitors requesting such accommodations should contact the organization team and WCA Delegate before the competition.
```
```suggestion
- 1h1+++) [ADDITION] Competitors may request to compete at a time different to the scheduled time for the round in a separate group. The WCA Delegate should carefully consider the fairness of the situation before approving any such requests. Competitors requesting such accommodations should contact the organization team and WCA Delegate before the competition.
```

wca-guidelines.md
```diff
@@ -38,6 +38,7 @@ To be more informative, each Guideline is classified using one of the following
 - 1h++) [RECOMMENDATION] All attempts in the final round of an event, as well as all 3x3x3 Fewest Moves attempts, should have the same scramble sequences for all competitors (i.e. only 1 group).
 - 1h1+) [CLARIFICATION] Scramblers/judges should only scramble for/judge other competitors in the same group if it is important for competition logistics.
 - 1h1++) [CLARIFICATION] For 3x3x3 Fewest Moves, judges may judge competitors in their own group before they have finished their attempt (i.e. while they are competing).
+- 1h1+++) [ADDITION] Competitors may request to compete at a time different to the scheduled time for the round in a separate group. The WCA Delegate should carefully consider the fairness of the situation before approving any such requests. Competitors requesting such accommodations should contact the organization team and WCA Delegate before the competition.
```
lgarron: I'd add some explicit caveats that this can't break 9l of E2b++.

In particular, I think we should definitely exclude all FMC attempts from this. Theoretically, someone could start slightly early or slightly late without violating E2b++, but 1) "the slightly early" case would violate E2b++ if they decide to finish their attempt before everyone else has started, and 2) the "slightly late" case is already covered by E2b++.

```suggestion
- 1h1+++) [ADDITION] Competitors may request to compete at a time different to the scheduled time for the round in a separate group. The WCA Delegate should carefully consider the fairness of the situation before approving any such requests, and must ensure that such attempts follow [Regulation 9l](regulations:regulation:9l) (i.e. they must start after all competitors have finished their attempts from any previous round of the same event, and finish before any competitor has started an attempt for any following round of the same event). Competitors requesting such accommodations should contact the organization team and WCA Delegate before the competition. Exception: such accommodations are not permitted for 3x3x3 Fewest Moves, which specifies its own limited accommodations in [Guideline E2b++](guidelines:guideline:E2b++).
```

wca-guidelines.md
```diff
@@ -38,7 +38,7 @@ To be more informative, each Guideline is classified using one of the following
...
-- 1h1+++) [ADDITION] Competitors may request to compete at a time different to the scheduled time for the round in a separate group. The WCA Delegate should carefully consider the fairness of the situation before approving any such requests. Competitors requesting such accommodations should contact the organization team and WCA Delegate before the competition.
+- 1h1+++) [ADDITION] Competitors may request to compete at a time different to the scheduled time for the round in a separate group. The WCA Delegate should carefully consider the fairness of the situation before approving any such requests, and must ensure that such attempts follow [Regulation 9l](regulations:regulation:9l) (i.e. they must start after all competitors have finished their attempts from any previous round of the same event, and finish before any competitor has started an attempt for any following round of the same event). Competitors requesting such accommodations should contact the organization team and WCA Delegate before the competition. Exception: such accommodations are not permitted for 3x3x3 Fewest Moves, which specifies its own limited accommodations in [Guideline E2b++](guidelines:guideline:E2b++).
```
Nevseros, line 41: Any reason there's `must` in parentheses?

Edit: I mean, it should definitely be `must` for finishing, but not for starting maybe?

Rouxles: I think in this case it's just a clarifying example and still works here

CarterKoala (approved):

blake98x (commented):
re: 1h+++ -- I think this should be re-phrased such that it doesn't give the assumption that these circumstances **will** be accommodated. There are definitely scenarios that are overlooked here such as a competitor wanting to complete an entire day's worth of events on some other day of the competition. I'd like to think that Delegates would be willing to entertain requests like this, but I expect there would be cases where it's outright infeasible.

Edit: I see this question was addressed above, but given that I brought it up after the change, perhaps it's worth considering. In particular, maybe we change ". The WCA Delegate should carefully consider the fairness of the situation before approving any such requests." to ". The WCA Delegate should carefully consider the fairness _and practicality_ of the situation before approving any such requests."

Nevseros:
> I'd like to think that Delegates would be willing to entertain requests like this, but I expect there would be cases where it's outright infeasible.

Should we actually try to catch every scenario like the one you mentioned? I think if, as in your example, the request is infeasible to accomodate, then the delegate would not accomodate it, and `practicality` would be counted by itself. In the end of the day, I do think this guideline is complete as it does not require the delegate to meet the competitor's request, but gives them an opportunity to do so at their discretion. 

Not sceptical about further changes, but I don't see the need for them.

blake98x:
My main concern is that competitors would be able to point to a regulation like this and proclaim that a Delegate doesn't necessarily think a request is unfair, but rather that they merely don't want to pursue it. Perhaps a helpful and more specific example would be a competitor completing their attempts for the main day of a multi-day competition on a day scheduled for PBQ events. It's _technically_ possible to accommodate, and there's no risk of knocking competitors out of rounds to which they were under the impression that they had advanced - but it's merely inconvenient on the end of the organizers/Delegates.

In this kind of scenario, I wouldn't think that practicality is covered by fairness, but I do agree that they overlap in most cases.

Rouxles:
I think I agree with Blake here - are there any other changes you want for this regulation? if not, I'll edit the current diff we have

blake98x:
Good from my end, at least! Nobody from the Board has spoken up internally against my suggestion. :P

Rouxles:
#1096

PR #1077, opened 2022-11-09
Fix timer related definitions use

Nevseros:
Partially resolves the problem with timer related definitions use, it's going to be a separate PR for Article B due to more complex changes strongly connected to this small fix.


JoaoViniciusSantos (approved):
LGTM!

PR #1078, opened 2022-11-09
Clarify timer starts and stops for Blindfolded Solving

Nevseros:
* Explicitly mentioned that solve start == attempt start for BLD;
* B2b, B2c, B2d updated so they only apply when the timer is the Stackmat timer;
* B2e $\rightarrow$ B2d2;
* Timer start when the stopwatch is the only timer made more lenient (B2f);
* A6a1 and A6a2 moved to Article b to replace B5b;
* Minor clarifications added to make the changes clear.

### DIFF
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -180,6 +180,7 @@ To be more informative, each Guideline is classified using one of the following 
 - B1+) [REMINDER] The competitor must use a puzzle without textures, markings, or other features that distinguish similar pieces (see [Regulation 3k](regulations:regulation:3k)). This should be given special attention for Blindfolded Solving.
 - B1b+) [RECOMMENDATION] Blindfolds should be checked by the WCA Delegate before use in the competition.
 - B2d+) [ADDITION] By default, the competitor starts the solve the first time that they start the timer after the judge has indicated that they are ready (see [Regulation B2a](regulations:regulation:B2a)). If they want to check that the timer is in working order, they must (ask and) receive confirmation from the judge each time before starting/resetting the timer during this phase. Penalty for starting and resetting a timer without confirmation from the judge: disqualification of the attempt (DNF).
+- B2f+) [CLARIFICATION] Since the competitor starts the solve by lifting the cover, penalties for starting the timer and for touching the cover while starting the attempt are omitted (see [Regulation B2b](regulations:regulation:B2b) and [Regulation B2c](regulations:regulation:B2c)).
 
 
 ## <article-C><one-handed><onehandedsolving> Article C: One-Handed Solving
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -180,7 +180,7 @@ To be more informative, each Guideline is classified using one of the following 
 - B1+) [REMINDER] The competitor must use a puzzle without textures, markings, or other features that distinguish similar pieces (see [Regulation 3k](regulations:regulation:3k)). This should be given special attention for Blindfolded Solving.
 - B1b+) [RECOMMENDATION] Blindfolds should be checked by the WCA Delegate before use in the competition.
 - B2d+) [ADDITION] By default, the competitor starts the solve the first time that they start the timer after the judge has indicated that they are ready (see [Regulation B2a](regulations:regulation:B2a)). If they want to check that the timer is in working order, they must (ask and) receive confirmation from the judge each time before starting/resetting the timer during this phase. Penalty for starting and resetting a timer without confirmation from the judge: disqualification of the attempt (DNF).
-- B2f+) [CLARIFICATION] Since the competitor starts the solve by lifting the cover, penalties for starting the timer and for touching the cover while starting the attempt are omitted (see [Regulation B2b](regulations:regulation:B2b) and [Regulation B2c](regulations:regulation:B2c)).
+- B2f+) [CLARIFICATION] Since the competitor starts the solve by lifting the cover, penalties for starting the timer and for touching the cover while starting the attempt do not apply (see [Regulation B2b](regulations:regulation:B2b) and [Regulation B2c](regulations:regulation:B2c)).
 
 
 ## <article-C><one-handed><onehandedsolving> Article C: One-Handed Solving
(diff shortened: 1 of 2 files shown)


wca-regulations.md
```diff
@@ -443,16 +441,16 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
...
+    - B2d) When using the Stackmat timer, the competitor starts the attempt by removing their hands from the timer, thus starting the timer. (This also starts the solve.)
         - B2d1) The competitor removes the cover from the puzzle after starting the timer.
-    - B2e) If a stopwatch is in use in addition to a Stackmat timer, the judge starts the stopwatch as soon as the competitor starts the solve.
-    - B2f) If a stopwatch is in use as the only timer, the competitor places their hands on the table (instead of a Stackmat timer). After confirming that they are ready, the competitor starts the solve by lifting the cover. The judge starts the stopwatch as soon as the competitor starts the solve.
+        - B2d2) If a stopwatch is in use in addition to the Stackmat timer, the judge starts the stopwatch as soon as the competitor starts the solve.
+    - B2f) When using a stopwatch as the only timer, the competitor confirms they are ready and starts the solve by lifting the cover. The judge starts the stopwatch as soon as the competitor starts the solve.
```
JoaoViniciusSantos, line 453: Why not use the procedure of starting the solve with hands down, and then starting the solve after the hands are lifted from the table? I believe that would be fairer, as it would be analogue to starting a stackmat timer

Nevseros: I don't see the reason to do so in favor of fairness, as stopwatch anyway makes the time much more inaccurate. But lenient regulations would make the procedure more clear and easy. We may discuss it tomorrow anyway.

wca-guidelines.md
```diff
@@ -180,6 +180,7 @@ To be more informative, each Guideline is classified using one of the following
 - B1+) [REMINDER] The competitor must use a puzzle without textures, markings, or other features that distinguish similar pieces (see [Regulation 3k](regulations:regulation:3k)). This should be given special attention for Blindfolded Solving.
 - B1b+) [RECOMMENDATION] Blindfolds should be checked by the WCA Delegate before use in the competition.
 - B2d+) [ADDITION] By default, the competitor starts the solve the first time that they start the timer after the judge has indicated that they are ready (see [Regulation B2a](regulations:regulation:B2a)). If they want to check that the timer is in working order, they must (ask and) receive confirmation from the judge each time before starting/resetting the timer during this phase. Penalty for starting and resetting a timer without confirmation from the judge: disqualification of the attempt (DNF).
+- B2f+) [CLARIFICATION] Since the competitor starts the solve by lifting the cover, penalties for starting the timer and for touching the cover while starting the attempt are omitted (see [Regulation B2b](regulations:regulation:B2b) and [Regulation B2c](regulations:regulation:B2c)).
```
Samuel-Baird: maybe "omitted" could be changed to "do not apply"? I feel like that should be more understandable to our broad competitor base :)

Jambrose777:
~~**Major issue**:
A6a1) and A6a2) should not be moved to blindfolded only regulations. It is possible to have an attempt be recorded only using a stopwatch (ie a 7x7 solve of 20 minutes).~~

Nevseros:
@Jambrose777 
> Major issue:
A6a1) and A6a2) should not be moved to blindfolded only regulations. It is possible to have an attempt be recorded only using a stopwatch (ie a 7x7 solve of 20 minutes).

It's not. See [A1b1](https://www.worldcubeassociation.org/regulations/#A1b1).

Jambrose777:
> @Jambrose777
> 
> > Major issue:
> > A6a1) and A6a2) should not be moved to blindfolded only regulations. It is possible to have an attempt be recorded only using a stopwatch (ie a 7x7 solve of 20 minutes).
> 
> It's not. See [A1b1](https://www.worldcubeassociation.org/regulations/#A1b1).

Ooh, thanks! Indeed I missed that.

wca-guidelines.md
Rouxles:
```suggestion
- B2f+) [CLARIFICATION] Since the competitor starts the solve by lifting the cover, penalties for starting the timer and for touching the cover while starting the attempt do not apply (see [Regulation B2b](regulations:regulation:B2b) and [Regulation B2c](regulations:regulation:B2c)).
```

Nevseros: We already use `omitted` in the redulations BTW (for example, in [12a2](https://www.worldcubeassociation.org/regulations/#12a2) and [A7b1](https://www.worldcubeassociation.org/regulations/#A7b1)), but maybe with a slightly different meaning.

Samuel-Baird: In both of those examples "omitted" is used in a sense of leaving something out or excluding it. When I first looked at this I was just aiming for it to be more understandable to a general audience but now that I'm looking at this again I'm not even sure if this would be a correct use of omitted. 

Swapping omitted out for its synonyms doesn't sound natural in English (the penalties are left out, the penalties are excluded) the synonyms give me a weird sense that the penalties should be, but are not in practice, applied. As a result, I believe that "do not apply" would be a better fit than omitted :)

Samuel-Baird: Both are currently used in the sense of something not being applied or being excluded which fits their context pretty well. After reading both of those regulations, `omitted` actually seems a bit odd to use in this regulation. 

Originally I was thinking that `omitted` may just be harder to understand for competitors with a more limited English vocabulary, but after testing it out by substituting synonyms I think it may not even be a normal (maybe even correct?) usage of the word. 

`penalties... are left out` and `penalties... are excluded` both sound a bit odd in this context, giving me a feeling that the penalties should apply, but are not applied in practice. I now believe that `do not apply` would be a better fit as it sounds a bit more confident and natural (and maybe has a more correct meaning?)

Rouxles (approved):

wca-regulations.md
```diff
@@ -443,16 +441,16 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
...
-    - B2b) The competitor uses their fingers to touch the elevated sensor surfaces of the timer. The competitor's palms must be facing down, and located on the side of the timer that is closer to them. Penalty: time penalty (+2 seconds).
-    - B2c) The competitor must have no physical contact with the puzzle or the puzzle cover while starting the attempt. Penalty: time penalty (+2 seconds).
-    - B2d) The competitor starts the attempt by removing their hands from the timer, thus starting the timer. (This also starts the solve.)
+    - B2b) When using the Stackmat timer, the competitor uses their fingers to touch the elevated sensor surfaces of the timer. The competitor's palms must be facing down, and located on the side of the timer that is closer to them. Penalty: time penalty (+2 seconds).
+    - B2c) When using the Stackmat timer, the competitor must have no physical contact with the puzzle or the puzzle cover while starting the attempt. Penalty: time penalty (+2 seconds).
```
Nevseros, line 449: @thewca/wrc-team 

While translating this one, I realised that we (for some reason) have a mention of a puzzle here. It should either be removed (since the puzzle is anyway covered) or added for the case of stopwatch-only use, or the puzzle should be mentioned in B2f+.

PR #1079, opened 2022-11-10
Addition for MBLD scoring

Rouxles:


JoaoViniciusSantos (approved):
LGTM!

Nevseros (commented):
Maybe it would be better to make it 9f2+? We usually use `record` instead of `write down`, WDYT on this one?

Samuel-Baird:
I generally find that having competitors write down the full time shown on the stopwatch helps competitors avoid truncating instead of rounding as [9f2](https://www.worldcubeassociation.org/regulations/#9f2) requires. Scoretakers tend to have a little more experience with the Regulations and if there are any disputes with the decimals then the scorecard having all of the decimals is helpful.

I believe that adding decimal places to this result helps avoid the idea of that judges should write the rounded time as opposed to the shown time.

wca-guidelines.md
```diff
@@ -118,6 +118,7 @@ To be more informative, each Guideline is classified using one of the following
 - 9b++) [ADDITION] Events other than those specified in [Regulation 9b](regulations:regulation:9b) may be held during a competition, but will be considered unofficial and therefore will not be included in the official results of the competition.
 - 9b3b+) [CLARIFICATION] The "Mean of 3" statistic does not affect the competitor's ranking in a "Best of 3" round (which is based on the best result). The WCA recognizes it outside of the competition format.
 - 9f1+) [EXAMPLE] If the timer displays 12.678 for an attempt, the original recorded time is 12.67 (drop any digit after a hundredth of a second). A judge may write the full-precision time on the score sheet (and keep the precision in the calculation of penalties), as long as the score taker truncates it for the official results.
+- 9f1++) [ADDITION] For 3x3x3 Multi-Blind, the judge should write down the number of puzzles solved, the number of puzzles attempted, and the final time. For example, if the competitor solves 3 cubes out of 5 attempted cubes in 40 minutes, the judge should write down 3/5 40:00
```
```suggestion
- 9f2+) [ADDITION] For 3x3x3 Multi-Blind, the judge should record the number of puzzles solved, the number of puzzles attempted, and the final time. For example, if the competitor solves 3 cubes out of 5 attempted cubes in 40 minutes, 30 seconds, and 25 centiseconds, the judge should write down 3/5 40:30.25
```

wca-guidelines.md
```diff
@@ -118,6 +118,7 @@ To be more informative, each Guideline is classified using one of the following
 - 9b++) [ADDITION] Events other than those specified in [Regulation 9b](regulations:regulation:9b) may be held during a competition, but will be considered unofficial and therefore will not be included in the official results of the competition.
 - 9b3b+) [CLARIFICATION] The "Mean of 3" statistic does not affect the competitor's ranking in a "Best of 3" round (which is based on the best result). The WCA recognizes it outside of the competition format.
 - 9f1+) [EXAMPLE] If the timer displays 12.678 for an attempt, the original recorded time is 12.67 (drop any digit after a hundredth of a second). A judge may write the full-precision time on the score sheet (and keep the precision in the calculation of penalties), as long as the score taker truncates it for the official results.
+- 9f2+) [ADDITION] For 3x3x3 Multi-Blind, the judge should record the number of puzzles solved, the number of puzzles attempted, and the final time. For example, if the competitor solves 3 cubes out of 5 attempted cubes in 40 minutes, 30 seconds, and 25 centiseconds, the judge should write down 3/5 40:30.25
```
lgarron: Nit: needs a final period.

I'm not enthused about the way the time is spelled out, but I think reasonable in this case.

Also, I *believe* it's common to write the time in parentheses, i.e. `3/5 (40:30.25)` — would that be worth specifying in Article H, even?
```suggestion
- 9f2+) [ADDITION] For 3x3x3 Multi-Blind, the judge should record the number of puzzles solved, the number of puzzles attempted, and the final time. For example, if the competitor solves 3 cubes out of 5 attempted cubes in 40 minutes, 30 seconds, and 25 centiseconds, the judge should write down 3/5 40:30.25
```
```suggestion
- 9f2+) [ADDITION] For 3x3x3 Multi-Blind, the judge should record the number of puzzles solved, the number of puzzles attempted, and the final time. For example, if the competitor solves 3 cubes out of 5 attempted cubes in 40 minutes, 30 seconds, and 25 centiseconds, the judge should write down 3/5 40:30.25.
```

```suggestion
- 9f2+) [ADDITION] For 3x3x3 Multi-Blind, the judge should record the number of puzzles solved, the number of puzzles attempted, and the final time. For example, if the competitor solves 3 cubes out of 5 attempted cubes in 40 minutes and 30 seconds, the judge should write down 3/5 40:30.
```

I'm aware the public consultation period is over but wish to object to the inclusion of centiseconds in this guideline.

Centiseconds do not form part of the result. Many stopwatches do not display centiseconds beyond a certain timeframe, so they will not be available in all cases. The accuracy of judges stopping stopwatches is not high enough that they are a meaningful figure.

Should the justification for this addition be that the centiseconds allow the rounding requirement in 9f2 to be applied, I will point out I believe that to be a flawed regulation in the first place. I have seen multiple instances of experienced judges failing to round up times. Stopwatches that do not display centiseconds effectively truncate by default. Truncation of all times would be a more elegant solution and the addition of centiseconds to this guideline to support the existing 9f2 is a poor way to address the current problem.

Should the centiseconds element of guideline be added I predict that, like the existing rounding requirements of 9f2, it will commonly be misapplied in practice. This is not a desirable outcome.

PR #1076, opened 2022-11-03
Altering Clock Regulations

JoaoViniciusSantos:
Altering the clock regulations I proposed on the Drive document I sent you guys in october, mainly so we can discuss most of these changes. **Nothing is definitive here!!!!**

PR #1080, opened 2022-11-11
Alternative Clock scrambling solution (All pins up)

JoaoViniciusSantos:
Basically an alternative to #1076.

I don't know if the wording is the best neither if 4d is the right place for this, but that was the best I could think. Let me know your opinions :)

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -169,6 +169,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 4d2) Pyraminx is scrambled starting with the yellow face (if not possible, then the lightest face) on the bottom and the green face (if not possible, then the darkest adjacent face) on the front.
     - 4d3) Square-1 is scrambled starting with the darker color on front (out of the 2 possible scrambling orientations).
     - 4d4) Clock is scrambled starting with either side in front and 12 o'clock pointing up.
+        - 4d4a) After scrambling the puzzle, the scrambler must orient all pins up, from the position the puzzle is held.
     - 4d5) Skewb is scrambled starting with the white face (if not possible, then the lightest face) on top and the green face (if not possible, then the darkest adjacent face) on the front-left.
 - 4f) Competition scramble sequences must be generated using a current official version of an official WCA scramble program (available [via the WCA website](https://www.worldcubeassociation.org/regulations/scrambles/)).
 - 4g) After scrambling a puzzle, the scrambler must verify that the puzzle is scrambled correctly. If the puzzle state is wrong, the scrambler must correct it (e.g. by solving the puzzle and applying the scramble sequence again).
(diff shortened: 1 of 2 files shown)


andrewtyberg (approved):

Samuel-Baird:
Just requiring scrambles to make clocks all pins up doesn't seem to solve the problem of scramblers mistaking which way is "up" in the scramble picture which may result in scramblers giving "all pin down" scrambles.

I think that a better solution would be to ignore pins for scrambles and instead require that puzzles must be delivered to the station laying down (as opposed to standing up). 

Pros:
-Resolves incidents of clocks having pin state altered during running or at the beginning of inspection
-Not open to abuse discussed in #1070  with competitors making moves during inspection
-Allows the use of standard competition equipment instead of specified covers or stands for clock
-It is already a common incident for inexperienced competitors to place clocks face down instead of standing up

Cons:
-there are 2 possible pin configurations based on what face the clock is placed down on all "up" or "down" this could be resolved by allowing competitors to place their clocks face down either way--as there is no concern of making moves in inspection. This also has the benefit that attempts involving clocks falling over after the judge has lifted the cover (such as the table getting bumped after a competitor has set a clock down standing up or the competitor accidentally dropping the puzzle) will not need to result in DNFs/extras but instead can stand. 

This would remove several regulations from article F but also re-add several regulations, resulting in an article of similar length but is much less incident prone than our current regulations. I would be very much in favor of something like this being implemented as opposed to #1070  in its current form :)


Here's an idea of potential working for article 6. I'm not sure that I like how I wrote it out, but as you can see it's 5 lines as opposed to the current 6.

F1) Standard speed solving procedures are followed, as described in Article A (Speed Solving). Additional regulations that supersede the corresponding procedures in Article A are described below.
F2) The judge places the scrambled puzzle onto the mat in a lying position.
F3) At the end of the inspection period, the competitor may place the puzzle in a standing or laying position
F3a) The competitor must not change the individual positions of any pins from their scrambled positions before the start of the solve. Penalty: disqualification of the attempt (DNF).
F3a1) Changing the position of pins as a result of the puzzle being placed in a lying position is not considered changing individual pin positions.

wca-regulations.md
```diff
@@ -169,6 +169,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 4d2) Pyraminx is scrambled starting with the yellow face (if not possible, then the lightest face) on the bottom and the green face (if not possible, then the darkest adjacent face) on the front.
     - 4d3) Square-1 is scrambled starting with the darker color on front (out of the 2 possible scrambling orientations).
     - 4d4) Clock is scrambled starting with either side in front and 12 o'clock pointing up.
+        - 4d4a) After scrambling the puzzle, the scrambler must orient all pins up, from the position the puzzle is held.
```
lgarron, line 172: 12 o'clock is already up, and the notation is consistent with the clock faces being on F and B. I'd write it like this:

```suggestion
        - 4d4a) After scrambling the puzzle, the scrambler must push all pins to the front (sometimes informally described as "pushed up"), from the position the puzzle is held.
```

Rouxles:
Not being implemented for this cycle at the moment.

abunickabhi:
I agree with Samuel that it is better to ignore the states of the pin rather than make sure all pins up in future scrambles if any rule modification for clock scrambles get implemented.

PR #1081, opened 2022-11-11
Reword A7a

andrewtyberg:
This is probably the least followed regulation. It's unnecessary for the judge to say "okay" after every non-penalized attempt.

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -425,9 +425,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - A6i) Time penalties for stopping the solve are cumulative.
 - A7) Recording results:
     - A7a) The judge tells the competitor the result.
-        - A7a1) If the judge finds that the puzzle is solved, they call "OKAY".
-        - A7a2) If the judge assigns any penalties, they call "PENALTY".
-        - A7a3) If the result is DNF, the judge calls "DNF".
+        - A7a1) If the judge assigns any penalties or a DNF, they should tell the competitor.
     - A7b) The judge records the result on a score sheet and signs next to the result to acknowledge that the result is correct, complete, properly formatted, and clearly readable.
         - A7b1) If penalties are assigned, the judge records the original recorded time displayed on the timer, along with any penalties. The format should be "X + T + Y = F", where X represents the sum of time penalties before/starting the solve, T represents the time displayed on the timer (the "original recorded time"), Y represents a sum of time penalties during/after the solve, and F represents the final result (e.g. 2 + 17.65 + 2 = 21.65). If X and/or Y is 0, the 0 terms are omitted (e.g. 17.65 + 2 = 19.65).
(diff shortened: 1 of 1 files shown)


andrewtyberg:
I think it would also be reasonable to completely remove all of A7a.

Nevseros:
I'm strongly opposed to rewording it as proposed here. "Okay" is the easiest way for the competitor to understand that they can touch the puzzle, that the judge inspected it. Instead of removing or rewording this significantly, IMO it would be better to make it less mandatory (i.e. add `should` with minor rewording).

andrewtyberg:
> I'm strongly opposed to rewording it as proposed here. "Okay" is the easiest way for the competitor to understand that they can touch the puzzle, that the judge inspected it. Instead of removing or rewording this significantly, IMO it would be better to make it less mandatory (i.e. add `should` with minor rewording).

Yeah, but the issue is that this has been in the regulations for a long time, and still almost no one actually follows this. At least in my region, without exaggerating, probably only 1-2% of judges actually do this. It's a good regulation in theory, but time has shown that this is not a rule that is followed. I'd be interested to hear if judges in Russia (or other places) follow these regulations.

Nevseros:
> I'd be interested to hear if judges in Russia (or other places) follow these regulations.

In Russia, we mention it during tutorial every time and strongly encourage judges to call "okay". Some judges really call it. I stand by my opinion that this regulation just should be reworded to make it less strict, not made this short. 

Proposed wording also doesn't indicate if there actually is a moment when judge inspects the puzzle. If the intention of A7a's is going to be saved, I'd say that we also should link A6e to it.

OJCubing:
I'm mostly with Andrew on this one, but perhaps it would be good to still cover the OKAY case. Maybe something like
`The judge should indicate to the competitor that the puzzle is solved, or that a penalty will be applied.`

wca-regulations.md
```diff
@@ -425,9 +425,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - A6i) Time penalties for stopping the solve are cumulative.
 - A7) Recording results:
     - A7a) The judge tells the competitor the result.
-        - A7a1) If the judge finds that the puzzle is solved, they call "OKAY".
-        - A7a2) If the judge assigns any penalties, they call "PENALTY".
-        - A7a3) If the result is DNF, the judge calls "DNF".
+        - A7a1) If the judge assigns any penalties or a DNF, they should tell the competitor.
```
```suggestion
        - A7a1) The judge should indicate to the competitor that the puzzle is solved, or that a penalty will be applied.
```

Jambrose777: Im in strong favor of this change opposed to removing the judge indicating the puzzle is solved entirely. The reasoning is https://www.worldcubeassociation.org/regulations/#A6e

Randomno:
In practice, I assume the judge has inspected the puzzle when they begin recording the result, as that is the next regulation in the sequence. If that is indeed what should indicate it, I think it can be explicitly stated.

Of course, judges don't always do their job properly and may not have inspected the puzzle fully before recording the result, so it may still not be suitable to touch the puzzle at that point.

Rouxles (approved):

PR #1083, opened 2022-11-13
Attempt should be stopped immediately after DNF Regulation violation

Nevseros:


wca-regulations.md
```diff
@@ -236,6 +236,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 9f1) All timed results under 10 minutes, except for 3x3x3 Multi-Blind, are measured and truncated to the nearest hundredth of a second. All timed averages and means under 10 minutes are measured and rounded to the nearest hundredth of a second.
     - 9f2) All timed results, averages, and means over 10 minutes, as well as all times for 3x3x3 Multi-Blind results, are measured and rounded to the nearest second (e.g. x.49 becomes x, x.50 becomes x+1).
     - 9f4) The result of an attempt is recorded as DNF (Did Not Finish) if the attempt is disqualified or unsolved/unfinished.
+        - 9f4a) If during an attempt a competitor violates the regulation that will result in disqualification of this attempt (DNF), the attempt should be stopped immediately after this violation.
```
```suggestion
        - 9f4a) If a competitor violates a regulation that would result in disqualification of their attempt, the attempt should be stopped immediately after this violation.
```

Nevseros: If the violation happened before the attempt, the attempt cannot be stopped, that's why I originally included `during an attempt`.

sam596:
I think this should be made clear that the attempt should be stopped only if it's **clear and obvious** that the attempt should be DNFed without Delegate intervention.

Totally random example: A competitor is solving, and a spectator sneezes, the competitor instinctively says "Bless you" out loud, the spectator responds "thank you". This is a violation of [A5a](https://www.worldcubeassociation.org/regulations/#A5a), but as a Delegate I would not DNF this solve, and I would not want a judge to stop their attempt because they believe it is now inevitably a DNF. Allow the solve to finish before making a decision like this in case of any erroneous or "too strict" interpretation of the regulations.

Nevseros:
> I think this should be made clear that the attempt should be stopped only if it's clear and obvious that the attempt should be DNFed without Delegate intervention.

I agree. @sam596, do you have a wording in your mind for that? Using something like "clear and obvious" in the regulations does not seems like a nice option.

wca-regulations.md
```diff
@@ -236,6 +236,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 9f1) All timed results under 10 minutes, except for 3x3x3 Multi-Blind, are measured and truncated to the nearest hundredth of a second. All timed averages and means under 10 minutes are measured and rounded to the nearest hundredth of a second.
     - 9f2) All timed results, averages, and means over 10 minutes, as well as all times for 3x3x3 Multi-Blind results, are measured and rounded to the nearest second (e.g. x.49 becomes x, x.50 becomes x+1).
     - 9f4) The result of an attempt is recorded as DNF (Did Not Finish) if the attempt is disqualified or unsolved/unfinished.
+        - 9f4a) If a competitor violates a regulation that would result in disqualification of their attempt, the attempt should be stopped immediately after this violation.
```
```suggestion
        - 9f4a) If a competitor clearly violates a regulation that would result in disqualification of their attempt, the attempt should be stopped immediately after this violation, at the discretion of the judge or WCA Delegate.
```

Rouxles: @Nevseros how does this sound? can remove clearly here

Nevseros: LGTM!

lgarron: I'd edit a little further and provide more specific guidance, to reduce ambiguity.
```suggestion
        - 9f4a) If a competitor violates a regulation clearly enough that the attempt is certain to be disqualified, the judge should immediately stop the attempt. If the judge is uncertain or a dispute could be disruptive (e.g. because an extra attempt could delay the competition), the judge should consult the WCA Delegate.
```

Rouxles: I think that works far better

Rouxles (approved):

PR #1085, opened 2022-11-15
Pushing clock pins in inspection is +2

andrewtyberg:
Assigning a DNF for this is overly harsh. Pushing a clock pin (or multiple pins) does not allow the competitor to see further into the solve the same way that turning a layer in other puzzles does, or that turning a clock dial does. Additionally, clock pins are easy to accidentally push compared to turning clock dials or turning layers on other puzzles. It is hard to see an instance where this could be abused. That is, there will never be a situation when it is “worth it” to move a pin during inspection and take the +2. Even someone who does not know how to solve a clock can easily push all 4 pins in under 2 seconds.

The only oddity with this proposal is that, if a competitor does accidentally push a pin during inspection, they would then be inclined to move all pins to their preferred locations. However, a solution to this would be to assign a +2 for EACH pin the competitor pushes during inspection. For example, if a competitor pushes three pins in inspection, they would be assigned three +2 penalties. (Although if the SAME pin is pushed multiple times, this does not incur multiple penalties.)

The last part does not have to be implemented and it is also a possibility to simply assign one +2 for any number of pins pushed during inspection.

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -516,7 +516,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - F2a) The organization team may enforce using a stand for the puzzle to prevent it from falling before the start of the attempt. If such stands are used, the organization team should make an announcement before the round starts.
         - F2a1) The judge must remove the stand from the mat immediately after the competitor removes the puzzle from it.
 - F3) At the end of the inspection period, the competitor places the puzzle in a standing position. Penalty: disqualification of the attempt (DNF).
-    - F3a) The competitor must not change the positions of any pins from their scrambled positions before the start of the solve. Penalty: disqualification of the attempt (DNF).
+    - F3a) The competitor must not change the positions of any pins from their scrambled positions before the start of the solve. Penalty: time penalty (+2 seconds).
 
 
 ## <article-H><multiple-blindfolded><multipleblindfoldedsolving> Article H: Multi-Blind Solving
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -208,6 +208,12 @@ To be more informative, each Guideline is classified using one of the following 
 - E3d++) [CLARIFICATION] Competitors should not consider a personal stopwatch or watch as the official time, and must submit their solution when the judge calls "STOP".
 
 
+## <article-F><clock><clocksolving> Article F: Clock Solving
+
+- F3a+) [CLARIFICATION] One time penalty is assigned for each pin moved before the start of the solve. For example, if the competitor moves 3 pins before the start of the solve, 3 time penalties are assigned.
+- F3a++) [CLARIFICATION] If the same pin is moved multiple times, only one time penalty is assigned for that pin.
+
+
 ## <article-H><multiple-blindfolded><multipleblindfoldedsolving> Article H: Multi-Blind Solving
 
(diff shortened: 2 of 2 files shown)


Samuel-Baird:
The WRC has not supported manipulating clock pins during inspection because of the risk it poses to cheating (making moves in inspection). This has the exact same problem as allowing pins to be manipulated during inspection, requires the same careful wording to disallow pins being twisted to make moves, and would not solve nearly as many problems as allowing pins during inspection would.

Either this push should be rejected on the same basis that the WRC has rejected pin manipulation during inspection or it should be disregarded an pin manipulation during inspection should be added.

andrewtyberg:
I've updated the PR to make time penalties not cumulative, as seems to be the general consensus, and I've clarified that turning a pin will result in a DNF.

Another option would be to remove F3a++ and instead have something like the following:

F3a) The competitor must not push pins during inspection. Penalty: time penalty (+2 seconds).
F3b) The competitor must not turn pins during inspection. Penalty: disqualification of the attempt (DNF).

Samuel-Baird:
> I've updated the PR to make time penalties not cumulative, as seems to be the general consensus, and I've clarified that turning a pin will result in a DNF.
> 
> Another option would be to remove F3a++ and instead have something like the following:
> 
> F3a) The competitor must not push pins during inspection. Penalty: time penalty (+2 seconds). F3b) The competitor must not turn pins during inspection. Penalty: disqualification of the attempt (DNF).

I believe that moving F3a++ to F3b would be appropriate for clarity.

Samuel-Baird:
> I've updated the PR to make time penalties not cumulative, as seems to be the general consensus, and I've clarified that turning a pin will result in a DNF.
> 
> Another option would be to remove F3a++ and instead have something like the following:
> 
> F3a) The competitor must not push pins during inspection. Penalty: time penalty (+2 seconds). F3b) The competitor must not turn pins during inspection. Penalty: disqualification of the attempt (DNF).

I believe that moving F3a++ to F3b would be appropriate for clarity as currently F3a++ doesn't act as a clarification, but rather as an addition.

wca-regulations.md
```diff
@@ -516,7 +516,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - F2a) The organization team may enforce using a stand for the puzzle to prevent it from falling before the start of the attempt. If such stands are used, the organization team should make an announcement before the round starts.
         - F2a1) The judge must remove the stand from the mat immediately after the competitor removes the puzzle from it.
 - F3) At the end of the inspection period, the competitor places the puzzle in a standing position. Penalty: disqualification of the attempt (DNF).
-    - F3a) The competitor must not change the positions of any pins from their scrambled positions before the start of the solve. Penalty: disqualification of the attempt (DNF).
+    - F3a) The competitor must not change the positions of any pins from their scrambled positions before the start of the solve. Penalty: time penalty (+2 seconds).
```
lgarron, line 519: Nit: I'd spell this out here:
```suggestion
    - F3a) The competitor must not change the positions of any pins from their scrambled positions before the start of the solve. Penalty for any number of pin changes: time penalty (+2 seconds).
```

Rouxles: We won't be implementing this change for this cycle, so we won't have to worry about this PR

Rouxles:
Closing this PR for now because it won't be implemented for Jan 2023

PR #1086, opened 2022-11-16
2k6 "expectation" -> "intention"

andrewtyberg:
I've always thought the word "expectation" doesn't really fit here. If someone is competing in 5BLD, they would probably "expect" to get a DNF, simply because the global DNF rate for the event is about 80%. If taken literally, this regulation would prevent competitors from competing an any event where they have a more than 50% DNF rate. Using the word "intention" is more logical and better fulfills the intent of the regulation.

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -100,7 +100,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 2k3) Behaves in a way that is unlawful/violent/indecent/unsafe, or intentionally damages venue facilities or personal property within the venue.
     - 2k4) Distracts or interferes with others during the competition.
     - 2k5) Fails to abide by WCA Regulations during the competition.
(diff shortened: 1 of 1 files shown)


lgarron:
The original was chosen specifically to include people who *want* to succeed in an event, but are lacking the skills — e.g. solving 4x4x4 using reduction but not knowing OLL or PLL parity.

I don't think "intention" is a significant improvement — it's one of those concepts that the law tends to stay away from, because it's a more subjective concept you can't measure from the outside. But I'd be interested in others' thoughts.

Randomno:
I feel like 2k6 has two separate parts anyway. If you can't solve the puzzle you don't meet the first part (fulfilling the event's requirements). And there are plenty of instances of getting an intentionally poor result while fulfilling the event's requirements (whatever that actually means, because I don't think it's well defined).

Rouxles:
closing for now, can revisit later

PR #1088, opened 2022-11-19
Allow clock pins to be pushed during inspection

Rouxles:


Samuel-Baird:
Everything in article F except F1 could be removed as they are not needed (one of the benefits of this change), then the following could be added:

`- F2) The competitor may push any pins, altering their positions, during inspection. `
`F2+) Pins must not be twisted/turned during inspection, altering the inner clock faces' position. Penalty: disqualification of the attempt (DNF).`

Alternatively, article F could be removed and these regulations could be moved elsewhere, perhaps [A3c](https://www.worldcubeassociation.org/regulations/#A3c)? 
`- A3c5) For clock, the competitor may push any pins, altering their positions, during inspection. `
`-A3c5+) For clock, pins must not be twisted/turned during inspection, altering the inner clock faces' position. Penalty: disqualification of the attempt (DNF).`

Also as a small note, `before the start of the solve` has been changed to `during inspection`  as this period is more frequently referred to as `inspection` than `before the start of the solve` in the Regulations

andrewtyberg:
I still think a guideline about not being allowed to turn pins may be a good idea.

sam596:
I agree that this means that Article F could be removed and A3c5 and A3c5+ created, would suggest this wording:

`- A3c5) For clock, the competitor may alter the state of pins during inspection.`

`- A3c5+) CLARIFICATION: Pins must not be twisted or turned such that the inner clock faces change position as this would constitute a move of the puzzle (see Regulation A3c1).`
or `... change position. This would constitute a move of the puzzle (see Regulation A3c1).`
or `... change position, constituting a move of the puzzle (see Regulation A3c1).`

A3c1 would already cover the scenario in Samuel's A3c6 proposal, making this clarification references it and clarifies that it would be considered a move of the puzzle.

Samuel-Baird:
> I agree that this means that Article F could be removed and A3c5 and A3c5+ created, would suggest this wording:
> 
> `- A3c5) For clock, the competitor may alter the state of pins during inspection.`
> 
> `- A3c5+) CLARIFICATION: Pins must not be twisted or turned such that the inner clock faces change position as this would constitute a move of the puzzle (see Regulation A3c1).` or `... change position. This would constitute a move of the puzzle (see Regulation A3c1).` or `... change position, constituting a move of the puzzle (see Regulation A3c1).`
> 
> A3c1 would already cover the scenario in Samuel's A3c6 proposal, making this clarification references it and clarifies that it would be considered a move of the puzzle.

I like this wording much better :) also, good point with A3c1, the line about not twisting pins can definitely go to the Guidelines rather than the Regulations. 

For A3c5+ I like the wording of the 3rd option best, it seems to flow better in my opinion

Rouxles:
closing for now, will return to this later

PR #1095, opened 2022-12-28
Change "Attempt should be stopped immediately after DNF Regulation violation"

Rouxles:
Changes thewca/wca-regulations#1083

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -244,7 +244,6 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 9f1) All timed results under 10 minutes, except for 3x3x3 Multi-Blind, are measured and truncated to the nearest hundredth of a second. All timed averages and means under 10 minutes are measured and rounded to the nearest hundredth of a second.
     - 9f2) All timed results, averages, and means over 10 minutes, as well as all times for 3x3x3 Multi-Blind results, are measured and rounded to the nearest second (e.g. X.49 becomes X, X.50 becomes X+1).
     - 9f4) The result of an attempt is recorded as DNF (Did Not Finish) if the attempt is disqualified or unsolved/unfinished.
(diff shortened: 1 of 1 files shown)


wca-regulations.md
```diff
@@ -244,7 +244,6 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 9f1) All timed results under 10 minutes, except for 3x3x3 Multi-Blind, are measured and truncated to the nearest hundredth of a second. All timed averages and means under 10 minutes are measured and rounded to the nearest hundredth of a second.
     - 9f2) All timed results, averages, and means over 10 minutes, as well as all times for 3x3x3 Multi-Blind results, are measured and rounded to the nearest second (e.g. X.49 becomes X, X.50 becomes X+1).
     - 9f4) The result of an attempt is recorded as DNF (Did Not Finish) if the attempt is disqualified or unsolved/unfinished.
```
Rouxles, line 246:
```suggestion
    - 9f4) The result of an attempt is recorded as DNF (Did Not Finish) if the attempt is disqualified or unsolved/unfinished.
        - 9f4a) If a competitor violates a regulation clearly enough that the attempt is certain to be disqualified, the judge should immediately stop the attempt. If the judge is uncertain or a dispute could be disruptive (e.g. because an extra attempt could delay the competition), the judge should consult the WCA Delegate after the solve.
```

CarterKoala (approved):

lgarron (changes requested):
I'd like to be careful about adding this, since it negates the point of the Guideline.
If the solve is going to take a long time, they should flag down and consult the Delegate *during* the solve.

Otherwise, the Guideline would just say "If the judge is uncertain or a dispute could be disruptive (e.g. because an extra attempt could delay the competition), the judge should not interrupt the solve.", which I think reduces its power.

Nevseros:
FWIW, I agree with Lucas. This addition is unnecessary IMO.

PR #1098, opened 2023-01-01
Small consistency fixes

Nanush7:
Line 202: `pyraminx` -> `Pyraminx`.
Line 380: duplicated "the" removed.


wca-regulations.md
```diff
@@ -199,7 +199,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - 5b5c) If more than one part with one colored face is affected, the puzzle is considered unsolved (DNF).
         - 5b5d) If one or more parts with more than one colored face are affected, the puzzle is considered unsolved (DNF).
         - 5b5e) Regulations [5b5c](regulations:regulation:5b5c) and [5b5d](regulations:regulation:5b5d) supersede [5b5a](regulations:regulation:5b5a) and [5b5b](regulations:regulation:5b5b).
-    - 5b6) Definition: A part is one or more fully connected components from a group of puzzle components that remain connected and do not move relative to each other during ordinary operation of the puzzle (e.g. a 3x3x3 center cap, a pyraminx tip, a 3x3x3 corner piece missing one face, one face of a 3x3x3 corner piece).
```
Nevseros, line 202: IMO `pyraminx` here as a puzzle is fine to have downcased, other places it's mentioned as an event. But IDK for sure, maybe worth fixing anyway.

Nanush7: If we are making that difference, then I will revert the changes. However, although Wikipedia is not an "official" thing, it's always uppercase there https://en.wikipedia.org/wiki/Pyraminx.

lgarron: I would definitely do uppercase, since we don't have lowercase puzzle names in the Regs.

Nevseros (approved):

Nanush7:
I won't merge it yet, in case I find something else.

lgarron (approved):

PR #1110, opened 2023-04-02
Add more detail to 2g3

Rouxles:


wca-regulations.md
```diff
@@ -76,7 +76,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 2e3) Competitors who have no citizenship may compete as "Stateless".
 - 2f) Competitors must obey venue rules and conduct themselves in a considerate manner.
 - 2g) Competitors must remain quiet when inside the designated competition area. Talking is permitted, but must be kept at a reasonable level, and away from competitors who are actively competing.
-    - 2g3) Competitors in a Competitor Waiting Area must not communicate with each other about the scrambled states of the puzzles of the round in progress. Penalty: disqualification of the competitor(s) from the event, at the discretion of the WCA Delegate.
+    - 2g3) Competitors must not communicate with each other about the scrambled states of the puzzles of the group/round in progress. Penalty: disqualification of the competitor(s) from the event, at the discretion of the WCA Delegate. Exception: For 3x3x3 Fewest Moves, competitors that have finished their attmpet may discuss their solution with other competitors that have finished their attempt (see the regulations for [Article E](regulations:article:E) and the guidelines for [Article E](guidelines:article:E)).
```
```suggestion
    - 2g3) Competitors must not communicate with each other about the scrambled states of the puzzles of the group/round in progress. Penalty: disqualification of the competitor(s) from the event, at the discretion of the WCA Delegate. Exception: For 3x3x3 Fewest Moves, competitors that have finished their attmpet may discuss their solution with others (see [Article E](regulations:article:E)).
```

I feel like talking to just guests (e.g. parents) should be allowed too. Also, we do not usually link to guidelines.

ohexter:
```suggestion
    - 2g3) Competitors in a group in progress must not communicate with other competitors in the group about the scrambled states of the group's puzzles unless both competitors are not eligible for any further attempts in the group. Penalty: disqualification of the competitor(s) from the event, at the discretion of the WCA Delegate. Exception: For 3x3x3 Fewest Moves and 3x3x3 Multi-Blind competitors may communicate with each other about the scramble sequence(s) for an individual attempt provided both have finished that attempt (see [Article E](regulations:article:E) and [Article H](regulations:article:H)).
```

While the previous use of "competitor waiting area" was problematic as it did not cover the possibility of competitors communicating at solving stations (especially while fixed seating is in use), its removal make the regulation too broad. "Competitor" alone could be interpreted as being any competitor at the competition, precluding delegates, judges etc. from communicating necessary information, for example while investigating a duplicate scramble. I have attempted a draft that keeps the restricted group as narrowly defined as possible. 

"Group" is preferable to "round" as it is more specific and limiting. "Group/round" is unnecessary as per 1h competitors in a round that is not subdivided can still be considered members of a single group.

"Not eligible for any further attempts in the group" is a bit awkward, but simply saying "finished their attempts" doesn't capture scenarios in which there are DNSs. I've not tried to capture this drafting nuance in the exception.

As multiblind also has individually scheduled attempts I have added this as a further exception. 

Should we consider a broader sanction here (disqualification from further events), as the existing one may be ineffective in certain cases? For instance, 2j2 means a competitor telling another about scramble 5 of a final round after performing that scramble receives only a very limited punishment from the application of this regulation. The broader power to disqualify further under 2j without precondition, or 2k5 for breaking this specific regulation still exists if necessary.

```suggestion
    - 2g3) Competitors must not communicate with each other about the scrambled states of the puzzles of the group/round in progress. Penalty: disqualification of the competitor(s) from the event, at the discretion of the WCA Delegate. Exception: For 3x3x3 Fewest Moves, competitors that have finished their attempt may share the scramble sequence with anyone who is not actively competing in the relevant attempt.
```

Nanush7 (changes requested):
```suggestion
    - 2g3) Competitors must not communicate with each other about the scrambled states of the puzzles of the group/round in progress. Penalty: disqualification of the competitor(s) from the event, at the discretion of the WCA Delegate. Exception: For 3x3x3 Fewest Moves, competitors that have finished their attempt may share pictures of the scramble with anyone who is not actively competing in the relevant attempt.
```

The initial idea was to clarify that competitors shouldn't be punished for sharing pictures of their solution while the attempt is still ongoing (as other competitors shouldn't have access to their phones anyway). It's not a good idea to allow competitors to talk about their solutions (as per the reasons discussed in the WRC meeting).

wca-regulations.md
Nanush7: Antonio's comment is in line with the following idea:

The initial idea was to clarify that competitors shouldn't be punished for sharing pictures of their solution while the attempt is still ongoing (as other competitors shouldn't have access to their phones anyway). It's not a good idea to allow competitors to talk about their solutions, as it is not easy to regulate and control.

@Nevseros @ohexter What do you think about this?

Nanush7 (approved):

wca-regulations.md
ohexter: Apologies for the slow response - I think keeping the wording of the exception focused on scramble sequences rather than solutions is good as that's what the regulation is about. 

I stand by the points in my April 3 comment, which it's a little disappointing to see were not resolved or addressed by the version that has now been committed and reviewed.

wca-regulations.md
```diff
@@ -76,7 +76,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 2e3) Competitors who have no citizenship may compete as "Stateless".
 - 2f) Competitors must obey venue rules and conduct themselves in a considerate manner.
 - 2g) Competitors must remain quiet when inside the designated competition area. Talking is permitted, but must be kept at a reasonable level, and away from competitors who are actively competing.
-    - 2g3) Competitors in a Competitor Waiting Area must not communicate with each other about the scrambled states of the puzzles of the round in progress. Penalty: disqualification of the competitor(s) from the event, at the discretion of the WCA Delegate.
+    - 2g3) Competitors must not communicate with each other about the scrambled states of the puzzles of the group/round in progress. Penalty: disqualification of the competitor(s) from the event, at the discretion of the WCA Delegate. Exception: For 3x3x3 Fewest Moves, competitors that have finished their attempt may share the scramble sequence with anyone who is not actively competing in the relevant attempt.
```
```suggestion
    - 2g3) Competitors in a group in progress must not communicate with other competitors in the group about the scrambled states of the group unless both competitors are not eligible for any further attempts in the group. Penalty: disqualification of the competitor(s) from the event, at the discretion of the WCA Delegate. Exception: For 3x3x3 Fewest Moves, competitors that have finished their attempt may share the scramble sequence with anyone who is not actively competing in the relevant attempt.
```

Rouxles: @ohexter what about this one?

ohexter: This looks good. While multi wasn't added as a further exception it could be argued that a round isn't "in progress" in between separately scheduled attempts, so that becomes less essential, although as drafted two competitors finishing an attempt early and mentioning anything scramble-related to each other would still be in breach.

Nanush7: Isn't the exception irrelevant with this wording?

Goosly:
What was unclear in the original 2g3?

"Competitors in a Competitor Waiting Area must not communicate with each other about the scrambled states of the puzzles of the round in progress." seems perfectly clear, and doesn't need an additional exception for FMC since FMC doesn't use waiting areas between attempts.

Nevseros:
@Goosly 

Cannot recall very clear, but I'm guessing we decided to change it as competitors were technically allowed to discuss scrambles outside of the waiting area during the round (e.g when they finished an attempt and are still on their way to the waiting area) + "round" is a bit too wide.

Nanush7:
As I said before, the initial idea was to explicitly allow competitors to share pictures of the solution sheet (FMC) while others are still competing.

PR #1112, opened 2023-04-03
2i interact -> use

Rouxles:


Nevseros:
I'd say that use is fine in this regulation, but other regulations are better be revisited. Like in the recent incident, just catching falling mic should be ok, but it makes me feel that the proposed change would make this attempt DNFed.

Nanush7:
I think "use" is the best option. As per the Cambridge Dictionary, use means "to put something such as a tool, skill, or building to a particular purpose". Holding a microphone or a cellphone wouldn't be considered "use" per se. On the other hand, I guess holding an electronic device is considered to be a "physical interaction" between the person and the object.

wca-regulations.md
```diff
@@ -86,7 +86,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - 2i1d) Food and drink.
     - 2i2) Competitors may use cameras at the solving station at the discretion of the WCA Delegate, but the following restrictions apply from the start of the attempt until the competitor stops the solve. Penalty for breaking a restriction: disqualification of the attempt (DNF).
         - 2i2a) Each camera monitor must be blank or out of sight of the competitor (see [Regulation A5b](regulations:regulation:A5b)).
-        - 2i2b) The competitor must not interact with (e.g. operate, hold, wear) any active camera. Exception: the competitor may wear a camera mounted on their head, as long as it is out of their sight and it is clear that they are not interacting with it (apart from wearing it).
+        - 2i2b) The competitor must not use (e.g. operate, hold, wear) any active camera. Exception: the competitor may wear a camera mounted on their head, as long as it is out of their sight and it is clear that they are not interacting with it (apart from wearing it).
```
```suggestion
        - 2i2b) The competitor must not use (e.g. operate, wear) any active camera. Exception: the competitor may wear a camera mounted on their head, as long as it is out of their sight and it is clear that they are not interacting with it (apart from wearing it).
```

If the idea is to allow competitors to pick a camera (i.e. if it falls during the attempt), maybe we should remove "hold" from the examples. I'm not sure what was the exact idea behind this one, though.

wca-guidelines.md
```diff
@@ -50,7 +50,7 @@ To be more informative, each Guideline is classified using one of the following
 - 2e+) [CLARIFICATION] If a competitor holds multiple citizenships, they may choose which country to represent at their first competition.
 - 2e++) [ADDITION] The WCA Delegate should also verify the name and date of birth by means of documents (e.g. a passport) of every new competitor at their first competition.
 - 2e3+) [CLARIFICATION] Stateless competitors have no national records and rankings, nor continental records and rankings.
-- 2i+) [ADDITION] Although the competitor may pick up a stopwatch to view the current time (when they are not blindfolded), they must not start, stop, pause, or otherwise interact with the timekeeping of the stopwatch.
+- 2i+) [ADDITION] Although the competitor may pick up a stopwatch to view the current time (when they are not blindfolded), they must not start, stop, pause, or otherwise use the stopwatch.
```
Nanush7: As the competitor would actually want to put the stopwatch ["to a particular purpose"](https://dictionary.cambridge.org/us/dictionary/english/use) (to know the time), I think we shouldn't change the wording to "use" in this case.

wca-regulations.md
Nevseros: I think picking up the camera is similar to the incident with a mic we resolved as OK recently, I don't see the reason to make the regulations about cameras more strict than for other electronic devices, so agree with the removal

wca-guidelines.md
Nanush7:
```suggestion
- 2i+) [ADDITION] Although the competitor may pick up a stopwatch to view the current time (when they are not blindfolded), they must not start, stop, pause, or otherwise interact with the timekeeping of the stopwatch.
```

Nanush7 (approved):

ohexter:
Should 2i3 also be changed from "interact" to "use" for further consistency? 2i2 and 2i2b verbs will now be aligned as "use" and a recent ruling on a phone also considered that verb (Irish Championship 2023), not the "interact" currently within 2i3? This issue was also relevant for NAC 2022.

PR #1115, opened 2023-04-03
Make A7c2 more lenient

Rouxles:


ohexter (approved):
Assessing intentionality may be an area of challenge, however this criterion is already present in several place in the regulations and I believe the proposed change is overall beneficial and in line with the decision for Altona Algorithms Attempt 2 2021.

Nanush7 (approved):
LGTM!

PR #1116, opened 2023-04-03
Stickers

Rouxles:


ohexter:
What are "hollow stickers"? Stickers with cut outs, or holographic stickers perhaps? I couldn't find this term in use anywhere so it is unlikely to be well understood.

wca-guidelines.md
```diff
@@ -72,7 +72,7 @@ To be more informative, each Guideline is classified using one of the following
 - 3d1b+) [CLARIFICATION] Patterns do not need to be present on every face, but only on faces where a distinct color cannot be found. Pieces must not have any features (e.g. textures, pattern irregularities) that significantly distinguish them from similar pieces.
 - 3h+) [CLARIFICATION] Puzzles may be refined internally by sanding, lubrication, or modifications which improve stability (e.g. magnets).
 - 3h++) [EXAMPLE] Examples of enhancements include: new moves are possible, normal moves are impossible, more pieces or faces are visible, colors on the backside of the puzzle are visible, moves are done automatically, or the puzzles have more/different solved states.
-- 3h+++) [EXAMPLE] Carbon fiber and hollowed stickers are considered enhancements.
+- 3h+++) [CLARIFICATION] Carbon fiber stickers with textures are considered enhancements.
```
Nanush7, line 75: Maybe like this?

ohexter: > Maybe like this?

```suggestion
- 3h+++) [CLARIFICATION] Patterned carbon fiber effect stickers are considered enhancements.
```

- "Carbon fiber effect" because they're not actually carbon fiber.
- "Patterned" not "textures" because I don't think they're necessarily textured, but a printed pattern (I don't have such a puzzle to actually check).

EdHollingdale:
Suggested improvement: "stickers with a 'carbon fibre patter effect's are NOT considered enhancements". Why don't we allow them given they give no real advantage. 

In Australia the main physical toy store chain sells the MoYu carbon fibre effect pyraminxes and skewbs. The only non-online way to buy those two puzzle types in Australia is with carbon fibre stickers - all the regulation (and this clarification) does is decrease fun.

OJCubing:
> Suggested improvement: "stickers with a 'carbon fibre patter effect's are NOT considered enhancements". Why don't we allow them given they give no real advantage.

Because carbon fibre patterns are not regulated and it would be difficult to draw the exact line in the sand trying to define them. Ultimately it means people can have patterns on their cube that show the orientation of the piece, which provides more information to the competitor. Whilst it would be difficult to take advantage of this with a carbon fibre puzzle specifically, it would be very easy to abuse if we allowed patterned stickers in general (e.g. just having an arrow all pointing the same way would make it easy to distinguish between pieces at a glance).

PR #1118, opened 2023-04-09
Scramble Secrecy

Rouxles:


Nanush7:
LGTM!

Nanush7 (approved):

EdHollingdale:
Isn't this already covered by 1c? Why does the scramble accountability policy and not e.g. the WCRP get a call out in the regulations?

Nevseros:
To explicitly mention WSAP, as it's been shown recently that some delegates take it too easy.

EdHollingdale:
Is that a problem with the regulations though? I think there needs to be better education of this for Delegates and comments about it in reports, sure, but it is already covered in the regulations. 

Sometimes Delegates forget to pay their competition Dues, but I don't think the Dues policy needs to be explicitly in the regulations because we email to follow up if they don't.

Rouxles:
@EdHollingdale following up on your email in this thread:

I think this in particular is worth explicitly mentioning in the regulations. For other policies (for example, the competition announcement policy, and everything that goes on with competition dues), there tend to be other teams that help handle this, and these issues don't tend to affect a lot of competitors, but what I think makes the scramble policy matter more is because if a competition violates this scramble policy by enough to remove results, way more results are affected, which ends up being catastrophic.

Explicitly putting this in the regulations will make more delegates and competitors (even though this isn't really the target audience) aware of this, which should (hopefully) help cause fewer of these issues.

EdHollingdale:
Thanks @Rouxles that is a fair point - of all the policies it is one which probably has the biggest flow on impact to individual competitors. I just don't want us to go down a route of putting everything in policies (back) into the regulations but this is probably fine given its importance and clearer link.

xsrvmy:
Also it is a policy that scramblers need to know about, and regular competitors can be scramblers.

Rouxles:
> I just don't want us to go down a route of putting everything in policies (back) into the regulations but this is probably fine given its importance and clearer link.

We're on the same page here!

Samuel-Baird:
Would it be possible to have this regulation become 1c1 as opposed to 1l since this does "fall under" 1c already? If they are already closely related then the labeling of the regulations might as well reflect that instead of having them separated at the start and end of article 1.

PR #1120, opened 2023-04-14
Make 5b5 more consistent

Nanush7:
## Relevant Information

**This proposal is an expansion of Viktor Zenk's idea.**

5b5f+ includes a key definition. Please, read it before proceeding.

There are (at least) two possible courses, one of them should be added explicitly to the wording:

1. **Valid positions are absolute**: The idea is that valid positions are not relative to the state of other pieces. Therefore, It does not matter if the nearest valid position is currently "claimed" by other piece, it is still valid when considering the final position of the affected piece.
2. **Valid positions are relative**: Valid positions are determined depending on the current state of other pieces.

### Example
In the following example, the possible paths described above have different outcomes:
1. DNF. The nearest valid positions for the affected corner are the positions currently occupied by the blue-red-yellow and orange-blue-yellow corners. Such positions do not render the piece solved.
2. OK. As other positions are occupied, the only valid position for the corner is the solved one (it is the only position available where a corner may be normally placed).

screen_shot_2023-04-13_at_5 49 23_pm_720

### DIFF
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -103,7 +103,7 @@ To be more informative, each Guideline is classified using one of the following 
 - 5b+) [ADDITION] The competitor is not permitted to cause a puzzle defect intentionally, unless correcting a previous puzzle defect (see [Regulation 5b3b](regulations:regulation:5b3b) and [Regulation 5b3c](regulations:regulation:5b3c)). Examples of intentionally causing puzzle defects: removing an attached center cap, rotating a corner piece, peeling a sticker.
 - 5b5+) [EXAMPLE] Examples of parts physically detached, resulting in a solved state: one 3x3x3 center cap, one big cube center piece, an inner piece of a big cube.
 - 5b5++) [EXAMPLE] Examples of parts physically detached, resulting in a DNF: two 3x3x3 center caps, two big cube center pieces, a 3x3x3 edge piece, a 4x4x4 edge piece ("wing"), any combination of two pieces that both have colored parts.
-- 5b5f+) [EXAMPLE] [Visual Guide](https://drive.google.com/file/d/15XszaCGNvy3Dk6X6qERzZWZaDH1RH04z/view).
+- 5b5f+) [EXAMPLE] There is a [Visual Guide](https://drive.google.com/file/d/15XszaCGNvy3Dk6X6qERzZWZaDH1RH04z/view) available for [Regulation 5b5f](regulations:regulation:5b5f).
 - 5b5f++) [EXAMPLE] A 4x4x4 Cube edge "wing" twisted on its spot is considered solved.
 - 5b5f+++) [EXAMPLE] A corner affected by a corner twist is considered to be placed in a valid position, but such position does not render the piece solved.
 - 5c+) [REMINDER] If a competitor has a puzzle defect, this does not grant them the right to an extra attempt.
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -105,7 +105,7 @@ To be more informative, each Guideline is classified using one of the following 
 - 5b5++) [EXAMPLE] Examples of parts physically detached, resulting in a DNF: two 3x3x3 center caps, two big cube center pieces, a 3x3x3 edge piece, a 4x4x4 edge piece ("wing"), any combination of two pieces that both have colored parts.
 - 5b5+++) [EXAMPLE] Example of parts physically attached but not fully placed, resulting in a solved state: a 5x5x5 center piece twisted in its spot.
 - 5b5++++) [EXAMPLE] Example of parts physically attached but not fully placed, resulting in a DNF: a 3x3x3 edge piece slightly popped out.
-- 5b5f+) [CLARIFICATION] The position and orientation of a piece is valid in terms of mechanism, if the piece can reach it when all the pieces of the puzzle are fully attached.
+- 5b5f+) [CLARIFICATION] The position and orientation of a piece is valid in terms of mechanism, if the piece can reach it when all the pieces of the puzzle are fully placed.
 - 5b5f++) [EXAMPLE] A 4x4x4 Cube edge "wing" twisted on its spot is considered solved.
 - 5b5f+++) [EXAMPLE] A corner affected by a corner twist is considered to be placed in a valid position, but such position does not render the piece solved.
 - 5c+) [REMINDER] If a competitor has a puzzle defect, this does not grant them the right to an extra attempt.
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -103,7 +103,7 @@ To be more informative, each Guideline is classified using one of the following 
 - 5b+) [ADDITION] The competitor is not permitted to cause a puzzle defect intentionally, unless correcting a previous puzzle defect (see [Regulation 5b3b](regulations:regulation:5b3b) and [Regulation 5b3c](regulations:regulation:5b3c)). Examples of intentionally causing puzzle defects: removing an attached center cap, rotating a corner piece, peeling a sticker.
 - 5b5+) [EXAMPLE] Examples of parts physically detached, resulting in a solved state: one 3x3x3 center cap, one big cube center piece, an inner piece of a big cube.
 - 5b5++) [EXAMPLE] Examples of parts physically detached, resulting in a DNF: two 3x3x3 center caps, two big cube center pieces, a 3x3x3 edge piece, a 4x4x4 edge piece ("wing"), any combination of two pieces that both have colored parts.
-- 5b5f+) [EXAMPLE] [Visual Guide](https://drive.google.com/drive/folders/10XKuNxAkFDE2THofKRUo_RdA9yPr0NMY).
+- 5b5f+) [EXAMPLE] [Visual Guide](https://drive.google.com/file/d/15XszaCGNvy3Dk6X6qERzZWZaDH1RH04z/view).
 - 5b5f++) [EXAMPLE] A 4x4x4 Cube edge "wing" twisted on its spot is considered solved.
 - 5b5f+++) [EXAMPLE] A corner affected by a corner twist is considered to be placed in a valid position, but such position does not render the piece solved.
 - 5c+) [REMINDER] If a competitor has a puzzle defect, this does not grant them the right to an extra attempt.
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -105,6 +105,9 @@ To be more informative, each Guideline is classified using one of the following 
 - 5b5++) [EXAMPLE] Examples of parts physically detached, resulting in a DNF: two 3x3x3 center caps, two big cube center pieces, a 3x3x3 edge piece, a 4x4x4 edge piece ("wing"), any combination of two pieces that both have colored parts.
 - 5b5+++) [EXAMPLE] Example of parts physically attached but not fully placed, resulting in a solved state: a 5x5x5 center piece twisted in its spot.
 - 5b5++++) [EXAMPLE] Example of parts physically attached but not fully placed, resulting in a DNF: a 3x3x3 edge piece slightly popped out.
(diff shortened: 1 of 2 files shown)


wca-guidelines.md
```diff
@@ -105,6 +105,9 @@ To be more informative, each Guideline is classified using one of the following
 - 5b5++) [EXAMPLE] Examples of parts physically detached, resulting in a DNF: two 3x3x3 center caps, two big cube center pieces, a 3x3x3 edge piece, a 4x4x4 edge piece ("wing"), any combination of two pieces that both have colored parts.
 - 5b5+++) [EXAMPLE] Example of parts physically attached but not fully placed, resulting in a solved state: a 5x5x5 center piece twisted in its spot.
 - 5b5++++) [EXAMPLE] Example of parts physically attached but not fully placed, resulting in a DNF: a 3x3x3 edge piece slightly popped out.
+- 5b5f+) [CLARIFICATION] The position and orientation of a piece is valid in terms of mechanism, if the piece can reach it when all the pieces of the puzzle are fully attached.
```
Nanush7: Can normally reach the position when the puzzle is fully assembled.

CarterKoala: "Fully attached" should be replaced with "fully placed" here to be consistent with the rest of 5b5

```suggestion
- 5b5f+) [CLARIFICATION] The position and orientation of a piece is valid in terms of mechanism, if the piece can reach it when all the pieces of the puzzle are fully placed.
```

wca-regulations.md
```diff
@@ -199,6 +199,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - 5b5c) If more than one part with one colored face is affected, the puzzle is considered unsolved (DNF).
         - 5b5d) If one or more parts with more than one colored face are affected, the puzzle is considered unsolved (DNF).
         - 5b5e) Regulations [5b5c](regulations:regulation:5b5c) and [5b5d](regulations:regulation:5b5d) supersede [5b5a](regulations:regulation:5b5a) and [5b5b](regulations:regulation:5b5b).
+        - 5b5f) If a piece is partially detached or not fully placed, its final position is considered to be its nearest valid position and orientation in the puzzle, in terms of mechanism. If the final position renders the piece solved, the piece is not considered to be affected by the puzzle defect.
```
CarterKoala, line 202: One possible wording suggestion someone sent to me:

wca-guidelines.md
```diff
@@ -105,6 +105,9 @@ To be more informative, each Guideline is classified using one of the following
 - 5b5++) [EXAMPLE] Examples of parts physically detached, resulting in a DNF: two 3x3x3 center caps, two big cube center pieces, a 3x3x3 edge piece, a 4x4x4 edge piece ("wing"), any combination of two pieces that both have colored parts.
 - 5b5+++) [EXAMPLE] Example of parts physically attached but not fully placed, resulting in a solved state: a 5x5x5 center piece twisted in its spot.
 - 5b5++++) [EXAMPLE] Example of parts physically attached but not fully placed, resulting in a DNF: a 3x3x3 edge piece slightly popped out.
```
CarterKoala: Note: the examples in the guidelines will need to be updated to account for this potential change, since they mostly refer to partially detached pieces

viktorzenk:
Hi, thank you for the work you're putting into this!

I would be in favor of option 1, **Valid positions are absolute**. This is what I had in mind with my original proposal, and I believe it is the simpler option of the two. With relative positioning, I feel things may get complicated if there are multiple pieces to consider.

I like the proposed wordings of 5b5f and 5b5f+. I think there is also a need for examples, but the meanings of 5b5f++ and 5b5f+++ are not immediately clear to me with the current proposed wordings. 

- 5b5f++ perhaps needs to clarify by how much the wing is twisted. I suppose that (anything up to) 180 degrees is meant, and that the piece is considered not affected because the nearest valid orientation is still the "correct" orientation per 5b5f, since a wing cannot mechanically be oriented in a different way.
- 5b5f+++ I think this wording is a bit confusing, and should mention that the piece is not in the correct orientation, connecting back to the wording of 5b5f.

Here are some proposed rewordings:

- 5b5f++) [EXAMPLE] A 4x4x4 Cube edge "wing" twisted 180 degrees in its spot is considered solved, since the nearest mechanically valid orientation renders the piece solved.
- 5b5f+++) [EXAMPLE] A corner affected by a corner twist is not considered solved, since the position is mechanically valid but the orientation is incorrect.

One additional thought I just had; Perhaps it should be clarified somewhere that these regulations do not apply if the piece is completely detached (popped). Otherwise one might think that for example a popped corner could be considered solved even if it is on the floor a meter away from the puzzle, as long as its slot is the nearest position in the puzzle.
I'm not sure if this is best added to 5b5f (e.g. "If a piece is **in contact with the puzzle but** partially detached or not fully placed, ..."), or as a separate clarification (i.e. 5b5f++++).

OJCubing:
> One additional thought I just had; Perhaps it should be clarified somewhere that these regulations do not apply if the piece is completely detached (popped). Otherwise one might think that for example a popped corner could be considered solved even if it is on the floor a meter away from the puzzle, as long as its slot is the nearest position in the puzzle. I'm not sure if this is best added to 5b5f (e.g. "If a piece is **in contact with the puzzle but** partially detached or not fully placed, ..."), or as a separate clarification (i.e. 5b5f++++).

I think the words "partially detached" covers exactly this. It should be clear that a piece that isn't in contact with the puzzle wouldn't be classified as partially detached.

Other comments I had: First of all thanks Viktor for the initial proposal, it's certainly made things easier for us!
Second, I think option 2 is my preferred one, just because it makes more intuitive sense, but option 1 is probably the best for actually regulating the outcome.

nsilvestri:
I'm quite confused by the phrase "in terms of mechanism". Given the context of previous discussions, I think I understand the intent of this phrase, but it could be clarified, especially for competitors who read this regulation.

Also, under this new wording, I have a question about a hypothetical incident might be ruled. A 3x3 edge pops during an attempt, and the competitor fully replaces it in a flipped orientation. At the end of the attempt, this flipped edge is the only unsolved piece. Knowing this would be a DNF if the timer is stopped in the puzzle's current state, the competitor slightly pops this piece, leaving it partially attached but not fully placed. This new regulation stipulates "its final position is considered to be its nearest valid position and orientation in the puzzle", which implicitly fixes its orientation, and the timer can be stopped and the puzzle ruled as solved.

Is this an accurate or intended interpretation of this regulation?

CarterKoala:
> I'm quite confused by the phrase "in terms of mechanism". Given the context of previous discussions, I think I understand the intent of this phrase, but it could be clarified, especially for competitors who read this regulation.

I agree it's not the best phrase :P We were trying to think of alternate wordings but we were having a bit of trouble. Maybe instead of "in terms of mechanism", we say something about the nearest "fully placed" state in 5b5f and remove 5b5f+ altogether? I'm not sure exactly, but maybe someone else has a better wording.

> Also, under this new wording, I have a question about a hypothetical incident might be ruled. A 3x3 edge pops during an attempt, and the competitor fully replaces it in a flipped orientation. At the end of the attempt, this flipped edge is the only unsolved piece. Knowing this would be a DNF if the timer is stopped in the puzzle's current state, the competitor slightly pops this piece, leaving it partially attached but not fully placed. This new regulation stipulates "its final position is considered to be its nearest valid position and orientation in the puzzle", which implicitly fixes its orientation, and the timer can be stopped and the puzzle ruled as solved. Is this an accurate or intended interpretation of this regulation?

In this case, 5b5f wouldn't even apply since the piece is fully placed (and not partially detatched). However, lets take a similar hypothetical where the flipped edge is slightly popped out. The closest valid state here would be the state you described in the first hypothetical - the cube is completely solved except for one edge that is flipped, which would result in a DNF. Even though this position can't be achieved with normal moves, the edge can be flipped in its spot with normal moves (take a solved cube and do F U' R U, which flips UF). The example in 5b5f+++ is intended to clarify that these types of positions should still be ruled as DNF. 

Both of your questions here are making me think that the right wording has something to do with the "closest postition where all pieces are fully placed." @thewca/wrc-team thoughts?

sarahstrong314:
We wouldn't need to clarify the definition of a "valid position" if 5b5 were modified so that a center pop at the end of the solve is considered a DNF instead of no penalty.

The current version of 5b5 and its sub-Regulations were introduced in May 2013. My understanding is that the "one part with one colour" rule in 5b5b was intended to ensure there's no penalty when a single center cap is detached at the end of the solve. This appeared to be influenced by older models of cubes such as the Haiyan Memory that had issues with caps falling off unless you put a piece of paper under a cap to secure it in place. However, the wording of 5b5b extended to an entire center piece being detached ruled as no penalty, resulting in all kinds of unexpected, unclear scenarios, especially with big cubes and non-NxNxN puzzles.

Here's what I'm thinking:

> [5b5b](https://www.worldcubeassociation.org/regulations/#5b5b)) If one ~part with one colored face~ **cap** is affected, the puzzle is considered solved.
> [5b5c](https://www.worldcubeassociation.org/regulations/#5b5c)) If ~more than one part~ **one or more parts** with one colored face is affected, the puzzle is considered unsolved (DNF). **Exception: See Regulation 5b5b.**

These rules are certainly harsher than the proposal, however I believe it aligns with the direction that the Regulations has been going in recently. For example, disallowing logos for all blindfolded events was harsh, but it made it so much easier for Delegates since we used to have to individually feel each logo to check whether it had a noticeable texture that could be used to distinguish it from other pieces.

Nanush7:
Apologies for the inactivity.

Please, have a look at the latest commit. This one was partially taken from James Hildreth's proposal in the email thread. The wording follows the second approach from the two described at the beginning of this PR. This wording does not introduce significant changes to the spirit of the previous proposal, but it aims to be more clear + explicitly includes the second option. I personally think that following the second approach may be more intuitive and therefore easier to judge.

Please, avoid suggesting the "blanketing" idea. We are aware of it, but we are currently trying to reach a more lenient option.

More examples should be added.

wca-guidelines.md
```diff
@@ -103,8 +103,9 @@ To be more informative, each Guideline is classified using one of the following
 - 5b+) [ADDITION] The competitor is not permitted to cause a puzzle defect intentionally, unless correcting a previous puzzle defect (see [Regulation 5b3b](regulations:regulation:5b3b) and [Regulation 5b3c](regulations:regulation:5b3c)). Examples of intentionally causing puzzle defects: removing an attached center cap, rotating a corner piece, peeling a sticker.
 - 5b5+) [EXAMPLE] Examples of parts physically detached, resulting in a solved state: one 3x3x3 center cap, one big cube center piece, an inner piece of a big cube.
 - 5b5++) [EXAMPLE] Examples of parts physically detached, resulting in a DNF: two 3x3x3 center caps, two big cube center pieces, a 3x3x3 edge piece, a 4x4x4 edge piece ("wing"), any combination of two pieces that both have colored parts.
-- 5b5+++) [EXAMPLE] Example of parts physically attached but not fully placed, resulting in a solved state: a 5x5x5 center piece twisted in its spot.
-- 5b5++++) [EXAMPLE] Example of parts physically attached but not fully placed, resulting in a DNF: a 3x3x3 edge piece slightly popped out.
+- 5b5f+) [EXAMPLE] [Visual Guide](https://drive.google.com/file/d/15XszaCGNvy3Dk6X6qERzZWZaDH1RH04z/view).
```
```suggestion
- 5b5f+) [EXAMPLE] The hyperlink contains a [Visual Guide](https://drive.google.com/file/d/15XszaCGNvy3Dk6X6qERzZWZaDH1RH04z/view) for regulation [5b5f](regulations:regulation:5b5f).
```

Nanush7: What about this one?
```suggestion
- 5b5f+) [EXAMPLE] There is a [Visual Guide](https://drive.google.com/file/d/15XszaCGNvy3Dk6X6qERzZWZaDH1RH04z/view) available for [Regulation 5b5f](regulations:regulation:5b5f).
```
"The hyperlink contains a ..." feels a bit weird IMO :p, but I would be fine with it if we can't find a better choice.

Rouxles (approved):

PR #1121, opened 2023-04-14
A6g Inconsistency Fix

### LINKED ISSUE #1119: Irrelevant exception in A6g
> [A6g](https://www.worldcubeassociation.org/regulations/#A6g)) The judge determines whether the puzzle is solved. They must not touch the puzzle before they have determined whether to assign a penalty for misalignment. Exception: For Clock, the judge will usually need to pick up the puzzle to verify both faces.

**Describe the typo**
Clock does not have a misalignment penalty, so the exception is irrelevant.

**Possible solutions**
* The last sentence can be omitted altogether.
* Replace "penalty for misalignment" with "penalty for misalignment or a penalty for an unsolved puzzle", although other exceptions should be included too (e.g. picking up the puzzle at the end of a blindfolded solve).

CarterKoala:
Closes #1119

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -434,7 +434,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - A6e2) Penalty for touching the puzzle after stopping the timer without applying any moves: time penalty (+2 seconds). Exception: if immediately after stopping the timer, the competitor briefly touches the puzzle without affecting its state, the time penalty may be lifted, at the discretion of the judge (see [Regulation 11f](regulations:regulation:11f)).
     - A6f) The competitor must not reset the timer until both the competitor and the judge have signed the score sheet. Penalty: disqualification of the attempt (DNF), at the discretion of the judge.
         - A6f1) If the competitor resets the timer before the result has been completely recorded, the judge must not write down the result from memory or from video or photographic evidence, and must disqualify the attempt instead (DNF).
-    - A6g) The judge determines whether the puzzle is solved. They must not touch the puzzle before they have determined whether to assign a penalty for misalignment. Exception: For Clock, the judge will usually need to pick up the puzzle to verify both faces.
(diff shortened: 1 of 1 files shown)


CarterKoala:
@sarahstrong314 Does this look good? Not exactly what you suggested but it's more concise and gets the job done.

sarahstrong314:
> @sarahstrong314 Does this look good? Not exactly what you suggested but it's more concise and gets the job done.

Would this mean that the judge can no longer pick up the puzzle during blindfolded events to check whether they are solved? If so, this would make judging 3x3x3 Multi-Blind particularly difficult.

wca-regulations.md
```diff
@@ -434,7 +434,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
...
     - A6f) The competitor must not reset the timer until both the competitor and the judge have signed the score sheet. Penalty: disqualification of the attempt (DNF), at the discretion of the judge.
         - A6f1) If the competitor resets the timer before the result has been completely recorded, the judge must not write down the result from memory or from video or photographic evidence, and must disqualify the attempt instead (DNF).
-    - A6g) The judge determines whether the puzzle is solved. They must not touch the puzzle before they have determined whether to assign a penalty for misalignment. Exception: For Clock, the judge will usually need to pick up the puzzle to verify both faces.
+    - A6g) The judge determines whether the puzzle is solved. They must not touch the puzzle before they have determined the state of the puzzle. Exception: For Clock, the judge will usually need to pick up the puzzle to verify both faces.
```
```suggestion
    - A6g) The judge determines whether the puzzle is solved. If the puzzle is misaligned, it must not be touched before the judge has determined its final state.
```

As far as I know, this rule is supposed to prevent anyone from touching the puzzle when it has a misalignment. So I think something like this should work.

Rouxles (approved):

PR #1126, opened 2023-05-08
Remove the blindfolded logo restriction

conard-milliononmars:
A logo on a center piece only impacts blindfolded solving if the logo can be detected while blindfolded, which means it is either engraved, a sticker, or otherwise changing the texture of that center piece. If the logo is smoothly part of the plastic with no detectable texture/elevation changes, then it does not provide additional information while blindfolded.

There are two main options, since allowing only undetectable logos is regarded as non-viable:
1. Allowing all logos in blind. Consider dropping your cube during blind: a competitor with a touch detectable logo can restore their orientation with 25% odds as compared to 4% odds without a touch detectable logo (quick maths, I'm tired). If all logos are allowed, it's on the competitor to use a cube with a detectable logo to maximize this advantage in the case of a cube drop. This has many pros, as stated by Edward: being nice to people, ease of enforcement, better availability of hardware, avoids damage caused by removing logos. Note: this also impacts MBLD in addition to cube drops, since puzzles are picked up from table while blind.

2. Close this PR: no logos allowed, cube drop recovery chance stays at 4%. The main pros are that it's also easily enforceable and that it maintains the true integrity of blind solving, but has cons that mirror the pros of option 1.

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -147,7 +147,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 3k2) Penalty for attempts done with puzzles which are not permitted: disqualification of the attempt (DNF). The following exceptions apply:
         - 3k2a) If a non-permitted puzzle is found before a round is complete, affected results in the round may be replaced with extra attempts, at the discretion of the WCA Delegate.
         - 3k2b) 3x3x3 Multi-Blind: if any puzzles are found to be non-permitted, such puzzles may be individually counted as unsolved (without disqualifying the entire attempt), at the discretion of the WCA Delegate.
-- 3l) A puzzle may have a logo on a colored part. If it does, it must have at most one colored part with a logo. Exception: For blindfolded events, a puzzle must not have a logo.
+- 3l) A puzzle may have a logo on a colored part. If it does, it must have at most one colored part with a logo. Exception: For blindfolded events, a puzzle may only have a logo that is not detectable by touch.
     - 3l1) The logo must be placed on a center piece. Exceptions for puzzles that do not have center pieces:
         - 3l1a) For Pyraminx and 2x2x2, the logo may be on any piece.
         - 3l1b) For Square-1, the logo must be on a piece in the equatorial slice.
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -147,7 +147,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 3k2) Penalty for attempts done with puzzles which are not permitted: disqualification of the attempt (DNF). The following exceptions apply:
         - 3k2a) If a non-permitted puzzle is found before a round is complete, affected results in the round may be replaced with extra attempts, at the discretion of the WCA Delegate.
         - 3k2b) 3x3x3 Multi-Blind: if any puzzles are found to be non-permitted, such puzzles may be individually counted as unsolved (without disqualifying the entire attempt), at the discretion of the WCA Delegate.
-- 3l) A puzzle may have a logo on a colored part. If it does, it must have at most one colored part with a logo. Exception: For blindfolded events, a puzzle may only have a logo that is not detectable by touch.
+- 3l) A puzzle may have a logo on a colored part. If it does, it must have at most one colored part with a logo.
     - 3l1) The logo must be placed on a center piece. Exceptions for puzzles that do not have center pieces:
         - 3l1a) For Pyraminx and 2x2x2, the logo may be on any piece.
         - 3l1b) For Square-1, the logo must be on a piece in the equatorial slice.


wca-regulations.md
```diff
@@ -147,7 +147,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
     - 3k2) Penalty for attempts done with puzzles which are not permitted: disqualification of the attempt (DNF). The following exceptions apply:
         - 3k2a) If a non-permitted puzzle is found before a round is complete, affected results in the round may be replaced with extra attempts, at the discretion of the WCA Delegate.
         - 3k2b) 3x3x3 Multi-Blind: if any puzzles are found to be non-permitted, such puzzles may be individually counted as unsolved (without disqualifying the entire attempt), at the discretion of the WCA Delegate.
-- 3l) A puzzle may have a logo on a colored part. If it does, it must have at most one colored part with a logo. Exception: For blindfolded events, a puzzle must not have a logo.
+- 3l) A puzzle may have a logo on a colored part. If it does, it must have at most one colored part with a logo. Exception: For blindfolded events, a puzzle may only have a logo that is not detectable by touch.
```
conard-milliononmars:
```suggestion
- 3l) A puzzle may have a logo on a colored part. If it does, it must have at most one colored part with a logo. Exception: For blindfolded events, a puzzle must not have a logo that is detectable by touch.
```

Possible better wording?

Nevseros:
Here we go again — more ambiguity. This meaning is the exact one that the Regulations moved from before, so I believe the only two ways are to allow all logos or to ban all of them (as is now).

conard-milliononmars:
I don't think this adds ambiguity. You can feel the logo or you can't -- this is a simple and reasonable test, that also makes common sense regarding the sight restriction of blind solving.

sebastianotronto:
We had this in the past, it did not work.

The problem is that some logos do not seem detectable by touch at first inspection, until someone comes up and claims they can feel it. A debate starts, and it is only settled when everybody who thought the logo was undetectable tried the following game with the puzzle in question: toss the cube in the air, catch it, hold it behind your back and try to detect the piece with the logo. Eventually, everybody is convinced that the logo is detectable. This is very time consuming.
In other cases this lead to inconsistent rulings ("the delegate in this other city let me use this cube last week!").

Frankly, I believe this is one of the best regulations we have. We had to learn it the hard way, I really do not want to see it reverted.

conard-milliononmars:
Fair enough -- I don't have a hardline stance here myself, just opening the discussion again since it was brought up. 

That said, if blind puzzle submissions that have a logo are approved by a Delegate then we could just say there is no room for debate beyond that decision, to eliminate that. Just a thought.

conard-milliononmars:
I'm also not really against allowing all logos, detectable or not. My main issue with it is it's an advantage for only some competitors (those who use a cube with a detectable logo). However, one could argue that anyone could add a logo to their cube if they want to also get that advantage, so it's their choice to forego the opportunity.

wca-regulations.md
conard-milliononmars:
```suggestion
- 3l) A puzzle may have a logo on a colored part. If it does, it must have at most one colored part with a logo.
```
Version allowing all logos (i.e. blind exception removed), for reference.

EdHollingdale:
As has been said, for good reasons the only two viable options are allow all logos, or no logos.

I think for a variety of reasons (being nice to people, ease of enforcement, availability of hardware, damage caused by removing logos) that allowing all logos is better. If this slightly changes the event and means a solver has a 40% success rate instead of a 35% success rate then that is something I can live with.

conard-milliononmars:
Yeah, I think the only odds impact is that 4% -> 25% recovery chance, which is then also only relevant when a cube is dropped (relatively infrequent, in my perspective). So I'd say it wouldn't even be a 5% success rate impact compared to current, probably less than 1%. 

It may have a bigger impact on MBLD than anything else, honestly.

sebastianotronto:
I don't think this should be a matter of probability, but of principle. Are we doing _blindfolded_ solving, or _blindfolded, except for this one piece that you can always locate exactly_ solving?

conard-milliononmars:
And on that principle alone, I'd say to close this PR and keep it how it is. There is something to be said for all the pros of relaxing the rule, but if we value that principle first then it doesn't matter.

julesDesjardin:
Hi, 

For the sake of completeness, I'd like to add the **actual** way of using such a logo, since the cube drop-off is merely a side effect of such a change. 
In the M2 method, UF(/FU) and DB(/BD) are notoriously different from the other targets, because the M2 move switches them. So if you encounter say UF as the first target of your pair, you need to do the UF alg, then your other target. However if it's the second target, you will have to do DB instead. 

There are several ways to avoid it : using letter pairs (like most top solvers who haven't switched to a more powerful method, so people around sub1), executing in pairs (what I personally do most of the time), or counting on the spot when you encounter it (which I have to do sometimes when the flow of my memo is such that I don't manage to keep track of where I am pairwise). 
Using a logo you can feel (I think the idea of "allowing only logos that can't be felt" has already been correctly disproved by previous commenters) allows you to skip counting for the last method, which is the one mostly used (from my experience) when people start to switch to M2. 
To do this, simply put a logo on the U face. When you encounter an UF/DB target, feel the D face with your ring finger (extremely easy to do undetected, because it's hidden by the cube, and the judge will most likely focus on the sight blocker anyway). If you don't feel the logo, you're on an even number of M2s, if you feel it, you're on an odd number. 

This completely ruins the event, because you get an information from the cube while blindfolded. 

Now, I'd actually like to move to what bothered me the most in this thread:

>As has been said, for good reasons the only two viable options are allow all logos, or no logos.
> 
> I think for a variety of reasons (being nice to people, ease of enforcement, availability of hardware, damage caused by removing logos) that allowing all logos is better. If this slightly changes the event and means a solver has a 40% success rate instead of a 35% success rate then that is something I can live with.

I'm being completely serious in this comment, and I don't want to be disrespectful, but I sincerely think that, with such a poor opinion on the integrity of solves, you're not fit to be a Delegate. Integrity and following rules must be our highest priority. With such a reasoning that "if this competitor used the logo to solve a cube blindfolded it's not a big deal", it's very natural to move to things like "this competitor got a +2 but it's very close, and it was because of a lockup, and the timer actually made the cube move, so it's not a big deal if we don't give the +2". And I hopefully shouldn't have to explain why such a reasoning must not come from a Delegate.

julesDesjardin:
Hi, 

For the sake of completeness, I'd like to add the **actual** way of using such a logo, since the cube drop-off is merely a side effect of such a change. 
In the M2 method, UF(/FU) and DB(/BD) are notoriously different from the other targets, because the M2 move switches them. So if you encounter say UF as the first target of your pair, you need to do the UF alg, then your other target. However if it's the second target, you will have to do DB instead. 

There are several ways to avoid it : using letter pairs (like most top solvers who haven't switched to a more powerful method, so people around sub1), executing in pairs (what I personally do most of the time), or counting on the spot when you encounter it (which I have to do sometimes when the flow of my memo is such that I don't manage to keep track of where I am pairwise). 
Using a logo you can feel (I think the idea of "allowing only logos that can't be felt" has already been correctly disproved by previous commenters) allows you to skip counting for the last method, which is the one mostly used (from my experience) when people start to switch to M2. 
To do this, simply put a logo on the U face. When you encounter an UF/DB target, feel the D face with your ring finger (extremely easy to do undetected, because it's hidden by the cube, and the judge will most likely focus on the sight blocker anyway). If you don't feel the logo, you're on an even number of M2s, if you feel it, you're on an odd number. 

This completely ruins the event, because you get an information from the cube while blindfolded. 

Now, I'd actually like to move to what bothered me the most in this thread:

>As has been said, for good reasons the only two viable options are allow all logos, or no logos.
> 
> I think for a variety of reasons (being nice to people, ease of enforcement, availability of hardware, damage caused by removing logos) that allowing all logos is better. If this slightly changes the event and means a solver has a 40% success rate instead of a 35% success rate then that is something I can live with.

I'm being completely serious in this comment, and I don't want to be disrespectful, but I sincerely think that, with such a poor opinion on the integrity of solves, you're not fit to be a Delegate. Integrity and following rules must be our highest priority. With such a reasoning that "if this competitor used the logo to solve a cube blindfolded it's not a big deal", it's very natural to move to things like "this competitor got a +2 but it's very close, and it was because of a lockup, and the timer actually made the cube move, so it's not a big deal if we don't give the +2". And I hopefully shouldn't have to explain why such a reasoning must not come from a Delegate.

lgarron:
> Frankly, I believe this is one of the best regulations we have. We had to learn it the hard way, I really do not want to see it reverted.

As much as I was reluctant to disallow logos in the first place (I had logos with my name on all my cubes that were sized to be indistinguishable from other stickers on the cube), I supported the initial change and I agree that this is still one of the most hard-won Regulations that is worth keeping — for all the reasons discussed in this thread.

Over the last few years, we've also learned how much cubers *will* adjust to advantages that were not "meant" to be part of solving, in the form of table use for OH. I don't think that such changes in speedcubing are intrinsically bad, but I think it's worth being careful to preserve the spirit of the event unless we specifically want to change it.

conard-milliononmars:
> I'm being completely serious in this comment, and I don't want to be disrespectful, but I sincerely think that, with such a poor opinion on the integrity of solves, you're not fit to be a Delegate.

I don't think this is a fair generalization of Ed. The whole point of these regulation discussions is changing the rules to be more X, where X is leniency or ease of understanding/enforcing or integrity or completeness etc. Talking about changing a rule is not synonymous with breaking a rule. 

If a rule is officially changed to allow X, then X is within the new vision of "integrity". Shaming the concept of discussing rule changes because it wouldn't have integrity under the current rule set would prevent any and all rule relaxation with an iron fist, and doesn't make sense.

conard-milliononmars:
For what it's worth, in an *ideal* world I would want only non-touch detectable logos in blind because I personally believe the spirit of blind is to use memory only. My middle ground fix would be delegate approval for cubes with logos, final answer no debate allowed, but many have pointed out this realm simply failed in the real world before and isn't viable.

All that said, I think there is a perfectly reasonable argument out there some may make that blindfolded is about not using your eyes, and that touch is a sensation that should not be restricted. That is to say, using your fingers does not violate the sanctity of having your eyes closed. 

You're claiming strongly that that concept violates integrity and even that someone shouldn't be a Delegate because of it, while an argument exists that paints it as perfectly within integrity. The world is a bit too big to make such aggressive and harsh claims, I think.

cc @julesDesjardin

Rouxles:
Yep - let's calm down a bit in this thread please 🥺 

To throw my 2 cents here, from my experience doing 3BLD even at a higher level, having a logo really helps with my reorientation (after lifting the box), which is an advantage that I think isn't supposed to be a part of blindfolded solving. Regardless, I think we're all clear here that we either allow all logos, or don't allow logos at all, and for me personally, I'm on the side of not allowing logos.

Nanush7:
Thank you for your participation!

This proposal won't be implemented in this cycle. Closing.

PR #1146, opened 2023-07-31
Sync with official

Nanush7:


Nevseros (approved):
I hope this will work. But if not, try reset the next time

PR #1152, opened 2023-09-13
Add A5b+++

Rouxles:


Nanush7 (commented):
Maybe "for the competitor" instead of "of the competitor"? Just a detail.

ohexter:
Consider 'detached' for popped as well as 'detaches' for 'pops out' in A5b++ to broaden the scope to parts otherwise detached, which presumably should be considered in the same manner?

More broadly 'pop' and variants are used in the regulations in a number of places without being defined and is somewhat colloquial. These could all potentially be replaced with detach etc.

wca-guidelines.md
```diff
@@ -183,6 +183,7 @@ To be more informative, each Guideline is classified using one of the following
 - A3c4++) [CLARIFICATION] The competitor may remove the object during the solve, if they did not do it during the inspection (e.g. they forgot).
 - A5b+) [CLARIFICATION] While inspecting or solving the puzzle, the competitor may touch the puzzle with any part of their body. Exception: 3x3x3 One-Handed (see [Regulation C1b](regulations:regulation:C1b)).
 - A5b++) [CLARIFICATION] If a part of the competitor's puzzle pops out (see [Regulation 5a](regulations:regulation:5a)), the competitor must not ask for assistance retrieving or locating the part and the judge should attempt to prevent others from assisting. If someone picks up a popped part for the competitor, the judge or competitor may ask them to place it on the solving station and not to assist any further.
+- A5b+++) [CLARIFICATION] If someone picks up a popped part of the competitor, the WCA Delegate should not grant the competitor an extra attempt unless a separate incident occurs.
```
```suggestion
- A5b++) [CLARIFICATION] If a part of the competitor's puzzle is popped out or detached (see [Regulation 5a](regulations:regulation:5a)), the competitor must not ask for assistance retrieving or locating the part and the judge should attempt to prevent others from assisting. If someone picks up a popped part for the competitor, the judge or competitor may ask them to place it on the solving station and not to assist any further.
- A5b+++) [CLARIFICATION] If someone picks up a popped or detached part of the competitor, the WCA Delegate should not grant the competitor an extra attempt unless a separate incident occurs.
```

```suggestion
- A5b++) [CLARIFICATION] If a part of the competitor's puzzle is popped out or detached (see [Regulation 5a](regulations:regulation:5a)), the competitor must not ask for assistance retrieving or locating the part and the judge should attempt to prevent others from assisting. If someone picks up a popped part for the competitor, the judge or competitor may ask them to place it on the solving station and not to assist any further.
- A5b+++) [CLARIFICATION] If someone picks up a popped or detached part of the competitor's puzzle, the WCA Delegate should not grant the competitor an extra attempt unless a separate incident occurs.
```
One further unconnected suggest for something that read a little oddly when looking at this again. The part belongs to the competitor but is not a physical part of them!

Nevseros: Oops, yeah, we noticed it but I forgot to include it into the suggestion. Thanks Oliver!

wca-guidelines.md
Samuel-Baird, line 1: Maybe better wording would be 
> The WCA Delegate should not grant the competitor an additional attempt solely based on someone picking up a popped or detached part of the competitor's puzzle.

Nanush7: Thank you Samuel! Why do you think it would be better? (honest question from non-native speaker).

Samuel-Baird: The first one just doesn’t flow well, I think it may have to do with the comma (which I think isn’t necessary?) and the order. Hard to explain, so it may just need a few other native English speakers to look at it :p

Nanush7: @Nevseros WDYT? I like Samuel's version :eyes:

Nevseros: Yeah, I think it sounds better overall. But "additional" should be changed to "extra", of course.

Nevseros (approved):
LGTM!

PR #1156, opened 2023-09-13
Changes from country to region

Rouxles:


ohexter:
In many of these cases 'region of representation' may be clearer than 'region', which, careful reading of 2e aside, is likely to be interpreted as where a competitor lives or competes. 2e1 also requires updating.

wca-states.md
Nanush7, line 1: I assume we have to change the codes/names of the regions affected by the recent decision? @ohexter Do you think the WRT can handle that?

Also, I removed the "credit" section for two reasons: WCA Documents don't include credits (apart from translations) and James Molloy's email is no longer valid (I assume). Anyway, I can add it back if required.

ohexter:
I'm not sure @thewca/wrt owns this - if it was originally created by James was it WFC related? 

Results - are there any database or other places we need to update for region vs. country and the recent changes?

SAuroux:
I am not sure if I understand the full extent of this change correctly. What codes would need to be updated in the database, is there maybe a concrete example?

In any case @thewca/wrt could support here, but so could @thewca/software-team .

FinnIckler:
Code was already changed in https://github.com/thewca/worldcubeassociation.org/pull/8166

CarterKoala:
> In many of these cases 'region of representation' may be clearer than 'region', which, careful reading of 2e aside, is likely to be interpreted as where a competitor lives or competes. 2e1 also requires updating.

I think we should be careful about using terms that are not referenced anywhere else. That said I don't have any better suggestions. Maybe it is still fine to use "region of representation" since it is just a clarifier of region?

Nanush7:
Sorry, I should have been more specific. I wanted to ask if any country code requires changes. For example:

Line 184 in wca-states.md: `(TW:_Asia) Taiwan`.

OJCubing:
The name of the file should be changed to "WCA List of Recognized Regions" rather than "WCA List of Recognized Regions of Representation". In particular, the regions are also used for the countryid of the competition (and I don't think we would say the competition is representing for that region).

wca-regulations.md
```diff
@@ -67,13 +67,13 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
...
+    - 2c2) Returning competitors must provide personal information consistent with their WCA profile. If a competitor wishes to change or update their personal information between competitions (e.g. a change of region of representation, name, or gender), they must [contact the WCA Results Team](https://www.worldcubeassociation.org/contact/website).
+- 2d) A competitor's name, region of representation, gender, and competition results are considered public information. All other personal information is considered confidential, and must not be disclosed to outside organizations/persons without the consent of the competitor.
+- 2e) Competitors must represent a region of which they hold citizenship. The WCA Delegate should verify citizenship by means of documents (e.g. a passport) at their first competition. If a competitor is found ineligible to represent the region under which they have registered, the competitor may be disqualified retroactively and/or suspended, at the discretion of the WCA Board.
+    - 2e1) The eligible regions of representation are defined by the [WCA List of Recognized Countries](https://www.worldcubeassociation.org/regulations/countries/).
```
```suggestion
    - 2e1) The eligible regions of representation are defined by the [WCA List of Recognized Regions](https://www.worldcubeassociation.org/regulations/countries/).
```

wca-states.md
```diff
@@ -1,9 +1,9 @@
-# <wca-title>WCA List of Recognized Countries
+# <wca-title>WCA List of Recognized Regions of Representation
```
Nevseros:
```suggestion
# <wca-title>WCA List of Recognized Regions
```

wca-states.md
```diff
@@ -1,9 +1,9 @@
...
 <wca-states>
-<version>Version: March 06, 2020
+<version>Version: January 1, 2024
 
 
-The WCA recognizes a total of 199 countries/regions:
+The WCA recognizes a total of 199 regions of representation:
```
Nevseros:
```suggestion
The WCA recognizes a total of 199 regions:
```

Nevseros (approved):
LGTM!

PR #1157, opened 2023-09-13
Remove "adjacent" from 4d2

Rouxles:


Nevseros:
Adjacent face may be the one that is on the bottom. Discussed in the meeting, closing.

PR #1158, opened 2023-09-28
Change interact to use in 2i3

Nevseros:
Edit: Closes #1160. (Incompatible).

PR #1160, opened 2023-09-29
Remove 2i3

Nanush7:
Given that it is now legal to *interact* with a device as long as it is clear that the competitor is not using it, 2i3 is not really necessary anymore.

Closes #1158. (Incompatible).


OJCubing:
Possibly worth moving this to be a reminder? And add that (smart) watches are fine to wear so long as you're not using them (this may already exist in the guidelines, I haven't checked properly just now).

Nevseros (approved):
LGTM!

PR #1163, opened 2023-09-29
Clarify timing inspection

Nevseros:


wca-regulations.md
```diff
@@ -415,7 +415,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 - A4) Starting the solve:
     - A4b) The competitor uses their fingers to touch the elevated sensor surfaces of the Stackmat timer. The competitor's palms must be facing down, and located on the side of the timer that is closer to the competitor. Penalty: time penalty (+2 seconds).
         - A4b1) The competitor must have no physical contact with the puzzle while starting the solve. Penalty: time penalty (+2 seconds).
-    - A4d) If a Stackmat timer is in use, the competitor should keep their hands on the timer until they see a green timer light. They start the solve by removing their hands from the timer (thus starting the timer).
+    - A4d) If a Stackmat timer is in use, the competitor should keep their hands on the timer until they see a green timer light. They start the solve by removing their hands from the timer and starting the timer.
```
CarterKoala:
```suggestion
    - A4d) If a Stackmat timer is in use, the competitor should keep their hands on the timer until they see a green timer light. The timer is started when they remove their hand(s) from the timer. The competitor starts the solve by starting the timer.
```

wca-regulations.md
```diff
@@ -463,7 +463,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
...
     - B2b) When using the Stackmat timer, the competitor uses their fingers to touch the elevated sensor surfaces of the timer. The competitor's palms must be facing down, and located on the side of the timer that is closer to them. Penalty: time penalty (+2 seconds).
     - B2c) When using the Stackmat timer, the competitor must have no physical contact with the puzzle or the puzzle cover while starting the attempt. Penalty: time penalty (+2 seconds).
-    - B2d) When using the Stackmat timer, the competitor starts the attempt by removing their hands from the timer, thus starting the timer. (This also starts the solve.)
+    - B2d) When using the Stackmat timer, the competitor starts the attempt by removing their hands from the timer and starting the timer. (This also starts the solve.)
```
CarterKoala:
```suggestion
    - B2d) When using the Stackmat timer, the competitor should keep their hands on the timer until they see a green timer light. The timer is started when they remove their hand(s) from the timer. The competitor starts the attempt and the solve by starting the timer.
```
Not sure if we should update "When using the Stackmat timer" to "If a Stackmat timer is in use" since that is what is used in Article A

wca-regulations.md
Nanush7:
```suggestion
    - A4d) If a Stackmat timer is in use, the competitor should keep their hands on the timer until they see the green timer light. The timer is started when the green light of the timer is on and the competitor removes a hand from the timer. The competitor starts the solve by starting the timer.
```

OJCubing:
I think this PR has missed the point entirely of my original intention for it. The current suggestion clarifies the start of the solve better from the competitor's point of view, but I think it is important to clarify from the judge's point of view.

For example:
A3b5) The judge stops timing inspection when the competitor lifts their hands from the timer and the timer has started.

Importantly, A) bad judges will stop timing inspection when the competitor puts their hands on the timer and B) even good judges will stop when the competitor lifts their hands from the timer, regardless of whether the timer has started or not.

I actually think it is too difficult to tell the timer has started, so in my opinion, the solve should start once the competitor releases their hands from the timer, regardless of whether it has started or not (in other words, no retries), but I think people aren't in favour of that.

wca-regulations.md
Nevseros: @Nanush7 IMO, overkill

CarterKoala (approved):

PR #1166, opened 2023-10-04
Fix inspection timeframe and add 2 clarifications

Nevseros:


nsilvestri:
I prefer this option over #1170 because it is clearer to me that passing the 15-second threshold numerically is also passing the penalty threshold. It's the same vein for people calling a 4.99 solve a "4" even though it is much closer to 5. By extension, it feels easier to communicate that a "15 or 16" is a penalty, and a "17 or higher" is a DNF, rather than make exceptions for "exactly 15".

Similarly, I feel like choosing `> 15.00` leaves some ambiguity around timers with thousandths-place precision. I could see a confused judge make the mistake of ruling a 15.001 as OK because they're used to dropping the thousandths-place when recording attempt times.

Rouxles (approved):

PR #1167, opened 2023-10-04
Remove Square-1 required scrambling orientation

Nevseros:
This is a controversial change and we will gather feedback on this. If you are a staff member, please hold on to discuss it via email a bit later ;).

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -174,7 +174,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 - 4d) Scrambling orientation:
     - 4d1) NxNxN Cubes and Megaminx are scrambled starting with the white face (if not possible, then the lightest face) on the top and the darkest adjacent green face (if not possible, then the darkest adjacent face) on the front.
     - 4d2) Pyraminx is scrambled starting with the yellow face (if not possible, then the lightest face) on the bottom and the green face (if not possible, then the darkest adjacent face) on the front.
-    - 4d3) Square-1 is scrambled starting with either side in front.
+    - 4d3) Square-1 is scrambled starting with either side in front (see [Regulation 12c1](regulations:regulation:12c1)).
     - 4d4) Clock is scrambled starting with either side in front and 12 o'clock pointing up.
     - 4d5) Skewb is scrambled starting with the white face (if not possible, then the lightest face) on top and the green face (if not possible, then the darkest adjacent face) on the front-left.
 - 4f) Competition scramble sequences must be generated using a current official version of an official WCA scramble program (available [via the WCA website](https://www.worldcubeassociation.org/regulations/scrambles/)).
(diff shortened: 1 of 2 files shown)


wca-regulations.md
```diff
@@ -174,7 +174,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
 - 4d) Scrambling orientation:
     - 4d1) NxNxN Cubes and Megaminx are scrambled starting with the white face (if not possible, then the lightest face) on the top and the darkest adjacent green face (if not possible, then the darkest adjacent face) on the front.
     - 4d2) Pyraminx is scrambled starting with the yellow face (if not possible, then the lightest face) on the bottom and the green face (if not possible, then the darkest adjacent face) on the front.
-    - 4d3) Square-1 is scrambled starting with the darker color on front (out of the 2 possible scrambling orientations).
+    - 4d3) Square-1 is scrambled starting with either side in front.
```
CarterKoala:
```suggestion
    - 4d3) Square-1 is scrambled starting with either side in front (see [Regulation 12c1](regulations:regulation:12c1)).
```

Nanush7: LGTM.

Nanush7:
Is the guideline really necessary? I think the new version of the reg is clear enough.

Nevseros:
It is clear, but I'd like us to avoid Mandela effect. We can remove this guideline later.

sebastianotronto:
What is the reasoning behind this change?

This would make Square-1 the only puzzled with colored pieces that allows multiple scrambling orientations.

Nevseros:
@sebastianotronto

We had some precedents when even top solvers didn't know there is a required orientation. We also allowed such "misscrambled" rounds in the past. There are almost no square-1 solvers for which orientation matters (AFAIU), so we decided there is no need for this.

alexmaasswca:
> @sebastianotronto
> 
> We had some precedents when even top solvers didn't know there is a required orientation. We also allowed such "misscrambled" rounds in the past. There are almost no square-1 solvers for which orientation matters (AFAIU), so we decided there is no need for this.

Putting a reminder on top of the scramble pages would completely solve that (though that's a tnoodle matter).

priazz:
> @sebastianotronto
> 
> We had some precedents when even top solvers didn't know there is a required orientation. We also allowed such "misscrambled" rounds in the past. There are almost no square-1 solvers for which orientation matters (AFAIU), so we decided there is no need for this.

Alessandro Ricci (wr7 for average) told me that after csp he can very frequently predict the first lin block of his solution. He doesn't have the time to predict all four possible blocks, so he only sticks to the green one with red in front. In his case, a change in orientation would impact the solve significantly, because he would be doing the opposite block. On the other hand, he also told me that for lars method users and other lower level lin users a y2 before the scramble would not change anything since these solvers already look for the best orientation for solving the cubeshape regardless of the colours. For what it's worth, he is against the proposed change.

I am personally also against the change since it would impact (some) high level solvers without being the solution of any real major issues from a competitor's, an organizer's or a delegate's perspective. Square-1 is also the easiest event to start scrambling in the correct orientation, even with unconventional color schemes.

Rouxles (approved):
approving, but not necessarily going to implement

Nevseros:
Decided to not implement

PR #1170, opened 2023-10-11
Alternate fix of inspection timeframe

CarterKoala:
Same idea as #1166, would implement one or the other. This one has 15.00 seconds of inspection as no penalty


nsilvestri:
Left this comment on the other PR but it seems more relevant here:

I feel like this option leaves some ambiguity around timers with thousandths-place precision. I could see a confused judge make the mistake of ruling a 15.001 as OK because they're used to dropping the thousandths-place when recording attempt times.

Rouxles (approved):
approve, but not necessarily implementing

Nevseros:
Closing and merging #1163 instead

PR #1171, opened 2023-10-12
Make cameras regulations more lenient

Nevseros:
Basically, to allow to film solves in any way.

### DIFF
--- a/wca-regulations.md
+++ b/wca-regulations.md
@@ -87,7 +87,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - 2i1d) Food and drink.
     - 2i2) Competitors may use cameras at the solving station at the discretion of the WCA Delegate, but the following restrictions apply from the start of the attempt until the competitor stops the solve. Penalty for breaking a restriction: disqualification of the attempt (DNF).
         - 2i2a) Each camera monitor must be blank or out of sight of the competitor (see [Regulation A5b](regulations:regulation:A5b)).
-        - 2i2b) The competitor must not use (e.g. operate) any active camera. Exception: the competitor may hold or wear a camera anywhere, as long as the camera's monitor is out of their sight and it is clear that they are not interacting with it (apart from wearing or holding it).
+        - 2i2b) The competitor must not use (e.g. operate) any active camera. Exception: the competitor may hold or wear a camera anywhere, as long as the camera's monitor is out of their sight or blank and it is clear that they are not interacting with it (apart from wearing or holding it).
     - 2i3) The competitor may have a cell phone in their pocket, as long as it is clear that they are not otherwise interacting with it.
     - 2i4) Competitors should turn off all cell phone notifications while competing to avoid disturbing the competition.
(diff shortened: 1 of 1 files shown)


mintydev789:
I think this makes sense. Looking at the back face using your camera screen isn't as fast as just looking yourself anyways. This is kinda similar to the issue with stickerless cubes, where it **seems** like an issue, but isn't one in practice.

The only issue I can potentially see is people abusing this by putting up the camera right next to the cube and having a decently-sized screen, so that it **is** faster, cause they can see the back face much more clearly. But the potential benefit of even this seems like it would be useless compared to the drawbacks. Maybe we can say that the camera must not be touching the mat?

mintydev789:
My bad, I didn't pay close attention when reviewing this. I'd still suggest allowing the screen to be visible even (just remove any wording about the screen). Just require the camera to be off the mat.

Samuel-Baird (commented):
Is there any particular reason why monitors must still be out of sight?

xsrvmy:
> Maybe we can say that the camera must not be touching the mat?

This should at most apply if the screen is visible. I have been to comps where the table is very small and there isn't much space to put a camera. I don't think a tripod stand would even fit off the mat. There is also no standard mat size.

mintydev789:
@xsrvmy then we could just say you can't touch the camera or have the cube touch the camera during the solve (although during inspection is fine)

wca-regulations.md
```diff
@@ -87,7 +87,7 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
         - 2i1d) Food and drink.
     - 2i2) Competitors may use cameras at the solving station at the discretion of the WCA Delegate, but the following restrictions apply from the start of the attempt until the competitor stops the solve. Penalty for breaking a restriction: disqualification of the attempt (DNF).
         - 2i2a) Each camera monitor must be blank or out of sight of the competitor (see [Regulation A5b](regulations:regulation:A5b)).
-        - 2i2b) The competitor must not use (e.g. operate, wear) any active camera. Exception: the competitor may wear a camera mounted on their head, as long as it is out of their sight and it is clear that they are not interacting with it (apart from wearing it).
+        - 2i2b) The competitor must not use (e.g. operate) any active camera. Exception: the competitor may hold or wear a camera anywhere, as long as the camera's monitor is out of their sight and it is clear that they are not interacting with it (apart from wearing or holding it).
```
```suggestion
        - 2i2b) The competitor must not use (e.g. operate) any active camera. Exception: the competitor may hold or wear a camera anywhere, as long as the camera's monitor is out of their sight or blank and it is clear that they are not interacting with it (apart from wearing or holding it).
```

ohexter: 'anywhere' reads somewhat awkwardly and only makes sense in the context of updating the previous regulation rather than having a regulation that stands on its own feet. If this context is considered necessary we could either flag this word for future removal or remove it now and add a 'previously...' guideline.

mintydev789:
2i2b could be worded more simply like this: the competitor must not operate any...

Nanush7:
Moving to another branch. See #1182.

PR #1172, opened 2023-10-12
Allow to move clock pins

Nevseros:
This is a controversial change.

### DIFF
--- a/wca-guidelines.md
+++ b/wca-guidelines.md
@@ -159,7 +159,7 @@ To be more informative, each Guideline is classified using one of the following 
 ## <article-12><notation><notation> Article 12: Notation
 
 - 12a2+) [CLARIFICATION] For example, both Rw and 2Rw are valid notation for the same physical move for the 3x3x3 Cube. 1Rw is not valid notation for any NxNxN Cube, and 3Rw is not valid notation for the 2x2x2 Cube or the 3x3x3 Cube (but 3Rw is valid notation for the 4x4x4 Cube and larger NxNxN Cubes).
-- 12g2+) [REMINDER] In the past, these notations were considered to be moves. Pin notation now can be applied during inspection period.
+- 12g2+) [REMINDER] In the past, pins notation was considered to be moves. Pins notation now can be applied during inspection period.
 - 12i+) [CLARIFICATION] Rotations do not change the puzzle state.
 
 
@@ -191,7 +191,6 @@ To be more informative, each Guideline is classified using one of the following 
 - A7g++) [CLARIFICATION] Individual penalties cannot be waived. A full extra attempt must be granted.
 - A7g+++) [RECOMMENDATION] The WCA Delegate should use stronger discretion when granting more than one extra attempt to the same competitor. If more than one extra attempt is granted, it should be for a different type of incident. Example: if the WCA Delegate already awarded an extra attempt for a competitor who did not wait long enough for the green light, they should not award another extra attempt if that competitor fails to wait long enough for the green light again.
 - A7g++++) [CLARIFICATION] A competitor may be considered inexperienced if they are a new competitor, it is their first time competing in a specific event, or they have not competed in a specific event for a significant period of time. The inexperience must be related to the incident.
-
 
 
 ## <article-B><blindfolded><blindfoldedsolving> Article B: Blindfolded Solving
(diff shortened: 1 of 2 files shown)


wca-regulations.md
```diff
@@ -529,11 +529,12 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
...
-- F2) The judge places the scrambled puzzle onto the mat in a standing position.
+- F2) The judge places the scrambled puzzle onto the mat in any position.
     - F2a) The organization team may enforce using a stand for the puzzle to prevent it from falling before the start of the attempt. If such stands are used, the organization team should make an announcement before the round starts.
         - F2a1) The judge must remove the stand from the mat immediately after the competitor removes the puzzle from it.
+        - F2a2) The competitor may change the state of the pins during the inspection. These changes are not counted as moves.
+        - F2a3) The competitor must not change the state of any inner clock faces in the puzzle during the inspection. Penalty: disqualification of the attempt (DNF).
 - F3) At the end of the inspection period, the competitor places the puzzle in a standing position. Penalty: disqualification of the attempt (DNF).
```
```suggestion
- F3) At the end of the inspection period, the competitor places the puzzle in any position.
```

wca-guidelines.md
```diff
@@ -159,6 +159,7 @@ To be more informative, each Guideline is classified using one of the following
 ## <article-12><notation><notation> Article 12: Notation
 
 - 12a2+) [CLARIFICATION] For example, both Rw and 2Rw are valid notation for the same physical move for the 3x3x3 Cube. 1Rw is not valid notation for any NxNxN Cube, and 3Rw is not valid notation for the 2x2x2 Cube or the 3x3x3 Cube (but 3Rw is valid notation for the 4x4x4 Cube and larger NxNxN Cubes).
+- 12g2+) [REMINDER] In the past, these notations were considered to be moves. Pin notation now can be applied during inspection period.
```
```suggestion
- 12g2+) [REMINDER] In the past, pins notation was considered to be moves. Pins notation now can be applied during inspection period.
```

wca-regulations.md
ohexter: I think it's orientation not position (location) that needs to be emphasised ('standing position' is a matter of orientation). And a penalty for placing the puzzle not on the mat would still apply. Making these updates would just lead this regulation becoming a copy of A3d.  So to avoid redundancy this reg could be removed and temporarily replaced with a reminder guideline like the proposed 12g2+.

Nevseros: That makes sense, thanks Oliver!

```suggestion
```

wca-guidelines.md
```diff
@@ -229,6 +229,11 @@ To be more informative, each Guideline is classified using one of the following
 - E3d++) [CLARIFICATION] Competitors should not consider a personal stopwatch or watch as the official time, and must submit their solution when the judge calls "STOP".
 
 
+## <article-F><clock><clocksolving> Article F: Clock Solving
+
+- F3+) [REMINDER] In the past, the competitor had to place the puzzle on a mat in a standing position. This is no longer required.
```
Nanush7, line 234:
```suggestion
- F3+) [REMINDER] In the past, the competitor had to place the puzzle onto the mat in a standing position. This is no longer required.
```

wca-guidelines.md
```diff
@@ -159,6 +159,7 @@ To be more informative, each Guideline is classified using one of the following
 ## <article-12><notation><notation> Article 12: Notation
 
 - 12a2+) [CLARIFICATION] For example, both Rw and 2Rw are valid notation for the same physical move for the 3x3x3 Cube. 1Rw is not valid notation for any NxNxN Cube, and 3Rw is not valid notation for the 2x2x2 Cube or the 3x3x3 Cube (but 3Rw is valid notation for the 4x4x4 Cube and larger NxNxN Cubes).
+- 12g2+) [REMINDER] In the past, pins notation was considered to be moves. Pins notation now can be applied during inspection period.
```
CarterKoala, line 162:
```suggestion
- 12g2+) [REMINDER] In the past, the penalty for changing state of the pins in inspection was a DNF. This is no longer the case. 
```
I don't think "pins notation" is correct wording and it doesn't make much sense to me. I think this is much clearer

Nevseros:
```suggestion
- 12g2+) [REMINDER] In the past, the penalty for changing state of the pins in inspection resulted in a DNF. This is no longer the case.
```

Maybe this?

PR #1173, opened 2023-10-13
Allow to push Clock pins during inspection

Nevseros:


mintydev789:
Wait, really? This has already been discussed and it's now certain this will be added? I strongly believe clock pins should just be all pushed in the same direction for consistency. I think allowing competitors to apply moves in inspection sets a bad precedent, and judges will be confused by yet another event that has an exception to the standard rules.

EdHollingdale:
With this PR 4g2 also can be removed :)

Pasqui03:
> Wait, really? This has already been discussed and it's now certain this will be added? I strongly believe clock pins should just be all pushed in the same direction for consistency. I think allowing competitors to apply moves in inspection sets a bad precedent, and judges will be confused by yet another event that has an exception to the standard rules.

I am a clock solver and I would like to say that allowing to push clock pins is a game changer in my opinion because it reduces plenty of extras for changing pins during transport and for fallen clocks. As a result, the duration of rounds for that event in schedules would be reduced and there would be more spaces for other events. 
To avoid the confusion by the judges, delegates/organizer should announce that pushing the clock pins is not DNF and it should be enough.

Mattia Pasquini (2019PASQ01)
WCA Junior Delegate

mintydev789:
@Pasqui03 changing pins during transport don't grant extras: https://www.worldcubeassociation.org/regulations/#4g2

mintydev789:
Also, we usually _think_ that telling people about an exception is enough, but it usually isn't. Most judges get used to one way of judging speed events, and they won't pay attention beyond that. If we make this change, we'll immediately have tons of extras due to judges stopping the solver after seeing them apply moves during inspection. Then you also have to make people understand the difference between a pin move and a gear move. What if they think a gear move is allowed too? I think we shouldn't underestimate the confusion this would cause for less-experienced judges.

CarterKoala:
> Wait, really? This has already been discussed and it's now certain this will be added? I strongly believe clock pins should just be all pushed in the same direction for consistency. I think allowing competitors to apply moves in inspection sets a bad precedent, and judges will be confused by yet another event that has an exception to the standard rules.

Definitely not. That's why we are bringing it for feedback - we are aware it is quite a controversial change.

Also just copying my response on the forum as for an explanation why we did not want to go with the 'same direction' approach:
> While this would definitely reduce common incidents, one major concern is the fairness of scrambles. For each scramble, different competitors will get very different pin states and would potentially have an advantage over someone else who got the “same” scramble. This is heightened for those that use 7-simul since the pin state can be a deciding factor on what orientation you start your solve in. There is also potential for abuse if you want the pins to be on the other side - you could just knock the clock over yourself to get a more favorable position. While this could possibly be solved by explicitly allowing this type of pin manipulation, we don’t think it is a great idea to allow pins to be changed via some method while not being allowed to be changed by others (such as just pushing one pin) which could be very confusing for judges.

mintydev789:
@CarterKoala oh, my thinking was that the scramble would specify to either push all pins in or out. We can choose either option as the default option. And then the competitor should not be allowed to force the one they prefer.

CarterKoala:
> @CarterKoala oh, my thinking was that the scramble would specify to either push all pins in or out. We can choose either option as the default option. And then the competitor should not be allowed to force the one they prefer.

Ah, but then you are getting rid of any benefit that making that change would provide - you still would not be able to use regular covers easily (as the orientation would be flipped when you place the cover down), and you would still have to give extras for clocks falling over.

AlexKatyukov:
One question: Why do we need pins as a scramble part if we do not count it as a solved state part?

In my opinion, the proposed solution is great! It's a game changer, but it's a game changer with clearly stated benefits.

mintydev789:
@CarterKoala I see what you mean. I'm still against it though due to my other arguments. I think we should keep this unchanged.

slongbehn:
Personally, I think allowing adjustments during inspection is quite strange. I agree that they should just all be pushed "up" or "down" on the second face when done scrambling.

wca-guidelines.md
```diff
@@ -229,6 +226,12 @@ To be more informative, each Guideline is classified using one of the following
 - E3d++) [CLARIFICATION] Competitors should not consider a personal stopwatch or watch as the official time, and must submit their solution when the judge calls "STOP".
 
 
+## <article-F><clock><clocksolving> Article F: Clock Solving
+
+- F1+) [REMINDER] In the past, the competitor had to place the puzzle onto the mat in a standing position. This is no longer required.
+- F1++) [REMINDER] In the past, the organization team was allowed to enforce unsing a stand for the puzzle. This is no longer allowed.
```
ticklemepierce:
```suggestion
- F1++) [REMINDER] In the past, the organization team was allowed to enforce using a stand for the puzzle. This is no longer allowed.
```

wca-guidelines.md
```diff
@@ -235,6 +232,12 @@ To be more informative, each Guideline is classified using one of the following
 - E3d++) [CLARIFICATION] Competitors should not consider a personal stopwatch or watch as the official time, and must submit their solution when the judge calls "STOP".
 
 
+## <article-F><clock><clocksolving> Article F: Clock Solving
+
+- F1+) [REMINDER] In the past, the competitor had to place the puzzle onto the mat in a standing position. This is no longer required.
+- F1++) [REMINDER] In the past, the organization team was allowed to enforce using a stand for the puzzle. This is no longer allowed.
```
```suggestion
```

wca-regulations.md
```diff
@@ -529,11 +528,8 @@ Note: Because Article and Regulation numbers are not reassigned when Regulations
...
-- F2) The judge places the scrambled puzzle onto the mat in a standing position.
-    - F2a) The organization team may enforce using a stand for the puzzle to prevent it from falling before the start of the attempt. If such stands are used, the organization team should make an announcement before the round starts.
-        - F2a1) The judge must remove the stand from the mat immediately after the competitor removes the puzzle from it.
-- F3) At the end of the inspection period, the competitor places the puzzle in a standing position. Penalty: disqualification of the attempt (DNF).
-    - F3a) The competitor must not change the positions of any pins from their scrambled positions before the start of the solve. Penalty: disqualification of the attempt (DNF).
+- F2) The competitor may change the state of the pins during the inspection. These changes are not counted as moves.
+- F3) The competitor must not change the state of any inner clock faces in the puzzle during the inspection. Penalty: disqualification of the attempt (DNF).
```
```suggestion
- F2) The organization team may enforce using a stand for the puzzle to prevent it from falling before the start of the attempt. If such stands are used, the organization team should make an announcement before the round starts.
- F3) The competitor may change the state of the pins during the inspection. These changes are not counted as moves.
- F4) The competitor must not change the state of any inner clock faces in the puzzle during the inspection. Penalty: disqualification of the attempt (DNF).
```

CarterKoala (approved):

PR #1174, opened 2023-10-14
Add Visual Guide for 3j

Nevseros:


Nanush7:
LGTM!

ohexter:
Inspired by recent discussion on scramble signatures - would WRC consider including one or more 'ink on cube' examples?

alexmaasswca:
Aren't google drive links not allowed to be opened in China? Should probably provide an alternative there.

Nanush7:
@hanwu85 maybe you can add a different link in the Chinese translation?

hanwu85:
> @hanwu85 maybe you can add a different link in the Chinese translation?

Good question.
@oyyq99999, regarding the current visual guide to [5b5f+](https://www.worldcubeassociation.org/regulations/translations/chinese/guidelines.html#5b5f+)

We are aiming to add another visual guide.
Are cubers in China able to open a file in google drive?
If not, any alternative solution?

oyyq99999:
No, most cubers in China can not access google drive. Maybe we should make translations for the guides and put it somewhere else. But I don't know where I can access the sources.

hanwu85:
This is something I put together for TC version.  You can take a look.  Still double checking with local delegate though.
I just copy every sentence from the original text.
https://github.com/hanwu85/WCA-TC-2023_Open/blob/Draft/5b5visual.md

alexmaasswca:
> This is something I put together for TC version. You can take a look. Still double checking with local delegate though. I just copy every sentence from the original text. https://github.com/hanwu85/WCA-TC-2023_Open/blob/Draft/5b5visual.md

no images though?

> > @hanwu85 maybe you can add a different link in the Chinese translation?
> 
> Good question. @oyyq99999, regarding the current visual guide to [5b5f+](https://www.worldcubeassociation.org/regulations/translations/chinese/guidelines.html#5b5f+)
> 
> We are aiming to add another visual guide. Are cubers in China able to open a file in google drive? If not, any alternative solution?

could host the images on github, btw that draft you posted doesn't have any images embedded

hanwu85:
@alexmaasswca 
> no images though?

Correct, there are no images, this github file are intened for translation review and record purpose.
Not for the public use.
After translation is approved, we will copy the translated text to the original files and release as a seprate PDF file.
At least this is the original intention.

You mentions that we could host image on github. 
Personally, I have not tried that yet, but it might not be a bad idea.

Rouxles (approved):

Issue #1129, opened 2023-05-15
Wording of 2c+


sarahstrong314:
**Describe the typo**
Guideline 2c+ uses the phrasing "reasonable nickname" when there are better alternatives.

> [2c+](https://www.worldcubeassociation.org/regulations/#2c1) ADDITION First-time competitors should register using their legal name. They may register using a reasonable nickname, at the discretion of the WCA Delegate.

People who go by a name that is different from their legal name do not necessarily go by a nickname. Their preferred name could be a middle name or a completely different name that they have chosen for personal reasons.

**Possible solution**
Rephrase 2c+ to avoid the use of the term "nickname" and remove Delegate discretion to prevent discrimination.
> [2c+](https://www.worldcubeassociation.org/regulations/#2c1) ADDITION First-time competitors should register using their legal name or a preferred name that they regularly use.

PR #1175, opened 2023-10-15
Update 2c+

sarahstrong314:
Updated 2c+ per #1129. 

A few things this rewording accomplishes:

* Uses the term "preferred real name" instead of "nickname" which is more respectful to individuals who go by a name different than their legal name. The "real" part is to make it clear that Internet usernames don't count as valid names.
* Eliminates the phrasing "should use their legal name", which sounds discriminatory to competitors who have very good reason to not use their legal name. (Note: I don't think the original word was actually discriminatory since the definition of 'should' in RFC 2119 is different from its everyday usage, but it sounds discriminatory when you use the everyday definition of 'should').
* Removes Delegate discretion from deciding what is an appropriate non-legal name. The previous wording sounded like a Delegate could discriminate against any name they're unfamiliar with, including ethnic names. I believe [2c1](https://www.worldcubeassociation.org/regulations/#2c1) ties up the loose end of dealing with a competitor trying to register under a purposefully fake or inappropriate name.


CarterKoala:
Thanks Sarah! I'm very happy with the idea for this change and would like to include it for January 2024. @thewca/wrc-team thoughts?

My initial impressions:

- I don't really like the term "preferred real name" since that isn't really a term I've ever heard used and it's not immediately clear to me what "real" means (I've mainly just heard "preferred name"). Maybe something like "Competitors must register using either their legal name or a name used in everyday life." would be better
- I'm not sure about removing delegate discretion. Someone has to decide if a name is appropriate, and the delegate is probably the best option (and will better understand how names function in their local community). 2c1 just shifts this from the Delegate to the Organization team, which IMO makes it worse. Any delegate that would discriminate against a name for a reason such as that is unfit to be a Delegate, and should probably be investigated by WEC.
- I wouldn't be opposed to making an explicit guideline that trans individuals are never required to share their legal name with the WCA

ohexter:
I agree with Carter's first and second bullets.

On the third point there shouldn't be a need to see their legal first name (and gender) based on our approach to these fields, but we still need to verify legal last name, date of birth and nationality. ID documents are likely to show all of these pieces of data together and I wouldn't want such a guideline to prevent the latter items from being verified, so there could be practical challenges.

sarahstrong314:
Hi Carter,

2c+ has been on my mind since earlier this year when I first critically thought about its wording. I've pondered all kinds of alternative wordings over the past few months and carefully settled on "preferred real name" for a few reasons:

* Although "preferred name" is a well established term in English, its nuance cannot be easily translated into other languages and its nuance in English can be missed by new English learners. By adding "real", it makes it clear that the name still has to meet the vague standards of what constitutes a name. The literal intepretation of "preferred name" invites Internet usernames that are clearly not real names.

* "A name used in everyday life" doesn't specify that it has to be *their* name and even if it's changed to "a name they use in everyday life" then it misses the edge case of someone who just adopted a new name since they wouldn't have started used it in their everyday life. This phrasing also has the issue of inviting Internet usernames.

Regarding the consequences of removing Delegate discretion in 2c+, why not update 2c1 so that registrations are complete at the discretion of the organization team **and WCA Delegate**? That way, there's still a way that Delegates can intervene in the rare event that a competitor registers under a purposefully offensive fake name.

CarterKoala:
Hmm... I think you make good points, but I still don't like "preferred real name". I'll think about it more. 

@ohexter Personally I would not make someone show me an ID in that scenario and would just have them confirm the information they provided verbally, but I think that that point is enough reason to not make it a guideline. 

I added this as [another feedback thread](https://forum.worldcubeassociation.org/t/january-2024-regulations-feedback-update-2c-regarding-allowed-names/42149) on the forum, though I don't expect it will get much attention.

slongbehn:
Thanks for bringing up this issue, Sarah!

If I read "preferred real name" I would be pretty confused, as "real name" is often used to refer to someone's legal name. At a glance, it reads as an oxymoron to me. (IE, "What is your _real_ name?" - something trans people hear often.) It is also just entirely awkward.

All forms I fill out now say "Preferred name." While this may invite those to enter "usernames" since it is an online form, I'd like to think most people would not do that. I agree with the "delegate discretion" remaining to disallow this from happening. I also do not expect this to be any more common than usual name issues we encounter. I agree with this part:

> Regarding the consequences of removing Delegate discretion in 2c+, why not update 2c1 so that registrations are complete at the discretion of the organization team and WCA Delegate? That way, there's still a way that Delegates can intervene in the rare event that a competitor registers under a purposefully offensive fake name.

In terms of other languages, [there are words/phrases for this in many languages](https://www.reddit.com/r/asktransgender/comments/wv2dmx/how_do_you_say_preferred_name_in_your_language/?rdt=40922) and it's becoming more common. 

A thought for rephrasing entirely: "Competitors with chosen or given first names that do not match their legal name may use another first name at the discretion of the WCA Delegate."

I also think it would be valuable for some (private) outreach to trans competitors/staff if possible, to determine how they might feel is best to address this issue.

sarahstrong314:
Hi Shain,

> If I read "preferred real name" I would be pretty confused, as "real name" is often used to refer to someone's legal name. At a glance, it reads as an oxymoron to me. (IE, "What is your real name?" - something trans people hear often.) It is also just entirely awkward.

I honestely had the opposite reaction. I made the PR soon after seeing "preferred real name" for the first time. To me, a trans person's real name is their preferred name. 

> In terms of other languages, [there are words/phrases for this in many languages](https://www.reddit.com/r/asktransgender/comments/wv2dmx/how_do_you_say_preferred_name_in_your_language/?rdt=40922) and it's becoming more common.

Thanks for sharing that! Really good to know. :)

> A thought for rephrasing entirely: "Competitors with chosen or given first names that do not match their legal name may use another first name at the discretion of the WCA Delegate."

That makes it sound like Delegates have the ability to stop a trans competitor from using their self-given name and instead use their birth name. That's the very wording I was trying to eliminate.

> I also think it would be valuable for some (private) outreach to trans competitors/staff if possible, to determine how they might feel is best to address this issue.

For sure! :)

slongbehn:
> That makes it sound like Delegates have the ability to stop a trans competitor from using their self-given name and instead use their birth name. That's the very wording I was trying to eliminate.

Out of curiosity, how do you feel about the phrasing without that section about delegate discretion?

sarahstrong314:
> Out of curiosity, how do you feel about the phrasing without that section about delegate discretion?

This?
 _"Competitors with chosen or given first names that do not match their legal name may use another first name"_

What would go before it? If it's the current first sentence _"First-time competitors should register using their legal name.",_ 
 it makes seem like people who don't go by their legal name are treated as an afterthought. I was being deliberate putting both in the same sentence.

slongbehn:
I really don’t feel like that is an “afterthought” but rather a clarification. If you feel that is the case it can be rephrased as:

“First-time competitors should register using their legal last name. Their first name may be a competitor’s legal first name, chosen name, or given name.”

sarahstrong314:
> I really don’t feel like that is an “afterthought” but rather a clarification. 

It can be both. _"First-time competitors should register using their legal name"_ is another phrasing I was trying to eliminate since it's questionable upon first read without knowing the intended definition of "should".

> “First-time competitors should register using their legal last name. Their first name may be a competitor’s legal first name, chosen name, or given name.”

Making separate rules for different parts of a name gets messy since there are different conventions around the world for the parts of a full name. Furthermore, there are valid reasons why someone would go by a different last name than their legal last name, such as someone who is married or divorced who goes by a new last name (or returned to a former last name) in their everyday life but didn't complete the legal paperwork to officially change it.

slongbehn:
> it's questionable upon first read without knowing the intended definition of "should".

That is the same issue I take with "real name." Where "should" is defined.

There is no possible way to conceivably account for every system. My intention with "legal last name" was to address this concern from @ohexter :

> we still need to verify legal last name, date of birth and nationality

It doesn't seem like any alternatives are willing to be considered from your point of view, so I will leave this discussion with my input to WRC that I am not in favor of the proposed phrasing. Thanks!

sarahstrong314:
Hi Shain,

> It doesn't seem like any alternatives are willing to be considered from your point of view, so I will leave this discussion with my input to WRC that I am not in favor of the proposed phrasing. Thanks!

Thanks for the conversation. :)

Just to clarify, I'm open to "preferred name" as it's definitely better then the current wording of "nickname". My initial thought when I first read the Guideline was that it should say "preferred name", but then I became aware of some potential loopholes. Based on the feedback so far, I've updated the PR to "preferred name" (without the "real").

Edit: I also added a commit to update 2c1 so that Delegate discretion can be used to stop competitors from competing under purposeful offensive/inappropriate/fake names.

Edit 2: I added one more commit to append an additional sentence to 2c+:

> First-time competitors who purposefully register using an inappropriate or offensive name may be disqualified at the discretion of the WCA Delegate.

This way there is Delegate discretion, but it's only towards competitors with clear malicious intent when filling out the name field as opposed all competitors who don't use their legal name.

lgarron:
I am very supportive of the intent of this change, but I'm a little concerned that the current phrasing will slowly lead to people competing with pseudonyms that are less and less connected to their real name. I don't know that we want to allow people to compete as "PewDiePie" or "GanCubeLover69".

I can't think of a good way to phrase it right now, but I'd like to see some sort of guard that expresses "nicknames and name-like pseudonyms are fine if they are what you use in real life, but distinct persona names should not generally be allowed" (with some defined process for when there are privacy concerns).

sarahstrong314:
Hi Lucas,

> I don't know that we want to allow people to compete as "PewDiePie" or "GanCubeLover69".

I agree that these types of names should _not_ be allowed. My original wording covered these cases but I've changed it since then.

 > I can't think of a good way to phrase it right now, but I'd like to see some sort of guard that expresses "nicknames and name-like pseudonyms are fine if they are what you use in real life, but distinct persona names should not generally be allowed" (with some defined process for when there are privacy concerns).
 
I think distinct persona names ought to be considered "inappropriate" in "_purposefully register using an **inappropriate** or offensive name_" but perhaps there's a better word or phrase to express it better.

Samuel-Baird:
Is delegate discrimination a realistic concern? Have there been any instances of this actually happening?

Instances of Delegates not allowing names because of *insert predudice here* is already dissalowed by the CoC/CoE, so I'm not sure there is a need to completely remove the ability of anyone from blocking innaproriate/non-serious names. 

To me it seems like it would be fine to make the wording more clear that competitors are welcome to use any sort of reasonable nick-name/alias/preffered name, at the discretion of the Delegate

sarahstrong314:
> Is delegate discrimination a realistic concern? Have there been any instances of this actually happening?

Probably not since it's covered within the CoC/CoE, however competitors won't all know that. We don't want competitors to come to a competition concerned that they will need to use their deadname. We also don't want competitors to think that Delegates have the ability scrutinize their real name if the Delegate doesn't think it looks real due to racial bias. My goal is to simply make 2c+ align with equitable practices since the way it's written now is misleading in unfortunate ways. It seems likely that the Guidelines will soon be merged with the Regulations which will make this Guideline more visible to competitors reading the Regulations.

Samuel-Baird:
> Probably not since it's covered within the CoC/CoE, however competitors won't all know that. We don't want competitors to come to a competition concerned that they will need to use their deadname. We also don't want competitors to think that Delegates have the ability scrutinize their real name if the Delegate doesn't think it looks real due to racial bias.

Luckily enough for us, most competitors won't read the regulations, let alone the guidelines, before competing for their first time, so a majority of them won't read 2c+ and misinterpret it ;) 

Maybe to avoid competitors misunderstandings that Delegates have discretion of denying or accepting names based on prejudice, we can remove delegate discretion from 2c+ and then create a 2c++ to visually seperate the ideas. 2c++ could specifies that Delegates have discretion to bar innapropriate/unreasonable names (with better wording ofc) and perhaps list a few examples of what would and would not be considered reasonable. 

On a side note, If the goal is avoiding incidents where competitors are worried about using their preffered names, then the single best place to make changes would be on the page where they create their profile (which I believe could use some more detail in other regards as well). I just made a PR for it [here](https://github.com/thewca/worldcubeassociation.org/issues/8428), @sarahstrong314 it would be great if you could give some input to help make the profile creation page clear that WCA is inclusive by not requiring competitors to legal names and that competitors can use nicknames/aliases/preffered names instead

image

c-goodyear:
The 2c1 change makes a lot of sense.

The current wording suggestion for 2c+ however leaves the door open to also allow for a competitor to freely choose a non-legal surname, which is not something we allow. This wording is way too loose in allowing pseudonyms to be accepted while following the letter of the regulation.

dunkOnIT:
What do we think about being more explicit with the wording? There is a clear intention here: **allow people - particularly trans people - to compete under the name which they most identify with**

There is also clear abuse we're trying to avoid: people spuriously giving a name which is not their preferred name. 

Should we not just explicitly state those cases/intentions in the guideline? I understand the desire to not make the guideline overly specific, but I wonder if by trying to generalize/abstract it too much, we make it less clear for delegates and more open to abuse.

sarahstrong314:
> What do we think about being more explicit with the wording? There is a clear intention here: **allow people - particularly trans people - to compete under the name which they most identify with**
> 
> There is also clear abuse we're trying to avoid: people spuriously giving a name which is not their preferred name.
> 
> Should we not just explicitly state those cases/intentions in the guideline? I understand the desire to not make the guideline overly specific, but I wonder if by trying to generalize/abstract it too much, we make it less clear for delegates and more open to abuse.

Do you have any recommendations for how to word this?

coder13:
Are we trying too hard to make sure people don't have the name "PewDiePie"? Is this a "this is why can't have nice things" issue? Should the delegate just go through the list of registrants to make sure no one is putting dummy names? People still can put any name in when signing up and likely do without reading the regulations. Can we trust people a bit more to put their preferred name?

SonjaBlack:
All in favor of more inclusive wording. As a suggestion how about "competitors should register using the name they prefer to be called in real life."

It's not as terse as "preferred real name", but I think it makes up for that in clarity of the intentions behind the regulation.

sarahstrong314:
I've made another edit per Sonja's suggestion and Callum's input. 

Here's what it looks like now:

> 2c+) [ADDITION] First-time competitors must register using their legal name or the name they prefer to be called in real life. Last names must be legal last names. First-time competitors who purposefully register using an inappropriate or offensive name may be disqualified at the discretion of the WCA Delegate.

* Avoids othering people who don't use their legal name by avoiding starting with the phrase "competitors should register using their legal name" then giving the exception. Instead, all competitors regardless of whether they use their legal name are treated as equal.
* "Name they prefer to be called in real life" is more wordy than "preferred name" but it closes the loophole of using Internet names. I agree with Cailyn that it would be nice if we could just trust people to not abuse the ability to use a preferred name though.
* Worded so that trans people have the choice to use their legal name if they feel that using their preferred name would put them in danger.
* Clarifies that last names must be legal last names. In other words, "preferred name" only applies to first and/or middle names, which is consistent with the name rules in the Delegate Crash Course. I honestly don't think we should be forcing people to use their legal last names, but I think that should be a separate discussion for another time.
* "Last names must be legal last names" is worded as such to account for individuals who have multiple last names or no legal last name.

Samuel-Baird:
Disqualification wasn’t previously included in 2c+, so that should probably be changed

sarahstrong314:
> Disqualification wasn’t previously included in 2c+, so that should probably be changed

Done. Here's what it looks like now:

> 2c+) [ADDITION] First-time competitors must register using their legal name or the name they prefer to be called in real life. Last names must be legal last names.

cubizh:
I've always appreciated the possibility of allowing competitors to register with other than their "legal name". To me, the definition of term "legal name", in this context, is the person's full name as it is written in some ID card, for most americans that includes a first name, a middle name and last name, in other parts of the world it may differ. Registering first and last names only has been common practice in many countries and for many competitors, and also in some places of the world, people often prefer to skip their first name and prefer to be called by their second name first. That has been accepted and I think most people do not register their legal name, and that's fine, because it's not easily exploitable combined with the date of birth. 

Regarding the issue that was raised, I'd say it could be more easily addressed by just quietly getting rid of the term "legal".
When I ask a trans person (or anyone for this matter) their "name", they don't give me their "legal name", or their "birth name", they give me the name they want to be addressed as, which can be a a combination of names they have in their ID, some nickname combined with a name or some other name they identify with, that has some logical reasoning.

To me, the main reason of the concern revolves around the "legal" part, which is a more formal definition that can be intimidating for some people, which is understandable. It would be more subtle and effective to just remove it, rather than making the point more visible to everyone, which probably has an undesired effect more than it benefits, particularly because it will open the door to issues that haven't been discussed here which will require a lot more of Delegate scrutiny (or some other entity) because even though we absolutely need to be accommodating to everyone, we also need to have ways to prevent trolling attempts and ways to maintain database integrity.

Nevseros (approved):
LGTM!

PR #1189, opened 2024-01-29
Automatically deploy WCA-Regulations to S3

FinnIckler:
Due to infrastructure changes we will deploy wca-regulations directly to S3 and serve them from there. These will automatically update on the website when the version file changes

### DIFF
--- a/.github/workflows/deploy-to-s3.yml
+++ b/.github/workflows/deploy-to-s3.yml
@@ -25,7 +25,7 @@ jobs:
         python -m pip install --upgrade pip
         pip install wrc
         # install wkhtmltopdf
-        wget https://github.com/wkhtmltopdf/wkhtmltopdf/releases/download/0.12.4/wkhtmltox-0.12.4_linux-generic-amd64.tar.xz -O wkhtml.tar.xz && tar -xf wkhtml.tar.xz --strip-components=1 -C /usr/local
+        wget https://github.com/wkhtmltopdf/wkhtmltopdf/releases/download/0.12.4/wkhtmltox-0.12.4_linux-generic-amd64.tar.xz -O wkhtml.tar.xz && sudo tar -xf wkhtml.tar.xz --strip-components=1 -C /usr/local
     - name: Deploy to S3
       run: |
         outputdir=/tmp/regulations
--- a/.github/workflows/deploy-to-s3.yml
+++ b/.github/workflows/deploy-to-s3.yml
@@ -24,7 +24,6 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install wrc
-        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
         # install wkhtmltopdf
         wget https://github.com/wkhtmltopdf/wkhtmltopdf/releases/download/0.12.4/wkhtmltox-0.12.4_linux-generic-amd64.tar.xz -O wkhtml.tar.xz && tar -xf wkhtml.tar.xz --strip-components=1 -C /usr/local
     - name: Deploy to S3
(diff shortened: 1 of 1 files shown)


FinnIckler:
I think merging this into draft would be the way to go?

.github/workflows/deploy-to-s3.yml
```diff
@@ -0,0 +1,40 @@
...
+      uses: aws-actions/configure-aws-credentials@v4
+      with:
+        aws-access-key-id: ${{ secrets.CI_CD_AWS_ACCESS_KEY_ID }}
+        aws-secret-access-key: ${{ secrets.CI_CD_AWS_SECRET_ACCESS_KEY }}
+        aws-region: ${{ secrets.AWS_DEFAULT_REGION }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install wrc
+        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
```
gregorbg: Is this line necessary, and where does it originally come from?

.github/workflows/deploy-to-s3.yml
```diff
@@ -0,0 +1,40 @@
...
+      with:
+        python-version: 3.9
+    - name: Configure AWS credentials
+      uses: aws-actions/configure-aws-credentials@v4
+      with:
+        aws-access-key-id: ${{ secrets.CI_CD_AWS_ACCESS_KEY_ID }}
+        aws-secret-access-key: ${{ secrets.CI_CD_AWS_SECRET_ACCESS_KEY }}
+        aws-region: ${{ secrets.AWS_DEFAULT_REGION }}
+    - name: Install dependencies
+      run: |
```
gregorbg, line 24: `wca-documents` has the approach of using an `install-dependencies.sh` script, whereas here you're installing stuff inline in the GitHub action. Can you decide on which is better and enforce it consistenly?

(I'm leaning towards installing stuff inline in the YML, in which case I should have probably made this comment over on the other PR...)

.github/workflows/deploy-to-s3.yml
```diff
@@ -0,0 +1,40 @@
...
+        outputdir=/tmp/regulations
+        mkdir -p $outputdir
+        wrc --target=json -o $outputdir .
+        wrc --target=html -o $outputdir .
+        wrc --target=pdf -o $outputdir .
+        # Update version
+        git_hash=$(git rev-parse --short "$GITHUB_SHA")
+        echo "$git_hash" > $outputdir/version
+        aws s3 sync $outputdir s3://wca-regulations/
```
gregorbg: More nit about consistency: You're directly dialing `s3://wca-regulations/`, whereas in the wca-documents PR you're passing the bucket name as an env and then dialing `s3://$SOME_ENV_VAR`. Please decide on one approach and use it consistently across our build scripts :)

gregorbg (changes requested):

.github/workflows/deploy-to-s3.yml
FinnIckler: it's from the other workflow in this repo

.github/workflows/deploy-to-s3.yml
FinnIckler: yeah because in the other action I'm using .sh files. I'll just make it inline everywhere

.github/workflows/deploy-to-s3.yml
FinnIckler: yeah I'll make everything inline

.github/workflows/deploy-to-s3.yml
FinnIckler: I removed it

gregorbg (approved):

PR #1191, opened 2024-01-31
Add Action to Move Current Regulations to History

FinnIckler:
This should be triggered before a new regulations cycle. It can be then linked to in [the history file](https://github.com/thewca/worldcubeassociation.org/blob/master/WcaOnRails/app/views/regulations/history.html.erb)

### DIFF
--- a/.github/workflows/cycle-regulations.yml
+++ b/.github/workflows/cycle-regulations.yml
@@ -1,0 +1,25 @@
+name: Move Regulations to History
+
+on:
+  workflow_dispatch:
+    inputs:
+      folder_name:
+        required: true
+        description: The name of the sub folder to move the regulations to, usually the date the regulations come out like '2023-08-01'
+
+jobs:
+  move:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-access-key-id: ${{ secrets.CI_CD_AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.CI_CD_AWS_SECRET_ACCESS_KEY }}
+          aws-region: ${{ secrets.AWS_DEFAULT_REGION }}
+      - name: Move current version to the history
+        run: |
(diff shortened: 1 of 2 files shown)


Nevseros:
This does not include adding a line saying that this version is not the current one, right?

FinnIckler:
You are right, I'll add that tomorrow

FinnIckler:
https://github.com/thewca/worldcubeassociation.org/blob/regulations-data/regulations/build/regulations/history/official/2023-08-01/index.html.erb
it looks like it also wasn't added the last time regulations were cycled!

Nevseros:
@FinnIckler 
> it looks like it also wasn't added the last time regulations were cycled!

Because we add regulations on the history page as soon as we announce them (2024-01-01 should be there too now, for example, but I've been lazy) to be able to reference a current regulation with permalink (during investigations, for example). 

Is it possible that upon adding a current version to the history the previous version file is edited with the disclaimer, but not the one that's being added? Like I did in these two PRs in total: [8183](https://github.com/thewca/worldcubeassociation.org/pull/8183) and [8283](https://github.com/thewca/worldcubeassociation.org/pull/8283).

FinnIckler:
The github action can't figure out what the current/previous version is unless you explicitly tell it through an input variable. What we could do is also add the current version to its history folder every time the official branch is updated and then have this action that just inserts the disclaimer for a given version.

FinnIckler:
So I can pick this one up again. What does WRC think about my suggestion about replacing the current regulations in history every time it's updated?

Nanush7:
How would the process look like from our side? We just need to update the version file in the history folder?

FinnIckler:
No, the process from your side would just be like now, you push to this repo to the official branch and that would replace both the latest version and the latest version in history. If you then need to add the disclaimer you can run this Action.

~~I am looking into it and I think I can't implement it here, instead the wrc compiler needs to put the current files into the history folde~~

I can just do it here

FinnIckler:
I added a step to the deploy-to-s3 action that copies the current version into the history. 
When you cycle regulations you just need to:
- use the cycle-regulations actions to add the This is NOT the current official version in there with the version you want to add it to as the input parameter
- Make a PR to the Main Repo to add a link to the new History file

Nanush7:
Thank you @FinnIckler! Are we ready to merge this?

Nanush7 (approved):
I don't have any way to test this, but I don't see any problem with the changes.

PR #1199, opened 2024-06-13
Compile merged Regulations and Guidelines

Nanush7:
Add html with the merged flag.

The new option generates a `full.html.erb` file in the output dir. So now we have three html files, two for the separated documents and one for the merged version.

See https://github.com/thewca/worldcubeassociation.org/pull/9541 and https://github.com/thewca/wca-regulations-compiler/pull/28.


gregorbg (approved):

gregorbg:
This now needs to move from `draft` branch to the `official` branch, what's the usual WRC procedure for that?

Nanush7:
We would normally require Board's approval, but this is not a change to the Regulations per se.

Having another WRC member approve the merge to official should be enough.

PR #1200, opened 2024-06-13
Compile merged Regulations and Guidelines

Nanush7:
Add html with the merged flag.

The new option generates a `full.html.erb` file in the output dir. So now we have three html files, two for the separated documents and one for the merged version.

See https://github.com/thewca/worldcubeassociation.org/pull/9541 and https://github.com/thewca/wca-regulations-compiler/pull/28.

### DIFF
--- a/.github/workflows/deploy-to-s3.yml
+++ b/.github/workflows/deploy-to-s3.yml
@@ -33,6 +33,7 @@ jobs:
         git_hash=$(git rev-parse --short "$GITHUB_SHA")
         wrc --target=json -g $git_hash -o $outputdir .
         wrc --target=html -g $git_hash -o $outputdir .
+        wrc --target=html --merged -g $git_hash -o $outputdir .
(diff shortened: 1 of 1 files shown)


Nevseros (approved):

lgarron:
> The new option generates a `full.html.erb` file in the output dir. So now we have three html files, two for the separated documents and one for the merged version.

I didn't catch that name until now — do you think it would be easy to switch to `merged.html.erb`?

I think "full" implies that either of the merged documents is "partial", which isn't really accurate.

gregorbg:
> I didn't catch that name until now — do you think it would be easy to switch to `merged.html.erb`?
> 
> I think "full" implies that either of the merged documents is "partial", which isn't really accurate.

Switching the name requires an update to the Python WRC-compiler, which is easy enough. It won't require another code change in this repository, just a new trigger of the GH Action with the updated `wrc` version.

PR #1217, opened 2024-09-04
Clarify the preferred approach when a judge forgets to use a sight blocker

CarterKoala:


wca-guidelines.md
```diff
@@ -207,8 +207,7 @@ To be more informative, each Guideline is classified using one of the following
...
 - B2f+) [CLARIFICATION] Since the competitor starts the solve by lifting the cover, penalties for starting the timer and for touching the cover while starting the attempt do not apply (see [Regulation B2b](regulations:regulation:B2b) and [Regulation B2c](regulations:regulation:B2c)).
-- B4c+) [ADDITION] If the judge forgot to put a sight blocker, or places it in a way that the competitor's view of the puzzle is not fully blocked, the WCA Delegate should grant an extra attempt.
-- B4c++) [ADDITION] If the WCA Delegate does not suspect that the competitor intentionally violated any WCA Regulations, the original attempt may stand.
+- B4c+) [ADDITION] If the judge forgot to put a sight blocker and the WCA Delegate does not suspect that the competitor intentionally violated any WCA Regulations, the original attempt may stand.
```
ohexter: Should this stay as B4c++ to avoid renumbering confusion?

Rouxles: i believe this has been done in the past (and regardless, for cleanliness/backwards-compatibility sake, it's probably better to keep it as B4c++)

CarterKoala: In general the numbers aren't changed, but it seemed reasonable here since the wording matched with the original B4c+ with just a change in the outcome. I'm fine either way though

```suggestion
- B4c++) [ADDITION] If the judge forgot to put a sight blocker and the WCA Delegate does not suspect that the competitor intentionally violated any WCA Regulations, the original attempt may stand.
```

ohexter (commented):
Switching to B4c++ for backwards compatibility

ohexter (approved):

PR #1256, opened 2024-09-04
Remove F2

CarterKoala:
Alternate to #1255


CarterKoala:
Closing as we will move forward with #1255

PR #1258, opened 2024-09-04
Requirements for placing the sight blocker (Judge)

CarterKoala:
Alternative to #1257


CarterKoala:
Moving forward with #1257

PR #1273, opened 2025-01-07
Add Action to Move Current Regulations to History (#1191)

dunkOnIT:
This should be triggered before a new regulations cycle. It can be then linked to in [the history
file](https://github.com/thewca/worldcubeassociation.org/blob/master/WcaOnRails/app/views/regulations/history.html.erb)


Nanush7 (approved):

Nanush7:
This is not a regulations change, so it shouldn't need approval from the Board.

PR #1279, opened 2025-02-02
Fix minor pluralisation

norwd:
This corrects "Regulation*s*/Guideline" to "Regulation*s*/Guideline*s*".


Nanush7 (approved):
Thank you spotting and fixing this!

PR #1285, opened 2025-04-09
Clerical updates to Guidelines

slongbehn:


wca-guidelines.md
```diff
@@ -175,7 +175,7 @@ To be more informative, each Guideline is classified using one of the following
 
 ## <article-12><notation><notation> Article 12: Notation
 
-- 12a2+) [CLARIFICATION] For example, both Rw and 2Rw are valid notation for the same physical move for the 3x3x3 Cube. 1Rw is not valid notation for any NxNxN Cube, and 3Rw is not valid notation for the 2x2x2 Cube or the 3x3x3 Cube (but 3Rw is valid notation for the 4x4x4 Cube and larger NxNxN Cubes).
+- 12a2+) [CLARIFICATION] For example, both Rw and 2Rw are valid notations for the same physical move for the 3x3x3 Cube. 1Rw is not valid notation for any NxNxN Cube, and 3Rw is not valid notation for the 2x2x2 Cube or the 3x3x3 Cube (but 3Rw is valid notation for the 4x4x4 Cube and larger NxNxN Cubes).
```
CarterKoala: I'm not convinced this one is correct. From my (brief) research, it seems that notation refers to the set of symbols, rather than individual symbols themselves. Do you have any research suggesting otherwise? 

Also see how the next sentence is formatted - 1Rw is not valid notation, rather than 1Rw is not **a** valid notation

ohexter: Agree with Carter, "notation" is more intuitively correct to me. As this is a bit more contentious it can be excluded from these clerical changes but considered later.

ohexter:
```suggestion
- 12a2+) [CLARIFICATION] For example, both Rw and 2Rw are valid notation for the same physical move for the 3x3x3 Cube. 1Rw is not valid notation for any NxNxN Cube, and 3Rw is not valid notation for the 2x2x2 Cube or the 3x3x3 Cube (but 3Rw is valid notation for the 4x4x4 Cube and larger NxNxN Cubes).
```

slongbehn: I see what you mean, yes, that makes sense.

KerrieJarman:
Approved by Board vote

CarterKoala (approved):

wca-guidelines.md
lgarron: > I'm not convinced this one is correct. From my (brief) research, it seems that notation refers to the set of symbols, rather than individual symbols themselves. Do you have any research suggesting otherwise?
> 
> Also see how the next sentence is formatted - 1Rw is not valid notation, rather than 1Rw is not **a** valid notation

Agreed. As the author of this Guideline: "notation" was intentional, I think "notations" is not as appropriate.

In fact, I would be concerned about landing this change in the official Regs, as this might make my job harder when documenting and standardizing notation for new events and puzzles.

Apologies for just noticing this PR now. Is it possible revert avoid landing this in `official` without going to the Board?

lgarron: Ah, I see that this part actually didn't land. Nothing actionable, then.

PR #1289, opened 2025-07-18
Merge the WCA Regulations and Guidelines into official

FinnIckler:
Any specific format you want to PRs like this?


CarterKoala:
Is this a question for WRC? If so I'm not really sure what you mean.

FinnIckler:
The commit created by this pr will show up linked on the regulations page
so we should make sure the commit message reflects the changes

ohexter:
The following (edited) text from the original commit will be a useful record:

> - Insert each Guideline into the Regulations, immediately following its corresponding Regulation. Former Guidelines keep their current reference, and are inserted at the same indentation level as their corresponding parent Regulation.
> - Update the preliminary Notes section of the Regulations to include relevant material from the Guidelines.
> - Change all references and links from “GuidelineXX+” to “RegulationXX+”, and update outdated URL links.
> - Make necessary clerical changes to remove existing references to the Guidelines, and reflect the Regulations' new role as a combined document.
> - Make no substantive changes that would impact the application of the Regulations at WCA competitions.

The history page will also link to the announcement and the full diff, so an exhaustive explanation isn't necessary here.

PR #1295, opened 2026-06-17
Add commas before "at the discretion of the judge" in Article I.

lgarron:
This matches the rest of the Regulations, especially when the clause comes after a parenthetical.


lgarron:
(This isn't at all critical; I just noticed it while reading Article I.)

